PR #35, opened 2015-08-12
Key length case and ably common

mattheworiordan:
@paddybyers whilst updating the feature spec in preparation for a complete Realtime & REST spec for Ruby with links to tests, I noticed that we are not using the standard `camelCase` for `keyLength` cipher param.

I tried to fix it, but I am experiencing two issues:

* I keep getting an invalid key length exception from Java for some reason with the 256 length key
* When checking to see if this problem is local, I checked CI, and I see that the tests are hanging every time.  See https://travis-ci.org/ably/ably-java/builds/73015292

As I don't have a very good development environment, would you mind looking at these two issues so that we can get Java passing again?


mattheworiordan:
Weirdly it seems a build has now passed, see https://travis-ci.org/ably/ably-java/jobs/75319097

paddybyers:
> I keep getting an invalid key length exception from Java for some reason with the 256 length key

Perhaps you have the export-restricted algorithms installed?

mattheworiordan:
> Perhaps you have the export-restricted algorithms installed?

Perhaps, if that is the case, I think we should add a note to the Readme as it's certainly not an obvious issue.  Added to that, the instructions to run a single test don't work either.

In terms of the stalling tests, I think there is an issue in the suite that would need to be fixed as I had the same stalling problems when running the suite locally.

PR #43, opened 2015-09-08
RSC1 and RSC2 - initialisation and default logging behaviour

iliyakostadinov:
(RSC1) The constructor accepts either an API key, a token string, or a set of ClientOptions. An exception is raised if invalid arguments are provided such as no API key, token and no means to create a token
(RSC2) The logger by default outputs to STDOUT (or other logging medium as appropriate to the platform) and the log level is set to warning

### DIFF
--- a/test/io/ably/test/rest/RestInit.java
+++ b/test/io/ably/test/rest/RestInit.java
@@ -38,27 +38,27 @@ public class RestInit {
 	 */
 	@Test
 	public void init_null_key_string() {
-	  try {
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Thanks, one small comment, but looks good  :+1: 
@paddybyers?

mattheworiordan:
@paddybyers can we merge?

paddybyers:
> can we merge?

Yes.

PR #47, opened 2015-11-03
Make inclusion of android-test project conditional on whether or not …

paddybyers:
…Android SDK is configured

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,3 @@ project(':lib') {
     outputs.upToDateWhen { false }
   }
 }
-
-project(':android-test') {
-}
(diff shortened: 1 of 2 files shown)


mattheworiordan:
@paddybyers Travis CI is failing currently because the submodule is referenced from git@ as opposed to via `http://`.  See https://github.com/ably/ably-java/blob/1db9a7f97d2a0a97e39ed22e73c29c1d2aaaecda/.gitmodules.  Weirdly though if you look on `master` at https://github.com/ably/ably-java/blob/master/.gitmodules, it uses the correct path. Can you ensure that `gradle-build` upon which this depends is updated accordingly as this should really pass Travis before we release?

PR #52, opened 2015-11-16
Set new thread as daemon to allow application shutdown

petergphillips:
Further to [pull request 51](https://github.com/ably/ably-java/pull/51), I have made the same change in the gradle build branch.

I tried the gradle-build branch as is, but my application still hung as connection manager was creating a new user thread.  This fix ensures that if the main thread is terminated, the separate connection manager thread is terminated too.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -362,6 +362,7 @@ public class ConnectionManager implements Runnable, ConnectListener {
 		synchronized(this) {
 			if(mgrThread == null) {
 				mgrThread = new Thread(this);
+				mgrThread.setDaemon(true);
 				state = states.get(ConnectionState.initialized);
 				creating = true;
 			}


mattheworiordan:
LGTM :+1:

mattheworiordan:
Is this not largely the same as #51?  Perhaps close that one in favour of this if this addresses the same problem?

mattheworiordan:
@paddybyers what is the status on this PR?

paddybyers:
> @paddybyers what is the status on this PR?

This is setting the connection manager thread to be a daemon thread, and I'm not sure I agree it is the correct thing to do. A daemon thread will be abruptly terminated by the VM as soon as there are no non-daemon threads attached to the VM, and this will prevent the library from performing normal shutdown; this will mean for example that if an application's main thread calls `close()` on the library and then just returns from `main()` it will typically shut down the VM so that the library won't send a `CLOSE` message to Ably.

Instead I changed `ConnectionManager` so that the manager thread exits in the `closed` state. This means that the VM can exit normally provided that the user has called `close()` on any connected `AblyRealtime` instance. If the user hasn't closed the connection explicitly, then the connection manager thread will prevent the VM from exiting unless it is forcibly shut down with `System.exit()`.

So it does depend on what we think we want to support. Should we require applications to call `close()` to close connections cleanly? If they don't, there will be lingering connection state in Ably until the connections time out.

I think we should also consider adding a shutdown hook so that `close()` is called if a connection is open when the VM is shutdown with `System.exit()`.

mattheworiordan:
I think it's acceptable that if a connection is opened, and not explicitly closed, then it should remain alive until explicitly closed. However, I would agree that we could certainly add a shutdown hook as it's unlikely anyone would want the explicit shut down a VM to be blocked by our library

mattheworiordan:
Should we close this PR and raise a separate issue perhaps?

paddybyers:
> However, I would agree that we could certainly add a shutdown hook as it's unlikely anyone would want the explicit shut down a VM to be blocked by our library

As things currently stand, an explicit shutdown via `System.exit()` will not be blocked. If the connection is open then it will be silently closed, and Ably will be left with an orphaned connection until it times out. A shutdown hook will prevent that - but we don't need it to enable the VM to exit when there is a `System.exit()`.

paddybyers:
> Should we close this PR and raise a separate issue perhaps?

https://github.com/ably/ably-java/issues/71

mattheworiordan:
Thanks @petergphillips for the contribution, but as mentioned, we've decided to go a different route on this.

petergphillips:
Okay @mattheworiordan, thanks for the update.

In our scenario we are not doing a `System.exit`, instead we are throwing an exception from the main thread.  The main thread is then terminating, but because there are other user threads still active the VM is prevented from exiting.

I therefore see #71 as a separate issue, since that is about ensuring a clean shutdown of Ably when the VM terminates.  In my scenario Ably is actually preventing the VM from terminating.

I've tried surrounding my application with a try finally block that just called `AblyRealtime.close`, but that didn't help, the connection manager user thread wasn't stopped.  Looking at the code the `stopThread` method is only called if the manager enters a disconnected state, which wasn't happening and I couldn't work out how to make that happen.

One way I could work around it is surround my application with a try finally block and then `System.exit` if I get an exception, but that seems rather nasty.  Hence my solution that makes the connection manager a daemon thread, which to me seems the right thing to do as it sounds more like a daemon than a user process...

mattheworiordan:
Ok, thanks for explaining @petergphillips.  I have not written Java for a very long time now, so will get one of our Java bods to look at this issue

mattheworiordan:
@petergphillips we feel this is an issue with our implementation that needs to be looked at, but we should not necessarily deamonise the thread.  I am going to raise an issue now that will be looked at.

mattheworiordan:
@petergphillips I have added https://github.com/ably/ably-java/issues/73, we will keep you posted on progress addressing this

PR #74, opened 2016-01-14
Fix leaking non-AblyExceptions on ConnectionManager#onMessage callback

### LINKED ISSUE #41: NullPointerException: Attempt to invoke interface method 'java.lang.String java.security.Principal.getName()' on a null object reference
1. Create a new Android project in Android Studio via the _Simple Activity_ template
2. Attach `ably-android.jar` in project
3. Replace code in MainActivity with code from [here](http://pastebin.com/kReKYm2V) and fix package name.
4. Run on device
5. Get exception when calling `channel.history()` on line 61: 

`io.ably.types.AblyException: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String java.security.Principal.getName()' on a null object reference`

Here's the full log:

```
W/ActivityThread﹕ Application com.example.stan.myapplication is waiting for the debugger on port 8100...
I/art﹕ Late-enabling -Xcheck:jni
I/art﹕ VMHOOK: rlim_cur : 0 pid:12814
I/System.out﹕ Sending WAIT chunk
I/System.out﹕ (INFO): io.ably.realtime.AblyRealtime: started
I/System.out﹕ (INFO): Auth(): using token auth with client-side signing
I/System.out﹕ (VERBOSE): io.ably.transport.ConnectionManager: requestState(): requesting connecting; id = null
I/System.out﹕ (VERBOSE): io.ably.transport.ConnectionManager: notifyState(): notifying connecting; id = null
I/System.out﹕ (INFO): TokenAuth.authorise():
I/System.out﹕ (INFO): TokenAuth.authorise(): requesting new token
I/System.out﹕ (INFO): Auth.requestToken(): using token auth with client-side signing
I/System.out﹕ (INFO): Auth.getTokenRequest(): generated signed request
D/libc﹕ [NET] android_getaddrinfofornet+,hn 12(0x726573742e6162),sn(),hints(known),family 0,flags 4
D/libc﹕ [NET] android_getaddrinfofornet-, err=8
D/libc﹕ [NET] android_getaddrinfofornet+,hn 12(0x726573742e6162),sn(),hints(known),family 0,flags 1024
D/libc﹕ [NET] android_getaddrinfofornet-, pass to proxy
D/libc﹕ [NET] android_getaddrinfo_proxy+
D/libc﹕ [NET] android_getaddrinfo_proxy get netid:0
D/Atlas﹕ Validating map...
D/libc﹕ [NET] android_getaddrinfo_proxy-, success
D/libc﹕ [NET] android_getaddrinfofornet+,hn 13(0x35342e39332e31),sn(),hints(known),family 0,flags 4
D/libc﹕ [NET] android_getaddrinfofornet-, SUCCESS
I/System.out﹕ (INFO): TokenAuth.setTokenDetails():
I/System.out﹕ (INFO): TokenAuth.getTokenDetails():
I/System.out﹕ (VERBOSE): io.ably.transport.ConnectionManager: setState(): setting connecting
V/ABLY CONNECTION CHANGE﹕ connecting
D/libc﹕ [NET] android_getaddrinfofornet+,hn 16(0x7265616c74696d),sn(),hints(known),family 0,flags 4
D/libc﹕ [NET] android_getaddrinfofornet-, err=8
D/libc﹕ [NET] android_getaddrinfofornet+,hn 16(0x7265616c74696d),sn(),hints(known),family 0,flags 1024
D/libc﹕ [NET] android_getaddrinfofornet-, pass to proxy
D/libc﹕ [NET] android_getaddrinfo_proxy+
D/libc﹕ [NET] android_getaddrinfo_proxy get netid:0
D/libc﹕ [NET] android_getaddrinfo_proxy-, success
I/System.out﹕ (VERBOSE): io.ably.transport.ConnectionManager: notifyState(): notifying connected; id = KLHEl1vQMzN8mX9S
I/System.out﹕ (VERBOSE): io.ably.transport.ConnectionManager: setState(): setting connected
V/ABLY CONNECTION CHANGE﹕ connected
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: RealtimeChannel(); channel = mobile:chat
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: attach(); channel = mobile:chat
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: setState(): channel = mobile:chat; setting attaching
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: subscribe(); channel = mobile:chat
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: attach(); channel = mobile:chat
D/libc﹕ [NET] android_getaddrinfofornet+,hn 16(0x7265616c74696d),sn(),hints(known),family 0,flags 4
D/libc﹕ [NET] android_getaddrinfofornet-, err=8
D/libc﹕ [NET] android_getaddrinfofornet+,hn 13(0x35342e39332e31),sn(),hints(known),family 0,flags 4
D/libc﹕ [NET] android_getaddrinfofornet-, SUCCESS
I/System.out﹕ (INFO): TokenAuth.getCredentials():
I/System.out﹕ (INFO): TokenAuth.authorise():
I/System.out﹕ (INFO): TokenAuth.authorise(): using cached token; expires = 1440689014722
W/System.err﹕ io.ably.types.AblyException: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String java.security.Principal.getName()' on a null object reference
W/System.err﹕ at io.ably.http.Http.getForHost(Unknown Source)
W/System.err﹕ at io.ably.http.Http.get(Unknown Source)
W/System.err﹕ at io.ably.http.PaginatedQuery.get(Unknown Source)
W/System.err﹕ at io.ably.realtime.Channel.history(Unknown Source)
W/System.err﹕ at com.example.stan.myapplication.MainActivity$1.onConnectionStateChanged(MainActivity.java:61)
W/System.err﹕ at io.ably.realtime.Connection.apply(Unknown Source)
W/System.err﹕ at io.ably.realtime.Connection.apply(Unknown Source)
W/System.err﹕ at io.ably.util.EventEmitter.emit(Unknown Source)
W/System.err﹕ at io.ably.realtime.Connection.onConnectionStateChange(Unknown Source)
W/System.err﹕ at io.ably.transport.ConnectionManager.setState(Unknown Source)
W/System.err﹕ at io.ably.transport.ConnectionManager.handleStateChange(Unknown Source)
W/System.err﹕ at io.ably.transport.ConnectionManager.run(Unknown Source)
W/System.err﹕ Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String java.security.Principal.getName()' on a null object reference
W/System.err﹕ at org.apache.http.impl.auth.BasicScheme.authenticate(BasicScheme.java:166)
W/System.err﹕ at org.apache.http.impl.auth.BasicScheme.authenticate(BasicScheme.java:142)
W/System.err﹕ at org.apache.http.client.protocol.RequestTargetAuthentication.process(RequestTargetAuthentication.java:96)
W/System.err﹕ at org.apache.http.protocol.BasicHttpProcessor.process(BasicHttpProcessor.java:290)
W/System.err﹕ at org.apache.http.protocol.HttpRequestExecutor.preProcess(HttpRequestExecutor.java:160)
W/System.err﹕ at org.apache.http.impl.client.DefaultRequestDirector.execute(DefaultRequestDirector.java:420)
W/System.err﹕ at org.apache.http.impl.client.AbstractHttpClient.execute(AbstractHttpClient.java:754)
W/System.err﹕ ... 12 more
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: setAttached(); channel = mobile:chat
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: setState(): channel = mobile:chat; setting attached
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: sendQueuedMessages()
I/System.out﹕ (VERBOSE): io.ably.realtime.Channel: sendQueuedMessages()

```

FYI @mattheworiordan @paddybyers

gokhanbarisaker:
Resolves #41


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -254,43 +254,49 @@ public void run() {
...
+				case NACK:
+					onNack(message);
+					break;
+				default:
+					onChannelMessage(message);
+			}
+		}
+		catch(Exception e) {
+			// Prevent any non-AblyException to be thrown
+			throw (e instanceof AblyException?(AblyException)e:new AblyException(e));
```
paddybyers: We have `AblyException.fromThrowable()` for this.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -254,43 +254,49 @@ public void run() {
...
-		case HEARTBEAT:
-			onHeartbeat(message);
-			break;
-		case ERROR:
-			ErrorInfo reason = message.error;
-			if(reason == null)
-				Log.e(TAG, "onMessage(): ERROR message received (no error detail)");
-			else
-				Log.e(TAG, "onMessage(): ERROR message received; message = " + reason.message + "; code = " + reason.code);
+	void onMessage(ProtocolMessage message) throws AblyException {
```
paddybyers: I think this is a detail, but I'm not sure I want these exceptions to propagate back to the caller here.

There's a catch around the call in `WsClient`, but that catch was intended to be handling any deserialisation failure of the message. An error in processing a correctly deserialised message, I think, should be handled here.

You might make this explicit by changing the code in `WsClient` to perform only the decoding within the try/catch, remove the throws declaration here, and handle the exception in the catch at the bottom of this method.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
gokhanbarisaker: There are several other usages of `new AblyException(...)`. Would you like me to convert all of them to factory method, and make constructor accessor private?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
gokhanbarisaker: @paddybyers @mattheworiordan 

TL,DR, 

I am strongly advocating against handling errors that caused by non-client-caused reasons. (e.g., provided message isn't compatible with our protocol, breaking our secred wow between client and server)


---

I can handle the thrown exception within the `ConnectionManager#onMessage` scope, but before that I would like to mention couple of things.



> An error in processing a correctly deserialised message, I think, should be handled here.

Like you said `that catch was intended to be handling any deserialisation failure of the message`, and we don't have a deserialisation issue here. What we are trying to handle here is a message that is not compatible with our protocol. If we try to handle that (and whole possible non-client-caused errors that might occur) using either,

  - validation, which introduces following side-effects
    - boilerplate code. Proof of concept, https://github.com/gokhanbarisaker/ably-java/commit/ba1564ca24e3b85af716316294c09de8f5210908
    - performance overhead, (this is a blind guess made without profiling code, actually a little arrogant of me to add it here without it :))
    - a little bit garbage creation.

  - try/catch blocks, which introduces following side-effects
    - lots of boilerplate code,
    - performance overhead (this is a blind guess made without profiling code, actually a little arrogant of me to add it here without it :))
    - although current handling for unchecked exceptions on upper scope just logs the data, trying to handle that specific exception can be easily confused with others.


We probably end up violating KISS principle, and arguably introduce over engineering. Also above side-effects multiply considering out client will process high throughput during realtime messaging.



If you are still OK with the things above, I am comfortable with handling the exception within current scope :)

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: Makes sense to me.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: Over to you @paddybyers as I don't have a strong opinion on this especially because it's Java

paddybyers: @gokhanbarisaker: overall, I agree: I don't want to add micro-level validation throughout (especially in the manner of your Monad Validator which is IMO needlessly complex and verbose and also creates unnecessary garbage) in order to validate the input from our own server.

However, even when the client library and server are working perfectly, things can go wrong: for example some cellular networks transcode JSON to HTML (!) if you're not running over TLS. Therefore it is possible that we will get deserialisation failures for messages (this is most likely) and if deserialisation succeeds it is possible we will get protocol errors. These errors don't need to be diagnosed but should at least be handled so we don't end up with exceptions propagating up to kill threads or whatever.

So I think the code in `WsClient` should have a try/catch around the deserialisation; if an error is detected it should be logged as protocol error, and the transport disconnected. Similarly any exception arising in this scope here should I think be caught, and logged as a protocol error with the transport disconnected.

WDYT?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: > Would you like me to convert all of them to factory method, and make constructor accessor private?

Personally I don't think it really adds value to have a static factory method if the constructor with the same arguments has the same (and obvious) meaning. However, it would make things a bit more consistent so we should probably do it.

gokhanbarisaker:
Import replacement (which seems ridiculus at first sight) was necessary for https://github.com/gokhanbarisaker/ably-java/commit/11d8d9eb6a68bc9ecfb75ffc0478f6299b26bbbd

This is the result of IntelliJ `Organize imports command` and one time modification to be compatible with IDE.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
gokhanbarisaker: > Therefore it is possible that we will get deserialisation failures for messages (this is most likely) and if deserialisation succeeds it is possible we will get protocol errors.

Which-ever network do that kind of messing with data (not corrupting, but modifying) and saying `Deal with it` should be root of all evil :). If that is really the case, you are most certainly right to do extra pre-caution on client side.

I will start working on that starting tomorrow.

paddybyers:
> This is the result of IntelliJ Organize imports command and one time modification to be compatible with IDE.

OK, previously it was based on what Eclipse `organise imports` was doing :)

lib/src/main/java/io/ably/lib/types/AblyException.java
```diff
@@ -13,86 +11,17 @@
...
-	public AblyException(Throwable cause) {
-		super(cause);
-		errorInfo = new ErrorInfo("Unexpected exception: " + cause.getLocalizedMessage(), 50000, 500);
-	}
-
-	/**
 	 * Constructor for use where there is an ErrorInfo available
 	 */
-	public AblyException(ErrorInfo reason) {
+	AblyException(Throwable throwable, ErrorInfo reason) {
```
paddybyers: I don't get why this argument is here

lib/src/main/java/io/ably/lib/types/AblyException.java
```diff
@@ -103,9 +32,10 @@ public static AblyException fromResponseStatus(String statusLine, int statusCode
 	public static AblyException fromThrowable(Throwable t) {
 		if(t instanceof AblyException)
 			return (AblyException)t;
-		if(t instanceof IOException)
-			return fromIOException((IOException)t);
-		return new AblyException(t);
+		if(t instanceof ConnectException || t instanceof UnknownHostException || t instanceof NoRouteToHostException)
```
paddybyers: Is this syntax compatible with all Android versions we want to support?

lib/src/main/java/io/ably/lib/types/Capability.java
```diff
@@ -1,17 +1,12 @@
 package io.ably.lib.types;
 
 
+import com.google.gson.*;
```
paddybyers: So IntelliJ forces you to use a wildcard import statement?

mattheworiordan:
> I think it would be a lower priority set of changes than the other functionality and stability-related changes that need to be done.

Agreed we need to focus on the higher priority items outlined in the spreadsheet.  @paddybyers are you happy for this to be merged though, or are there issues in this PR that need addressing first? Perhaps we could tackle those quickly and get onto more important issues?

lib/src/main/java/io/ably/lib/types/Capability.java
gokhanbarisaker: It seems like it

lib/src/main/java/io/ably/lib/types/AblyException.java
gokhanbarisaker: My bad, `throwable` needs to be passed down to super constructor.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -209,18 +202,6 @@ public void finalize() {
 		dispose();
 	}
 
-	private static class HostFailedException extends AblyException {
-		private static final long serialVersionUID = 1L;
-		public HostFailedException(Throwable cause) {
-			super(cause);
-		}		
-		private static AblyException checkFor(Throwable t) {
-			if(t instanceof ConnectException || t instanceof UnknownHostException || t instanceof NoRouteToHostException)
```
paddybyers: I preferred this, compared with your proposed change, because here `HostFailedException` knows whether or not a given `Throwable` applies. With your change, that decision is now distant from the declaration of the class.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -236,7 +217,7 @@ private static AblyException checkFor(Throwable t) {
 					} catch(HostFailedException bhe2) {}
```
paddybyers: Is this an auto-generated identifier? Can't it be `hfe` ?

paddybyers:
> are you happy for this to be merged though, or are there issues in this PR that need addressing first?

I think the intent behind the changes here is good and I'm happy to merge once the comments above have been addressed please;, or we go for a very much simpler change which just introduces a like-for-like static factory method for the constructor without other changes.

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: I have moved the `HostFailedException` outside `Http` for it was forcing parent class `AblyException` constructor to be public.

Also, I have moved the creation logic to single factory method for a class deciding whether itself or its super class going to be created seemed like an anti-pattern to me. It is like expecting `String.valueOf("foo".toCharArray());` to return `CharSequence` instance depending on given argument.

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: We are throwing AblyException on line 220.

```java
throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
```

lib/src/main/java/io/ably/lib/http/Http.java
paddybyers: > I have moved the HostFailedException outside Http for it was forcing parent class AblyException constructor to be public.

That's fine.

> Also, I have moved the creation logic to single factory method for a class deciding whether itself or its super class going to be created seemed like an anti-pattern to me.

I understand that, but I don't like the way that `HostFailedException` is declared in one place, but the logic that decides what it is is somewhere else.

I would be happier perhaps with `HostFailedException` perhaps being a static inner class of `AblyException`.

mattheworiordan:
> I think the intent behind the changes here is good and I'm happy to merge once the comments above have been addressed please

@gokhanbarisaker you happy to quickly address these comments so that we can close this PR?

gokhanbarisaker:
@mattheworiordan I have already replied them. Only pending change is moving HostFailedException into Ably Exception. I will do that tonight

paddybyers:
> I will do that tonight

thanks

gokhanbarisaker:
@mattheworiordan @paddybyers  Although the latest previous build passes, we are receiving flaky results from serverside. i.e., for the same commit following builds are,

Passing: https://travis-ci.org/ably/ably-java/builds/107422175
Failing: https://travis-ci.org/gokhanbarisaker/ably-java/builds/107422172

paddybyers:
> lthough the latest build passed, we are receiving flaky results from serverside. i.e., for the same commit following builds are,

The sandbox environment was in an odd state so I've redeployed it just now. Let me know if there are other issues.

paddybyers:
Thanks

Issue #54, opened 2015-12-09
Host fallback for REST


mattheworiordan:
It's possible our host fallback is not working correctly following a chat with a customer who is experiencing issues.  They have confirmed at the time they are unable to curl to rest.ably.io, however I would then expect the client to route to a different data centre.

See error below:

```
2015-12-03 08:10:02.905 ERROR 11292 --- [       Timer-26] messaging.MessengerImpl                  : Connection failed; no host available
io.ably.types.AblyException: Connection failed; no host available
	at io.ably.http.Http.post(Unknown Source)
	at io.ably.rest.Channel.publish(Unknown Source)
	at messaging.AblyMessagingRestImpl.sendMessage(AblyMessagingRestImpl.java:31)
	at demo.SchedulerCallable.sendLocationMessage(SchedulerCallable.java:186)
	at demo.SchedulerCallable.process(SchedulerCallable.java:51)
	at demo.SchedulerCallable$1.run(SchedulerCallable.java:80)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
.(INFO): io.ably.rest.AblyRest: started
(INFO): Auth(): anonymous, using basic auth
```

mattheworiordan:
Related to https://github.com/ably/ably-java/issues/14

Issue #14, opened 2015-04-18
Token expiry causes alternative host names to be used


mattheworiordan:
If I have read the code correctly, when a request is made to the default end point of `rest.ably.io` and a status code outside of 200-299 is returned, then a fallback host is used to make that request again.

However, if a token has expired (code 40140), that too will trigger it to reissue the request to a fallback host, when what should happen is the token should be reissued if possible or it should fail immediately as retrying with fallback hosts is pointless.

See https://github.com/ably/ably-ruby/blob/master/lib/ably/rest/client.rb#L334-L347 in the Ruby library.

Have I misread it?

mattheworiordan:
In fact, we should probably review which codes at https://github.com/ably/ably-common/blob/master/protocol/errors.json should trigger a retry.  Lots of them such as credential failures, invalid body, max message size exceeded etc. don't need to retry IMHO.

paddybyers:
> Have I misread it?

I think so.

The fallback hosts are used if a `HostFailedException` is thrown, which only occurs when the problem was a `ConnectException`, `UnknownHostException`, or `NoRouteToHostException`.

mattheworiordan:
@paddybyers but if the server responds with a 500 error code shouldn't we be using the fallback hosts?  For example, if the load balancers are unable to service a request, the client will get a response in the 5xx range.

paddybyers:
> if the server responds with a 500 error code shouldn't we be using the fallback hosts?

The fallback hosts are only for the case that the server is unreachable. By definition, if we get any error code from the server, then it's the correct host to use.

mattheworiordan:
I disagree.  The fallback hosts are to route around network issues or a faulty or overloaded data centre.  If a data centre becomes unhealthy for any reason, but the load balancers are still operating, all clients will continue to use the faulty data centre.  We don't want this.

mattheworiordan:
Part of our USP is our global presence and distributed architecture.  The client libraries should automatically distribute load and move traffic to other working data centres when a single data centre for whatever reason is unable to cope with the current load.

mattheworiordan:
@kouno and I discussed and think that the following strategy would be sensible:

* If a request fails because of network errors (see https://github.com/ably/ably-java/blob/0275671edd98e7090181696ed6dd34a01cbbd96f/src/io/ably/http/Http.java#L327), then we fallback immediately as we do
* If a request fails because of a 5xx type error, then we retry one more time, and then fallback.

This could apply to both REST & Realtime.

PR #80, opened 2016-01-29
Add httpMaxRetryCount && Simplify http fallback flow

gokhanbarisaker:
Resolves: #54, #14

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -5,22 +5,19 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Field;
-import java.net.ConnectException;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.NoRouteToHostException;
-import java.net.URL;
-import java.net.UnknownHostException;
+import java.net.*;
+import java.nio.charset.Charset;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
 
-import io.ably.lib.realtime.AblyRealtime;
 import io.ably.lib.realtime.Connection;
-import io.ably.lib.realtime.ConnectionState;
-import io.ably.lib.rest.AblyRest;
 import io.ably.lib.rest.Auth;
 import io.ably.lib.rest.Auth.AuthMethod;
 import io.ably.lib.transport.Defaults;
+import io.ably.lib.transport.Hosts;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
 import io.ably.lib.types.ErrorInfo;
@@ -41,16 +38,65 @@ import io.ably.lib.util.Serialisation;
 public class Http {
 
 	public interface ResponseHandler<T> {
-		public T handleResponse(int statusCode, String contentType, Collection<String> linkHeaders, byte[] body) throws AblyException;
+		T handleResponse(int statusCode, String contentType, Collection<String> linkHeaders, byte[] body) throws AblyException;
 	}
 
 	public interface BodyHandler<T> {
-		public T[] handleResponseBody(String contentType, byte[] body) throws AblyException;
+		T[] handleResponseBody(String contentType, byte[] body) throws AblyException;
 	}
 
 	public interface RequestBody {
-		public byte[] getEncoded();
-		public String getContentType();
+		byte[] getEncoded();
+		String getContentType();
+	}
+
+	private static class Response {
+		int statusCode;
+		String statusLine;
+		Map<String,List<String>> headers;
+		String contentType;
+		int contentLength;
+		byte[] body;
+
+		/**
+		 * Returns the value of the named header field.
+		 * <p>
+		 * If called on a connection that sets the same header multiple times
+		 * with possibly different values, only the last value is returned.
+		 *
+		 *
+		 * @param   name   the name of a header field.
+		 * @return  the value of the named header field, or {@code null}
+		 *          if there is no such field in the header.
+		 */
+		public String getHeaderField(String name) {
+			List<String> fields = getHeaderFields(name);
+
+			if (fields == null || fields.isEmpty()) {
+				return null;
+			}
+
+			return fields.get(fields.size() - 1);
+		}
+
+		/**
+		 * Returns the value of the named header field.
+		 * <p>
+		 * If called on a connection that sets the same header multiple times
+		 * with possibly different values, only the last value is returned.
+		 *
+		 *
+		 * @param   name   the name of a header field.
+		 * @return  the value of the named header field, or {@code null}
+		 *          if there is no such field in the header.
+		 */
+		public List<String> getHeaderFields(String name) {
+			if(headers == null) {
+				return null;
+			}
+
+			return headers.get(name.toLowerCase());
+		}
 	}
 
 	public static class JSONRequestBody implements RequestBody {
@@ -81,19 +127,30 @@ public class Http {
 	 *     Public API
 	 *************************/
 
-	public Http(AblyRest ably, ClientOptions options) {
-		this.ably = ably;
+	public Http(ClientOptions options, Auth auth) {
+		this.options = options;
+		this.auth = auth;
+		this.host = options.restHost;
 		this.scheme = options.tls ? "https://" : "http://";
 		this.port = Defaults.getPort(options);
 	}
 
-	private String getPrefHost() {
-		if(ably instanceof AblyRealtime) {
-			Connection connection = ((AblyRealtime)ably).connection;
-			if(connection.state == ConnectionState.connected)
-				return connection.connectionManager.getHost();
-		}
-		return Defaults.getHost(ably.options);
+	/**
+	 * Sets host for this HTTP client
+	 *
+	 * @param host URL string
+	 */
+	public void setHost(String host) {
+		this.host = host;
+	}
+
+	/**
+	 * Gets host for this HTTP client
+	 *
+	 * @return
+     */
+	public String getHost() {
+		return host;
 	}
 
 	/**
@@ -185,7 +242,6 @@ public class Http {
 		if(authHeader != null && !renew) {
 			return authHeader;
 		}
-		Auth auth = ably.auth;
 		if(auth.getAuthMethod() == AuthMethod.basic) {
 			authHeader = "Basic " + Base64Coder.encodeString(auth.getBasicCredentials());
 		} else {
@@ -210,37 +266,38 @@ public class Http {
 	}
 
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
-		try {
-			URL url = buildURL(scheme, getPrefHost(), path, params);
-			return httpExecute(url, method, headers, requestBody, true, responseHandler);
-		} catch(AblyException.HostFailedException bhe) {
-			/* one of the exceptions occurred that signifies a problem reaching the host */
-			String[] fallbackHosts = Defaults.getFallbackHosts(ably.options);
-			if(fallbackHosts != null) {
-				for(String host : fallbackHosts) {
-					try {
-						URL url = buildURL(scheme, host, path, params);
-						return httpExecute(url, method, headers, requestBody, true, responseHandler);
-					} catch(AblyException.HostFailedException bhe2) {}
+		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host)?options.httpMaxRetryCount:0;
+		String hostCurrent = this.host;
+		URL url;
+
+		do {
+			url = buildURL(scheme, hostCurrent, path, params);
+
+			try {
+				return httpExecute(url, method, headers, requestBody, true, responseHandler);
+			} catch (AblyException.HostFailedException e) {
+				retryCountRemaining--;
+
+				if (retryCountRemaining >= 0) {
+					Log.d(TAG, "Connection failed to host `" + hostCurrent + "`. Searching for new host...");
+					hostCurrent = Hosts.getFallback(hostCurrent);
+					Log.d(TAG, "Switched to `" + hostCurrent + "`.");
 				}
 			}
-			throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
-		}
+		} while (retryCountRemaining >= 0);
+
+		throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
 	}
 
 	<T> T httpExecute(URL url, String method, Param[] headers, RequestBody requestBody, boolean withCredentials, ResponseHandler<T> responseHandler) throws AblyException {
+		Response response;
 		HttpURLConnection conn = null;
-		InputStream is = null;
-		int statusCode = 0;
-		String statusLine = null;
-		String contentType = null;
-		byte[] responseBody = null;
-		List<String> linkHeaders = null;
 		boolean credentialsIncluded = false;
 		try {
-
 			/* prepare connection */
 			conn = (HttpURLConnection)url.openConnection();
+			conn.setConnectTimeout(options.httpOpenTimeout);
+			conn.setReadTimeout(options.httpRequestTimeout);
 			conn.setDoInput(true);
 			if(method != null) {
 				conn.setRequestMethod(method);
@@ -260,62 +317,30 @@ public class Http {
 
 			/* send request body */
 			if(requestBody != null) {
-				conn.setDoOutput(true);
-				byte[] body = requestBody.getEncoded();
-				int length = body.length;
-				conn.setFixedLengthStreamingMode(length);
-				conn.setRequestProperty(CONTENT_TYPE, requestBody.getContentType());
-				conn.setRequestProperty(CONTENT_LENGTH, Integer.toString(length));
-				OutputStream os = conn.getOutputStream();
-				os.write(body);
+				writeRequestBody(requestBody, conn);
 			}
 
-			/* get response */
-			statusCode = conn.getResponseCode();
-			statusLine = conn.getResponseMessage();
-			if(statusCode != HttpURLConnection.HTTP_NO_CONTENT) {
-				contentType = conn.getContentType();
-				int contentLength = conn.getContentLength();
-				int successStatusCode = (method == POST) ? HttpURLConnection.HTTP_CREATED : HttpURLConnection.HTTP_OK;
-				is = (statusCode == successStatusCode) ? conn.getInputStream() : conn.getErrorStream();
-				if(is != null) {
-					int read;
-					if(contentLength == -1) {
-						ByteArrayOutputStream baos = new ByteArrayOutputStream();
-						byte[] buf = new byte[4 * 1024];
-						while((read = is.read(buf)) > -1) {
-							baos.write(buf, 0, read);
-						}
-						responseBody = baos.toByteArray();
-					} else {
-						int idx = 0;
-						responseBody = new byte[contentLength];
-						while((read = is.read(responseBody,  idx, contentLength - idx)) > -1) {
-							idx += read;
-						}
-					}
-				}
-			}
+			response = readResponse(conn);
 		} catch(IOException ioe) {
 			throw AblyException.fromThrowable(ioe);
 		} finally {
-			try {
-				if(is != null)
-					is.close();
-				if(conn != null)
-					conn.disconnect();
-			} catch(IOException ioe) {}
+			if(conn != null)
+				conn.disconnect();
 		}
 
-		if(statusCode == 0) {
+		if (response.statusCode == 0) {
 			return null;
 		}
 
-		if(statusCode < 200 || statusCode >= 300) {
+		if (response.statusCode >=500 && response.statusCode <= 504) {
+			throw AblyException.fromErrorInfo(ErrorInfo.fromResponseStatus(response.statusLine, response.statusCode));
+		}
+
+		if(response.statusCode < 200 || response.statusCode >= 300) {
 			/* get any in-body error details */
 			ErrorInfo error = null;
-			if(responseBody != null && responseBody.length > 0) {
-				ErrorResponse errorResponse = ErrorResponse.fromJSON(new String(responseBody));
+			if(response.body != null && response.body.length > 0) {
+				ErrorResponse errorResponse = ErrorResponse.fromJSON(new String(response.body));
 				if(errorResponse != null) {
 					error = errorResponse.error;
 				}
@@ -327,14 +352,14 @@ public class Http {
 				String errorMessageHeader = conn.getHeaderField("X-Ably-ErrorMessage");
 				if(errorCodeHeader != null) {
 					try {
-						error = new ErrorInfo(errorMessageHeader, statusCode, Integer.parseInt(errorCodeHeader));
+						error = new ErrorInfo(errorMessageHeader, response.statusCode, Integer.parseInt(errorCodeHeader));
 					} catch(NumberFormatException e) {}
 				}
 			}
 
 			/* handle www-authenticate */
-			if(statusCode == 401) {
-				String wwwAuthHeader = conn.getHeaderField(WWW_AUTHENTICATE);
+			if(response.statusCode == 401) {
+				String wwwAuthHeader = response.getHeaderField(WWW_AUTHENTICATE);
 				if(wwwAuthHeader != null) {
 					boolean stale = (wwwAuthHeader.indexOf("stale") > -1) || (error != null && error.code == 40140);
 					if(withCredentials && (stale || !credentialsIncluded)) {
@@ -348,8 +373,8 @@ public class Http {
 				Log.e(TAG, "Error response from server: " + error);
 				throw AblyException.fromErrorInfo(error);
 			} else {
-				Log.e(TAG, "Error response from server: statusCode = " + statusCode + "; statusLine = " + statusLine);
-				throw AblyException.fromErrorInfo(ErrorInfo.fromResponseStatus(statusLine, statusCode));
+				Log.e(TAG, "Error response from server: statusCode = " + response.statusCode + "; statusLine = " + response.statusLine);
+				throw AblyException.fromErrorInfo(ErrorInfo.fromResponseStatus(response.statusLine, response.statusCode));
 			}
 		}
 
@@ -357,8 +382,87 @@ public class Http {
 			return null;
 		}
 
-		linkHeaders = conn.getHeaderFields().get(LINK);
-		return responseHandler.handleResponse(statusCode, contentType, linkHeaders, responseBody);
+		List<String> linkHeaders = response.getHeaderFields(LINK);
+		return responseHandler.handleResponse(response.statusCode, response.contentType, linkHeaders, response.body);
+	}
+
+	private void writeRequestBody(RequestBody requestBody, HttpURLConnection conn) throws IOException {
+		conn.setDoOutput(true);
+		byte[] body = requestBody.getEncoded();
+		int length = body.length;
+		conn.setFixedLengthStreamingMode(length);
+		conn.setRequestProperty(CONTENT_TYPE, requestBody.getContentType());
+		conn.setRequestProperty(CONTENT_LENGTH, Integer.toString(length));
+		OutputStream os = conn.getOutputStream();
+		os.write(body);
+	}
+
+	private Response readResponse(HttpURLConnection connection) throws IOException {
+		Response response = new Response();
+		response.statusCode = connection.getResponseCode();
+		response.statusLine = connection.getResponseMessage();
+
+		/* Store all header field names in lower-case to eliminate case insensitivity */
+		Map<String, List<String>> caseSensitiveHeaders = connection.getHeaderFields();
+		response.headers = new HashMap<>(caseSensitiveHeaders.size(), 1f);
+
+		for (Map.Entry<String, List<String>> entry : caseSensitiveHeaders.entrySet()) {
+			if (entry.getKey() != null) {
+				response.headers.put(entry.getKey().toLowerCase(), entry.getValue());
+			}
+		}
+
+		if(response.statusCode == HttpURLConnection.HTTP_NO_CONTENT) {
+			return response;
+		}
+
+		response.contentType = connection.getContentType();
+		response.contentLength = connection.getContentLength();
+
+		int successStatusCode = (POST.equals(connection.getRequestMethod())) ? HttpURLConnection.HTTP_CREATED : HttpURLConnection.HTTP_OK;
+		InputStream is = (response.statusCode == successStatusCode) ? connection.getInputStream() : connection.getErrorStream();
+
+		try {
+			response.body = readInputStream(is, response.contentLength);
+		} catch (NullPointerException e) {
+			/* nothing to read */
+		} finally {
+			if (is != null) {
+				try {
+					is.close();
+				} catch (IOException e) {}
+			}
+		}
+
+		return response;
+	}
+
+	private byte[] readInputStream(InputStream inputStream, int bytes) throws IOException {
+		/* If there is nothing to read */
+		if (inputStream == null) {
+			throw new NullPointerException("inputStream == null");
+		}
+
+		int bytesRead = 0;
+
+		if (bytes == -1) {
+			ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+			byte[] buffer = new byte[4 * 1024];
+			while((bytesRead = inputStream.read(buffer)) > -1) {
+				outputStream.write(buffer, 0, bytesRead);
+			}
+
+			return outputStream.toByteArray();
+		}
+		else {
+			int idx = 0;
+			byte[] output = new byte[bytes];
+			while((bytesRead = inputStream.read(output,  idx, bytes - idx)) > -1) {
+				idx += bytesRead;
+			}
+
+			return output;
+		}
 	}
 
 	private void appendParams(StringBuilder uri, Param[] params) {
@@ -392,6 +496,7 @@ public class Http {
 		return result;
 	}
 
+
 	/*************************
 	 *     Private state
 	 *************************/
@@ -410,20 +515,22 @@ public class Http {
 		}
 	}
 
-	private final AblyRest ably;
+	private final ClientOptions options;
+	private final Auth auth;
 	private final String scheme;
+	private String host;
 	private final int port;
 	private String authHeader;
 	private boolean isDisposed;
 
-	private static final String TAG              = Http.class.getName();
-	private static final String LINK             = "Link";
-	private static final String ACCEPT           = "Accept";
-	private static final String CONTENT_TYPE     = "Content-Type";
-	private static final String CONTENT_LENGTH   = "Content-Length";
-	private static final String JSON             = "application/json";
-	private static final String WWW_AUTHENTICATE = "WWW-Authenticate";
-	private static final String AUTHORIZATION    = "Authorization";
+	private static final String TAG               = Http.class.getName();
+	private static final String LINK              = "Link";
+	private static final String ACCEPT            = "Accept";
+	private static final String CONTENT_TYPE      = "Content-Type";
+	private static final String CONTENT_LENGTH    = "Content-Length";
+	private static final String JSON              = "application/json";
+	private static final String WWW_AUTHENTICATE  = "WWW-Authenticate";
+	private static final String AUTHORIZATION     = "Authorization";
 
 	static final String GET                      = "GET";
 	static final String POST                     = "POST";
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -5,6 +5,7 @@ import java.util.Iterator;
 import java.util.Map;
 
 import io.ably.lib.rest.AblyRest;
+import io.ably.lib.transport.Defaults;
 import io.ably.lib.transport.ITransport;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ChannelOptions;
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -21,7 +21,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Random;
+
 
 public class ConnectionManager implements Runnable, ConnectListener {
 
@@ -51,10 +51,16 @@ public class ConnectionManager implements Runnable, ConnectListener {
 	public static class StateIndication {
 		final ConnectionState state;
 		final ErrorInfo reason;
-		boolean useFallbackHost;
+		final boolean fallback;
+
 		public StateIndication(ConnectionState state, ErrorInfo reason) {
+			this(state, reason, false);
+		}
+
+		public StateIndication(ConnectionState state, ErrorInfo reason, boolean fallback) {
 			this.state = state;
 			this.reason = reason;
+			this.fallback = fallback;
 		}
 	}
 	
@@ -91,11 +97,11 @@ public class ConnectionManager implements Runnable, ConnectListener {
 	@SuppressWarnings("serial")
 	public static final HashMap<ConnectionState, StateInfo> states = new HashMap<ConnectionState, StateInfo>() {{
 		put(ConnectionState.initialized, new StateInfo(ConnectionState.initialized, true, false, false, false, 0, null));
-		put(ConnectionState.connecting, new StateInfo(ConnectionState.connecting, true, false, false, false, Defaults.connectTimeout, null));
+		put(ConnectionState.connecting, new StateInfo(ConnectionState.connecting, true, false, false, false, Defaults.TIMEOUT_CONNECT, null));
 		put(ConnectionState.connected, new StateInfo(ConnectionState.connected, false, true, false, false, 0, null));
-		put(ConnectionState.disconnected, new StateInfo(ConnectionState.disconnected, true, false, false, true, Defaults.disconnectTimeout, REASON_DISCONNECTED));
-		put(ConnectionState.suspended, new StateInfo(ConnectionState.suspended, false, false, false, true, Defaults.suspendedTimeout, REASON_SUSPENDED));
-		put(ConnectionState.closing, new StateInfo(ConnectionState.closing, false, false, false, false, Defaults.connectTimeout, REASON_CLOSED));
+		put(ConnectionState.disconnected, new StateInfo(ConnectionState.disconnected, true, false, false, true, Defaults.TIMEOUT_DISCONNECT, REASON_DISCONNECTED));
+		put(ConnectionState.suspended, new StateInfo(ConnectionState.suspended, false, false, false, true, Defaults.TIMEOUT_SUSPEND, REASON_SUSPENDED));
+		put(ConnectionState.closing, new StateInfo(ConnectionState.closing, false, false, false, false, Defaults.TIMEOUT_CONNECT, REASON_CLOSED));
 		put(ConnectionState.closed, new StateInfo(ConnectionState.closed, false, false, true, false, 0, REASON_CLOSED));
 		put(ConnectionState.failed, new StateInfo(ConnectionState.failed, false, false, true, false, 0, REASON_FAILED));
 	}};
@@ -119,7 +125,7 @@ public class ConnectionManager implements Runnable, ConnectListener {
 		queuedMessages = new ArrayList<QueuedMessage>();
 		pendingMessages = new PendingMessageQueue();
 		state = states.get(ConnectionState.initialized);
-		String transportClass = Defaults.transport;
+		String transportClass = Defaults.TRANSPORT;
 		/* debug options */
 		if(options instanceof DebugOptions)
 			protocolListener = ((DebugOptions)options).protocolListener;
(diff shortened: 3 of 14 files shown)


gokhanbarisaker:
Test cases for that will require utilising `HostFailedException`. It will be available once we resolve #74

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -404,6 +395,19 @@ private URL buildURL(String uri, Param[] params) {
 		return result;
 	}
 
+	/**
+	 * TODO: Fill javadoc
```
paddybyers: If this is private we don't necessarily need a javadoc comment, but please either add a comment or remove the TODO. It's not worth leaving like this.

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -2,9 +2,12 @@
 
 import io.ably.lib.types.ClientOptions;
 
+import java.util.Arrays;
+import java.util.List;
+
 public class Defaults {
 	public static final int protocolVersion     = 1;
-	public static final String[] FALLBACK_HOSTS = new String[] {"A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com"};
+	public static final List<String> FALLBACK_HOSTS = Arrays.asList("A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com");
```
paddybyers: Have we lost the alignment here?

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -12,42 +15,15 @@
 	public static final int connectTimeout      = 15000;
 	public static final int disconnectTimeout   = 30000;
 	public static final int suspendedTimeout    = 120000;
-	public static final int cometRecvTimeout    = 90000;
-	public static final int cometSendTimeout    = 10000;
+	public static final int cometRecvTimeout    = 90000; /* FIXME: Should we remove this unused variable? */
```
paddybyers: Yes, please remove.

lib/src/main/java/io/ably/lib/transport/Defaults.java
gokhanbarisaker: List gives us useful apis without sacrificing performance considering we just have 5 items.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -81,21 +77,14 @@
...
-		this.ably = ably;
+	public Http(ClientOptions options, Auth auth) {
+		this.options = options;
+		this.auth = auth;
+		this.activeHost = options.restHost;
 		this.scheme = options.tls ? "https://" : "http://";
 		this.port = Defaults.getPort(options);
 	}
 
-	private String getPrefHost() {
```
paddybyers: The reason for this was that we would prefer to use the same host for REST requests as we are connected to, in the case of a realtime library. I would prefer that we retain this, because if it uses a different fallback host it is likely to be in a different region, which causes problems for certain requests.

paddybyers:
I like this, but please see the comments above.

lib/src/main/java/io/ably/lib/transport/Defaults.java
paddybyers: My comment was about whitespace.

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: I think `Http` class only initialized within `AblyRest` class. Therefore, it seems like that `instanceof` check will never be true

paddybyers: but `AblyRealtime extends AblyRest`, so I think in the realtime case it is true.

gokhanbarisaker: I must have missed that. I will bring that back asap

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -563,19 +559,11 @@ public synchronized void onTransportUnavailable(ITransport transport, TransportP
...
-			this.fallback = fallback;
 			this.connectionKey = connection.key;
 			this.connectionSerial = String.valueOf(connection.serial);
-			String[] fallbackHosts;
-			if(fallback && (fallbackHosts = Defaults.getFallbackHosts(options)).length > 0) {
-				fallbackHosts = Defaults.getFallbackHosts(options);
-				this.host = fallbackHosts[random.nextInt(fallbackHosts.length)];
-			} else {
-				this.host = Defaults.getHost(options, host, true);
-			}
```
mattheworiordan: Am I missing something here, as far as I can tell we have removed the fallback host functionality fro Realtime connections in this commit?  The [Random class](https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L825) is not even used anymore, and the connection implementation still references fallback hosts yet does not use them, see https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L574-L578

paddybyers:
> we don't necessarily default to the last successful fallback host

The previous functionality wasn't reverting to the last successful host for an HTTP request. It was saying that if this is a realtime instance, **and it is currently connected**, then that host is the preferred host.

mattheworiordan:
>  It was saying that if this is a realtime instance, and it is currently connected, then that host is the preferred host.

The [`activeHost` is set](https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/http/Http.java#L83) when the [Http class is instanced for the REST lib](https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/rest/AblyRest.java#L71). As far as I can tell, [every REST request then uses `activeHost`](https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/http/Http.java#L218), and if that request fails just once, [`activeHost` is set to a random fallback host](https://github.com/gokhanbarisaker/ably-java/blob/8b59f05590e123462f2603ffccdd9a696f4e0434/lib/src/main/java/io/ably/lib/http/Http.java#L225) for all remaining requests.  Am I missing something, what has the realtime connection got to do with that?

gokhanbarisaker:
> The activeHost is set when the Http class is instanced for the REST lib. As far as I can tell, every REST request then uses activeHost, and if that request fails just once, activeHost is set to a random fallback host for all remaining requests. Am I missing something, what has the realtime connection got to do with that?

I have already updated that behavior with [latest commit](https://github.com/gokhanbarisaker/ably-java/commit/c2e636434fa857c09554be476a149b70fd029e08). Now [we don't have persistent fallback for host variable](https://github.com/gokhanbarisaker/ably-java/commit/c2e636434fa857c09554be476a149b70fd029e08#diff-8a8a56ece4e68a2d0fbeb6639c2e5469R222).

mattheworiordan:
As far as I can tell from this PR, the following is not being done correctly:

* Preferred REST host when Realtime is connected no longer works as [you removed it](https://github.com/ably/ably-java/pull/80/files#r51344160)
* All REST requests from the Realtime library (such as Channel history) now always use the incorrect Realtime endpoint, see https://github.com/ably/ably-java/pull/80/files#diff-b8b1ac0a34230df724c2576f32005521R54
* You are [only catching HostFailedException](https://github.com/ably/ably-java/pull/80/files#diff-8a8a56ece4e68a2d0fbeb6639c2e5469R233) for fallback attempts, but this is not broad enough for a rest request retry, see http://docs.ably.io/client-lib-development-guide/features/#RSC15d. [Same applies for realtime](https://github.com/ably/ably-java/pull/80/files#diff-060685a947a7032665de7ac9d33bb6ffR484)
* A style concern admittedly, but I really dislike evaluating variables that include a postfix or prefix operation, see https://github.com/ably/ably-java/pull/80/files#diff-8a8a56ece4e68a2d0fbeb6639c2e5469R228.  Could we not decrement the retry variable explicitly when we make the attempt to simplify readability of this code?
* My comment about inconsistent use of [upper case as well as missing underscores is still not addressed](https://github.com/ably/ably-java/pull/80/files#diff-7ce091f2af6c350f703933730f4b9245R21)
* Setting `httpMaxRetryCount` in all tests seems unnecessary, see https://github.com/ably/ably-java/pull/80/files#diff-d86628d178e27c35757c30abecfa69f3R75. I don't think you responded to that one?
* Not sure how `FALLBACKS` could ever be empty [given it's static and final](https://github.com/ably/ably-java/pull/80/files#diff-7ce091f2af6c350f703933730f4b9245R11), see https://github.com/ably/ably-java/pull/80/files#diff-633aa103b8a41039ef0c67f40e6a610aR21 and https://github.com/ably/ably-java/pull/80/files#diff-060685a947a7032665de7ac9d33bb6ffR486

BTW. Test coverage is looking good :)

gokhanbarisaker:
> Preferred REST host when Realtime is connected no longer works as you removed it

I am setting it explicitly on AblyRealtime constructor. https://github.com/ably/ably-java/pull/80/files#diff-b8b1ac0a34230df724c2576f32005521R54


> All REST requests from the Realtime library (such as Channel history) now always use the incorrect Realtime endpoint, see https://github.com/ably/ably-java/pull/80/files#diff-b8b1ac0a34230df724c2576f32005521R54

Done, https://github.com/gokhanbarisaker/ably-java/commit/9ede727b9e6a0f06e1eb6045c929c8a35d8136f2


> You are only catching HostFailedException for fallback attempts, but this is not broad enough for a rest request retry, see http://docs.ably.io/client-lib-development-guide/features/#RSC15d. Same applies for realtime

Done, https://github.com/gokhanbarisaker/ably-java/commit/b21c73ce6ff83fb9c7bd3c6c7db504cb94ea8e7f


> A style concern admittedly, but I really dislike evaluating variables that include a postfix or prefix operation, see https://github.com/ably/ably-java/pull/80/files#diff-8a8a56ece4e68a2d0fbeb6639c2e5469R228. Could we not decrement the retry variable explicitly when we make the attempt to simplify readability of this code?

Done, https://github.com/gokhanbarisaker/ably-java/commit/55746eb980cd7917a2f51d8bd13aae31a24405f9


> My comment about inconsistent use of upper case as well as missing underscores is still not addressed

Done, https://github.com/gokhanbarisaker/ably-java/commit/0fcb999e7f9c8acdd2829e4909127b6cf0eb3299


> Setting httpMaxRetryCount in all tests seems unnecessary, see https://github.com/ably/ably-java/pull/80/files#diff-d86628d178e27c35757c30abecfa69f3R75. I don't think you responded to that one?

Done, https://github.com/gokhanbarisaker/ably-java/commit/9e07f45f5dafbf676e2d93f3d912792437e0513c


> Not sure how FALLBACKS could ever be empty given it's static and final, see https://github.com/ably/ably-java/pull/80/files#diff-633aa103b8a41039ef0c67f40e6a610aR21 and https://github.com/ably/ably-java/pull/80/files#diff-060685a947a7032665de7ac9d33bb6ffR486

Done, https://github.com/gokhanbarisaker/ably-java/commit/5e8a5238cfefbe189b83d66291cd7fb95a402190

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -50,6 +51,7 @@ public AblyRealtime(ClientOptions options) throws AblyException {
 		super(options);
 		channels = new Channels();
 		connection = new Connection(this);
+		http.setHost(options.realtimeHost);
```
paddybyers: You're missing the point. The host that you happen to be connected to will change from time to time, because the fallback hosts behaviour applies to realtime connections, not just rest requests. So it won't work to freeze the realtime host here for the lifetime of the library.

paddybyers:
> I am setting it explicitly on AblyRealtime constructor

See https://github.com/ably/ably-java/pull/80#discussion_r52355212

lib/src/test/java/io/ably/lib/test/rest/RestAuthTest.java
```diff
@@ -106,6 +106,7 @@ public void authinit2() {
 			opts.port = testVars.port;
 			opts.tlsPort = testVars.tlsPort;
 			opts.tls = testVars.tls;
+			opts.httpMaxRetryCount = testVars.httpMaxRetryCount;
```
mattheworiordan: All of these httpMaxRetryCount configurations are still here and not related to the test in hand at all?  Why can't they simply use the default?

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Provides fallback host alternative for given host
+     *
+     * @param host
+     * @return Successor host that can be used as a fallback
+     */
+    public static String getFallback(String host) {
+        // Keep trying with host
+        if (FALLBACKS.isEmpty()) return host;
```
mattheworiordan: You've still missed this one

mattheworiordan:
@gokhanbarisaker I've added a few more comments...

gokhanbarisaker:
> No test for a 500 response? Given the changes you have made here, it would be nice to know that this works as intended and that we don't have a reg…

Done, https://github.com/gokhanbarisaker/ably-java/commit/d196d3fd5e0ef2178f210c8758ba398de8b68f58


> Should we not have an else here to ensure that the host is reset to the default REST endpoint when fallback is false?

Absolutely. Done, https://github.com/gokhanbarisaker/ably-java/commit/af2b5fbffdcbfaa1b0ad7a88bf871471f1892575


> This should be changed still, it should be HTTP_MAX_RETRY_COUNT i.e. you have to use underscores consistently for word breaks

Done, https://github.com/gokhanbarisaker/ably-java/commit/25cf2470dd50933109a1c85a86067c58a5c02c07


> All of these httpMaxRetryCount configurations are still here and not related to the test in hand at all? Why can't they simply use the default?

Done, https://github.com/gokhanbarisaker/ably-java/commit/1c6e9c2a065f226f4fbf2e500ac1b3c7b3afc7a9

Also, I have updated the fallback handling behavior.


> I think the logic is slightly wrong here, it should be while (retryCountRemaining >= 0); Example: Assume max retries is 2 First run against primar…

Done, https://github.com/gokhanbarisaker/ably-java/commit/fbba8db137aac18e153b58d9c500cfc555627ce2

lib/src/test/java/io/ably/lib/test/realtime/RealtimeInitTest.java
```diff
@@ -94,7 +94,6 @@ public void init_host() {
 			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
 			opts.restHost = "some.other.host";
 			ably = new AblyRealtime(opts);
-			assertEquals("Unexpected host mismatch", Defaults.getHost(opts), opts.restHost);
```
mattheworiordan: Why is this removed?

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+     * Spec: RSC15a
+     * </p>
+     *
+     * @throws Exception
+     */
+    @Test
+    public void http_execute_nofallback() throws Exception {
+        Http http = Mockito.spy(new Http(new ClientOptions(), null));
+
+        String hostExpected = "non.fallback.host";
```
mattheworiordan: I think I am missing something here because you are checking that no fallback is used, yet that host name is not a product host anyway so no fallback would ever be used?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeInitTest.java
gokhanbarisaker: AblyRealtime is using Http#setHost https://github.com/ably/ably-java/pull/80/files#diff-8a8a56ece4e68a2d0fbeb6639c2e5469R136

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+        String responseExpected = "Meow";
+        ArgumentCaptor<URL> argument = ArgumentCaptor.forClass(URL.class);
```
mattheworiordan: Is this a URL argument, can we not simply name it `url`?  Makes all the remaining parts of this far more readable

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+        http.setHost(hostExpected);
+
+        /* Partially mock http */
+        doReturn(responseExpected)
```
mattheworiordan: I find this a bit confusing, there is no mention of HTTP status codes in these mocks at all?  Reading this entire mock is really not intuitive, is there any way to make it more readable by perhaps defining variables for the arguments so that you have something more readable?  Unless someone looks up the arguments in an IDE / in the source, all of these statements do not tell me what is going on here

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void http_execute_singlefallback() throws Exception {
+        Http http = Mockito.spy(new Http(new ClientOptions(), null));
+
+        String hostExpectedPattern = PATTERN_HOST_FALLBACK;
+        String responseExpected = "Meow";
+        ArgumentCaptor<URL> argument = ArgumentCaptor.forClass(URL.class);
+        http.setHost("non.fallback.host");
```
mattheworiordan: If the host is not the default production host, how would this test pass?

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Exactly. There won't be a fallback and this test case is checking that. Maybe the host name might sound confusing.

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+
+        String responseActual = (String) http.ablyHttpExecute("", "", new Param[0], new Param[0],
+                Mockito.mock(Http.RequestBody.class), Mockito.mock(Http.ResponseHandler.class));
+
+
+        /* Verify
+         *   - http call executed twice,
+         *   - with a fallback host,
+         *   - and delivered expected response */
+        verify(http, times(2)).httpExecute(argument.capture(), anyString(), aryEq(new Param[0]),
```
mattheworiordan: Perhaps I am being a bit stupid here, but I don't see how, in the above code, you are providing an invalid response once, and then a valid response a second time?  Is there not a clearer way to mock these HTTP requests so that it's you are working at an HTTP level instead of at this level where everything is mocked out and just not clear?

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: We don't have anything to do with HTTP status codes with this test case. That behavior is partially mocked here. Of course, I can change variable names. Do you have suggestions for that?

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+        /* Partially mock http */
+        doAnswer(new GrumpyAnswer(excessiveFallbackCount, new Http.HostFailedException(new Exception()), ""))
+                .when(http).httpExecute(argument.capture(), anyString(), aryEq(new Param[0]),
+                any(Http.RequestBody.class), anyBoolean(), any(Http.ResponseHandler.class));
+
+
+        /* Verify
+         *   - ably exception with 404 status code is thrown
+         */
+        ErrorInfo expectedErrorInfo = new ErrorInfo("Connection failed; no host available", 404, 80000);
```
mattheworiordan: Fallback host should NEVER be used for a 404 error code, see http://docs.ably.io/client-lib-development-guide/features/#RSC15d

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+     */
+    @Test
+    public void http_execute_throws_500() throws MalformedURLException, AblyException {
+        Http http = new Http(new ClientOptions(), null);
+        Http.RequestBody requestBody = new Http.ByteArrayRequestBody(new byte[0], NanoHTTPD.MIME_PLAINTEXT);
+        URL url = new URL("http://localhost:" + server.getListeningPort() + "/status/500");
+
+        /* Verify
+         *   - HostFailedException is thrown */
+        thrown.expect(Http.HostFailedException.class);
```
mattheworiordan: This is not testing that the fallback is used as far as I can tell?  It's just checking that a 500 throws an error?

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,447 @@
...
+    @Test
+    public void http_execute_throws_504() throws MalformedURLException, AblyException {
+        Http http = new Http(new ClientOptions(), null);
+        Http.RequestBody requestBody = new Http.ByteArrayRequestBody(new byte[0], NanoHTTPD.MIME_PLAINTEXT);
+        URL url500 = new URL("http://localhost:" + server.getListeningPort() + "/status/504");
+
+        /* Verify
+         *   - HostFailedException is thrown */
+        thrown.expect(Http.HostFailedException.class);
+        http.httpExecute(url500, Http.GET, new Param[0], requestBody, false, null);
```
mattheworiordan: I am not sure why we needed 5 different tests here, surely you would instead have one test that has a for loop if you wanted to be exhaustive and test all codes?  Also, why are you not testing that codes outside of 500-504 are not using fallback?  Testing requires both positive and negative testing to be sure it's working as expected else you can easily get false positives.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -476,16 +482,11 @@ private StateIndication checkSuspend(StateIndication stateChange) {
...
-			String[] fallbackHosts = Defaults.getFallbackHosts(options);
-			if(fallbackHosts != null && fallbackHosts.length > 0) {
-				/* we will try a fallback host */
-				StateIndication fallbackConnectRequest = new StateIndication(ConnectionState.connecting, null);
-				fallbackConnectRequest.useFallbackHost = true;
-				requestState(fallbackConnectRequest);
-				/* returning null ensures we stay in the connecting state */
-				return null;
-			}
+		if(pendingConnect != null && checkConnectivity()) {
```
mattheworiordan: Perhaps this is outside of this PR, but firstly you are not checking if fallback hosts are supported here, but secondly you are performing an internet up check for every single retry which is unnecessary.  If the primary host fails, you check that the internet is up, and if it is, you try all the fallback hosts up to the max number.  There is no need to keep checking if the internet is up between each fallback host attempt.

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -2,43 +2,26 @@
...
-			host = options.restHost;
-			if(host == null)
-				host = Defaults.REST_HOST;
-		}
+public class Defaults {
+	public static final int PROTOCOL_VERSION = 1;
+	static final List<String> HOST_FALLBACKS = Arrays.asList("A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com");
+	public static final String HOST_REST = "rest.ably.io";
+	public static final String HOST_REALTIME = "realtime.ably.io";
+	public static final int PORT                    = 80;
```
mattheworiordan: Just a styling comment, but why align these `= x` yet not align the ones above or the `TRANSPORT` line below?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -272,62 +312,31 @@ private static AblyException checkFor(Throwable t) {
...
-					is.close();
-				if(conn != null)
-					conn.disconnect();
-			} catch(IOException ioe) {}
+			if(conn != null)
+				conn.disconnect();
 		}
 
-		if(statusCode == 0) {
+		if (response.statusCode == 0) {
```
mattheworiordan: Out of interest, in what condition would we validly return `null` from an `httpExecute` as opposed to raising an exception?

lib/src/main/java/io/ably/lib/transport/Defaults.java
gokhanbarisaker: Minor side effect of renaming. Just fixed

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Different status codes creates different values. Testing every status code would require me going from 1xx to 5xx. Would you really want me to do that?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -222,35 +262,35 @@ private static AblyException checkFor(Throwable t) {
...
+					Log.d(TAG, "Connection failed to host `" + hostCurrent + "`. Searching for new host...");
+					hostCurrent = Hosts.getFallback(hostCurrent);
+					Log.d(TAG, "Switched to `" + hostCurrent + "`.");
 				}
 			}
-			throw new AblyException("Connection failed; no host available", 404, 80000);
-		}
+		} while (retryCountRemaining >= 0);
+
+		throw new AblyException("Connection failed; no host available", 404, 80000);
```
mattheworiordan: I am not sure why we would throw this generic exception instead of re-raising the last exception failure?

lib/src/main/java/io/ably/lib/http/Http.java
paddybyers: > Out of interest, in what condition would we validly return null from an httpExecute as opposed to raising an exception?

A 204 response?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -81,19 +116,25 @@
...
-		return Defaults.getHost(ably.options);
+	public boolean isFallbackSupported() {
+		return host.matches(PATTERN_HOST_PROD);
+	}
+
+	/**
+	 * Sets host for this HTTP client
+	 *
+	 * @param host URL string
+     */
```
mattheworiordan: Invalid indentation

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Yes, and error handling is testing within other test cases. Let's keep the unit tests simple and clean.

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: This is displayed similar within IDE. This is why I am pushing `Reformating code` change. Currently I am unable to detect these mistakes with bare eyes, and these are only visible when github is using different spacing for tabs

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: This is same with old code. @paddybyers might help on that

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: This is testing behavior when client is failed more than (httpMaxRetryCount + 1). e.g., Assume httpMaxRetryCount is 1

  - Execute a request to host A. Let's assume it is failed
  - Fallback to host B. Let's assume it is also failed

Now can't do fallback and try again any more because of httpMaxRetry count constraint. Therefore, we are throwing an exception with 404

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: It turns out it does not pass currently, and I need to update the expected behavior.

gokhanbarisaker: Done

lib/src/test/java/io/ably/lib/test/realtime/RealtimeInitTest.java
mattheworiordan: So this worries me on two counts.  Firstly, if you ave a test without an assertion, this is not a test.  Secondly, this is a realtime test which is supposed to test that when the REST host is set, all rest requests go to that host.  Equally, if a realtimeHost is set, then all realtime requests go to that host.  We need that test as I am not confident that is working without suitable test coverage.

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Done

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: But the spec says `Validates http is not using any fallback host when we receive valid response from http's host`.  If you are not using the default host, then that condition will never apply anyway, see http://docs.ably.io/client-lib-development-guide/features/#RSC15b.  So this test is not testing anything really because a fallback host would never have been used with a non-default host

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: Perhaps I am missing how the mocking works, but I would imagine a simple way to mock in English would be:

* Ensure default host is being used
* Issue an HTTP request to an Ably endpoint
* Mock the `GET /channels/test/message` request with a HTTP status code 200 response and body `[{ "dummy": "response" }]`
* Ensure the request succeeded, no fallback was used, and the body is `[{ "dummy": "response" }]`

I just don't see that type of logic in this test unfortunately, possibly because it's obfuscated with mocks that do not describe what they are doing clearly

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: Certainly not, I was surprised you wrote tests for all 5 50x status codes. I am saying that you have 5 test for 500-504 which could have more easily be put into a for loop, and you have no test to ensure that say a 505 does NOT result in a fallback host check, which means you are only testing the conditions you expect.

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: Sorry, I was not clear, why is this checking for a status code of zero and returning null? A status code of zero indicates a failed response unless I am misunderstanding the nuances of some underlying library here?

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Yes, because I don't need to test whole flow just to test one simple behavior. This is why unit tests are fast and precise, unlike integration tests.

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: But your comment above from the spec states `Validates {@code Http} is performing a fallback when api returns a response code between 500 and 504`.  So this test is not testing that?

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: If you use tabs only, then you won't have this issue

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: Ok, this needs to be addressed, but understand this is a result of https://github.com/ably/ably-java/pull/80/files#r52723376

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: thanks

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: Don't get me wrong, I am a big advocate of unit tests, but in combination with integration tests.  I just don't feel we can currently, with just these unit tests, be confident that the behaviour is exactly as expected.

gokhanbarisaker: Would you like me to keep these tests and add new ones that hit the network? or would you like me to remove them?

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: Exactly

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: I do like unit tests, so I would prefer just one or two that hit the network to prove that the library is doing what it should do, and these unit tests to quickly test the permutations as you are

gokhanbarisaker: Got it. I will add new tests that actually makes the request

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
gokhanbarisaker: I solved that issue here https://github.com/ably/ably-java/pull/80/files#diff-060685a947a7032665de7ac9d33bb6ffR588

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: `Http` is the name of the class here, not the protocol :)

lib/src/test/java/io/ably/lib/http/HttpTest.java
gokhanbarisaker: I couldn't figure out how to specify `ExpectedException` to expect multiple thrown exceptions within same test case. This is why I moved them to their own test cases

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
gokhanbarisaker: Is it possible that we create another task for that?

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
mattheworiordan: I see two problems with this approach:

1. You are setting the REST Host to the RealtimeHost in all situations. This is wrong, we only use the fallback host for REST requests when the realtime host is using that fallback host.
2. If a realtime request falls back to a fallback host, but then later resumes using the default realtime host, all future REST requests will continue to go to the incorrect fallback host, see https://github.com/ably/ably-java/pull/80/files#diff-060685a947a7032665de7ac9d33bb6ffR588

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: I know that, but `is performing a fallback` is the part I was talking about, it's not performing any fallback in this test

lib/src/test/java/io/ably/lib/http/HttpTest.java
mattheworiordan: A `try` `catch` block would do the trick if the testing framework is unable to do that

lib/src/main/java/io/ably/lib/http/Http.java
paddybyers: @mattheworiordan as per [my comment](https://github.com/ably/ably-java/issues/92#event-548339243) on the other issue. This should be `AblyException.fromThrowable(...)` which will return the original exception if it was already an `AblyException`, and wrap it if not.

Down the line, the reporting of an error from an `AblyException` should refer to the `cause` if there is one.

lib/src/main/java/io/ably/lib/http/Http.java
paddybyers: I don't think the HTTP request will ever validly return a `statusCode` of 0. That is, it would only happen if the server really did put a `0` in the status line of the response. If it is unable to parse the status line the returned code will be -1.

I think the correct behaviour there is to throw an exception signifying an invalid response if the `statusCode` is not in the range 2xx-5xx.

The previous code looks like it was checking for 0 as a way of deciding whether or not there had been any response, but the correct action in that case would have been to throw instead of silently returning `null`.

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: Ok, @gokhanbarisaker can you implement this recommendation?

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: Ok, agreed, @gokhanbarisaker can you fix as well please?

gokhanbarisaker:
Rebased to 8a151b4edfa228ddef806fa10d2f9ce2be1ac09c

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: @gokhanbarisaker are you not going to address this?  If not, can you raise a separate issue for this problem so that it is fixed?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: Sure, can you please create an issue so that we can ignore it in the context of this PR?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -594,7 +587,17 @@ private void connectImpl(StateIndication request) {
...
+		pendingConnect = new ConnectParams(options);
+
+		if (request.fallback && Hosts.isRealtimeFallbackSupported(options.realtimeHost)) {
+			String hostFallback = Hosts.getFallback(getHost());
+			pendingConnect.host = hostFallback;
+			ably.http.setHost(hostFallback);
+		}
+		else {
+			pendingConnect.host = options.realtimeHost;
+			ably.http.setHost(options.realtimeHost);
```
mattheworiordan: We discussed that this was supposed to be `restHost`?

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -2,43 +2,32 @@
...
-	public static final int PORT                = 80;
-	public static final int TLS_PORT            = 443;
-	public static final int connectTimeout      = 15000;
-	public static final int disconnectTimeout   = 30000;
-	public static final int suspendedTimeout    = 120000;
-	public static final int cometRecvTimeout    = 90000;
-	public static final int cometSendTimeout    = 10000;
-	public static final String[] transports     = new String[]{"web_socket"};
-	public static final String transport        = "io.ably.lib.transport.WebSocketTransport$Factory";
+	public static final int PROTOCOL_VERSION        = 1;
```
mattheworiordan: What is this for?  PROTOCOL_VERSION is `0.8` anyway as far as I was aware? http://docs.ably.io/client-lib-development-guide/features/#G4

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -2,43 +2,32 @@
...
+	static final List<String> HOST_FALLBACKS        = Arrays.asList("A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com");
+	public static final String HOST_REST            = "rest.ably.io";
+	public static final String HOST_REALTIME        = "realtime.ably.io";
+	public static final int PORT                    = 80;
+	public static final int TLS_PORT                = 443;
+	public static final int TIMEOUT_CONNECT         = 15000;
+	public static final int TIMEOUT_DISCONNECT      = 30000;
+	public static final int TIMEOUT_SUSPEND         = 120000;
+	/* TO313 */
+	public static final int TIMEOUT_HTTP_OPEN = 4000;
```
mattheworiordan: Why the lack of indentation again?

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -2,43 +2,32 @@
...
-	}
 
-	public static String getHost(ClientOptions options, String host, boolean ws) {
-		if(host == null) {
-			host = options.restHost;
-			if(host == null)
-				host = Defaults.REST_HOST;
-		}
+	public static final String[] TRANSPORTS         = new String[]{"web_socket"};
+	public static final String TRANSPORT = "io.ably.lib.transport.WebSocketTransport$Factory";
```
mattheworiordan: Again, indentation above & below?

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -0,0 +1,56 @@
...
+     * </p>
+     * <p>
+     * Spec: RSC15b
+     * </p>
+     *
+     * @param host
+     * @return true, if the given host is qualified for a retry against a fallback host. Otherwise, false.
+     */
+    public static boolean isRestFallbackSupported(String host) {
+        return host.matches(PATTERN_REST_PROD);
```
mattheworiordan: This seems overkill, why not simply `host == "rest.ably.io"`?  Same applies below

paddybyers: Well it should be `host.equals( .... )`; you can't use `==` to test for string equality unless you know both strings are interned.

mattheworiordan: Sure, I suppose what I meant was a regex is quite unnecessary for a direct match

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: Ignore me, I did add an issue :confused: and didn't update this comment thread, see https://github.com/ably/ably-java/issues/93

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+	@Test
+	public void http_execute_nofallback() throws Exception {
+		Http http = Mockito.spy(new Http(new ClientOptions(), null));
+
+		String responseExpected = "Lorem Ipsum";
+		String hostExpected = Defaults.HOST_REST;
+		http.setHost(hostExpected);
+		ArgumentCaptor<URL> url = ArgumentCaptor.forClass(URL.class);
+
+        /* Partially mock http */
```
mattheworiordan: Minor I know, but indentation again

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+				"", /* Ignore */
+				new Param[0], /* Ignore */
+				new Param[0], /* Ignore */
+				mock(Http.RequestBody.class), /* Ignore */
+				mock(Http.ResponseHandler.class) /* Ignore */
+		);
+
+
+        /* Verify
+		 *   - http call executed once,
```
mattheworiordan: Indentation again

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+	@Test
+	public void http_execute_singlefallback() throws Exception {
+		Http http = Mockito.spy(new Http(new ClientOptions(), null));
+
+		String hostExpectedPattern = PATTERN_HOST_FALLBACK;
+		String responseExpected = "Lorem Ipsum";
+		ArgumentCaptor<URL> url = ArgumentCaptor.forClass(URL.class);
+		http.setHost(Defaults.HOST_REST);
+
+        /* Partially mock http */
```
mattheworiordan: Indentation...

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+				.httpExecute(
+						url.capture(), /* capture url arguments passed down httpExecute to assert fallback behavior executed with valid fallback url */
+						anyString(), /* Ignore */
+						aryEq(new Param[0]), /* Ignore */
+						any(Http.RequestBody.class), /* Ignore */
+						anyBoolean(), /* Ignore */
+						any(Http.ResponseHandler.class) /* Ignore */
+				);
+
+        /* Call method with real parameters */
```
mattheworiordan: I'll stop commenting on indentation, please review in Github and correct in one go...

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+		AblyException.HostFailedException hfe;
+
+		for (int statusCode = 500; statusCode <= 504; statusCode++) {
+			url = new URL("http://localhost:" + server.getListeningPort() + "/status/" + statusCode);
+			hfe = null;
+
+			try {
+				http.httpExecute(url, Http.GET, new Param[0], requestBody, false, null);
+			} catch (AblyException.HostFailedException e) {
+				hfe = e;
```
mattheworiordan: Why are we not checking that the fallback hosts are used for 50x errors?

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -0,0 +1,637 @@
...
+	 * when host & fallback hosts are unreachable. Then, finally throws an error.
+	 * </p>
+	 * <p>
+	 * Spec: RSC15a
+	 * </p>
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void http_ably_execute_fallback() throws AblyException {
```
mattheworiordan: Could we add another integration test that confirms fallbacks are not used when the default host is not used?

mattheworiordan:
This is looking good @gokhanbarisaker, we're pretty much there.  Few indentation issues, and a few other comments, but I think pretty much ready to go

PR #86, opened 2016-02-05
Fix recover spec regex pattern

gokhanbarisaker:
Fixes https://travis-ci.org/ably/ably-java/jobs/107128933#L406


paddybyers:
LGTM, thanks

PR #94, opened 2016-02-15
Convert enum variable naming to lowercase

### LINKED ISSUE #90: Lower case PresenceMessage.Action enum
If you look at [ConnectionState](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/realtime/ConnectionState.java#L6-L15), the enums are all lowercase allowing something like:

```java
switch (state) {
  case connected:
  case failed:
}
```

and the following:

```java
connection.on(ConnectionState.connected, listener) { ... }
```

Same applies for [ChannelState](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/realtime/ChannelState.java#L6-L13).

However, [`PresenceMessage.Action`](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/types/PresenceMessage.java#L28-L37) uses upper case making the API interface for devs consistent.  For example, a user will now need to do the following (once https://github.com/ably/ably-java/issues/21 is resolved):

```java
presence.subscribe(PresenceMessage.Action.ENTER, listener) { ... }
``` 

and

```java
switch (action) {
  case ENTER:
  case LEAVE:
}
```

I believe we should aim for consistency in these public enums and should probably just go lowercase. 

WDYT @paddybyers?

gokhanbarisaker:
Resolves #90

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Channel.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Channel.java
@@ -108,7 +108,7 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 			throw AblyException.fromErrorInfo(connectionManager.getStateErrorInfo());
 
 		/* send attach request and pending state */
-		ProtocolMessage attachMessage = new ProtocolMessage(Action.ATTACH, this.name);
+		ProtocolMessage attachMessage = new ProtocolMessage(Action.attach, this.name);
 		try {
 			setState(ChannelState.attaching, null);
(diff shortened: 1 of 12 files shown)


mattheworiordan:
LGTM, although I see you applied your lowercase to all Enums.  I am happy with this, but @paddybyers should give his thumbs up.

paddybyers:
I didn't expect that everything would change, but consistent is good, so LGTM.

PR #95, opened 2016-02-15
Presence overloading

### LINKED ISSUE #89: Presence enter, update, leave methods need to be overloaded
The `enter`, `update`, `leave` methods cannot be called without a listener and/or without a payload.  They should be overloaded to allow for this.

This should probably also apply to `*Client` such as `enterClient` as well.

See http://docs.ably.io/client-lib-development-guide/features/#idl for the IDL

gokhanbarisaker:
Resolves #89

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -191,8 +191,9 @@ public class Presence {
 	}
 
 	/**
-	 * Enter a specified client into this channel. The given client will be added to the
-	 * presence set and presence subscribers will see an empty presence message.
+	 * Enter a specified client into this channel. The given clientId will be added to
+	 * the presence set and presence subscribers will see a corresponding presence message
+	 * with an empty data payload.
 	 * This method is provided to support connections (eg connections from application
 	 * server instances) that act on behalf of multiple clientIds. In order to be able to
 	 * enter the channel with this method, the client library must have been instanced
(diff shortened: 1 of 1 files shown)


gokhanbarisaker:
I am not sure about the behavior when data is null. Can you please confirm below lines (javadoc about empty message) before merging?

- https://github.com/ably/ably-java/pull/95/files#diff-fc7f109a620e3e0188e0516eb98ed0f5R195
- https://github.com/ably/ably-java/pull/95/files#diff-fc7f109a620e3e0188e0516eb98ed0f5R241
- https://github.com/ably/ably-java/pull/95/files#diff-fc7f109a620e3e0188e0516eb98ed0f5R283

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -192,6 +192,34 @@ public void leave(CompletionListener listener) throws AblyException {
 
 	/**
 	 * Enter a specified client into this channel. The given client will be added to the
+	 * presence set and presence subscribers will see an empty presence message.
```
mattheworiordan: This is right but the wording is confusing.  I would instead say:
`The given clientId will be added to the presence set with an empty member data payload.`

paddybyers:
LGTM

mattheworiordan:
@gokhanbarisaker can you add the additional overloads mentioned and I will merge?

paddybyers:
> can you add the additional overloads mentioned and I will merge?

Which ones? I don't think it's a good idea to add overloads that rely on that disambiguating rule I mentioned (see [comment](https://github.com/ably/ably-java/pull/95#commitcomment-16100433))

mattheworiordan:
> Which ones? I don't think it's a good idea to add overloads that rely on that disambiguating rule I mentioned (see comment)

Oops, I didn't see the second line of [your comment](https://github.com/ably/ably-java/pull/95#commitcomment-16100433) weirdly. Why don't you want the disambiguating rule if it works?  Those overloaded methods are supported in [our IDL](http://docs.ably.io/client-lib-development-guide/features/#idl) and in [our presence docs](http://docs.ably.io/realtime/presence/) (see API reference section above http://docs.ably.io/realtime/presence/#sync-complete)

paddybyers:
> Why don't you want the disambiguating rule if it works?

I just think it adds a needless level of confusion and ambiguity. The developer might need to go and look up the spec before he knows what will happen. If he passes `(null, null)` for `data` and `listener` he'll get a compile error. Instead all the confusion goes away if he simply passes `null` for `data`.

> Those overloaded methods are supported in our IDL

IMO that's fine so long as it adds convenience, but unnecessary if it risks adding confusion.

mattheworiordan:
Ok fine @paddybyers, will merge

Issue #17, opened 2015-04-22
Message publish overloaded without a listener


mattheworiordan:
I imagine in many cases a developer won't care if a message was published or not and as such won't need to pass in a CompletionListener.  It would be nice to provide an overloaded version of this method without the CompletionListener requirement

mattheworiordan:
FYI, in Javascript & Ruby this is easily done by simply not passing in the listener callback, but in Java this is just a bit more cumbersome because an explicit `Null` or Listener object is required.

PR #98, opened 2016-02-17
Overload publish APIs

gokhanbarisaker:
Resolves: #17


paddybyers:
LGTM

mattheworiordan:
Where we are calling methods without a listener, it would be good to at least add test coverage to one of these methods so we can be sure that the null listener does not cause any problems.  Would you mind doing that please @gokhanbarisaker

gokhanbarisaker:
Do you want me to check either null listener argument creates any problem within underlying API? or would you like me to just confirm passed null listener argument is delivered to underlying API?

mattheworiordan:
> Do you want me to check either null listener argument creates any problem within underlying API? or would you like me to just confirm passed null listener argument is delivered to underlying API?

I would like to see that that call works so simply copy & paste an existing test that uses a listener and use the new overloaded method.

gokhanbarisaker:
Got it

paddybyers:
Otherwise LGTM, thanks

Issue #36, opened 2015-08-19
Realtime Presence and Channel untilAttach functionality is missing


mattheworiordan:
See https://github.com/ably/ably-js/commit/885dd83c0cde9cc017cb4623b98c30eeaebb2667 where it was introduced in JS for Channel.  See also https://github.com/ably/ably-ruby/blob/382cff363d02e82ef3860a7612493a820750f4a9/spec/acceptance/realtime/presence_history_spec.rb#L56 and https://github.com/ably/ably-ruby/blob/245ac42b0e31cd03393022eb35952796a1a094b9/spec/acceptance/realtime/channel_history_spec.rb#L155 for the Ruby implementation.

We need this functionality for the chat demos we are being building.

mattheworiordan:
Note when this is released we must update the docs at www.ably.io/documentation/realtime/channels-messages and www.ably.io/documentation/realtime/presence

PR #102, opened 2016-02-19
Add until attach functionality to Presence & Channel

gokhanbarisaker:
Resolves: #36

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Channel.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Channel.java
@@ -20,8 +20,11 @@ import io.ably.lib.util.EventEmitter;
 import io.ably.lib.util.Log;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+
 
 /**
  * A class representing a Channel belonging to this application.
@@ -475,6 +478,37 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 		queuedMessages.clear();
 	}
 
+	static Param[] replacePlaceholderParams(Channel channel, Param[] placeholderParams) throws AblyException {
+		HashSet<Param> params = new HashSet<>();
+
+		Param param;
+		for(int i = 0; i < placeholderParams.length; i++) {
+			param = placeholderParams[i];
+
+			if(KEY_UNTIL_ATTACH.equals(param.key)) {
+				if("true".equalsIgnoreCase(param.value)) {
+					if (channel.state != ChannelState.attached) {
+						throw AblyException.fromErrorInfo(new ErrorInfo("option untilAttach requires the channel to be attached", 40000, 400));
+					}
+
+					params.add(new Param(KEY_FROM_SERIAL, channel.attachSerial));
+				}
+				else if(!"false".equalsIgnoreCase(param.value)) {
+					throw AblyException.fromErrorInfo(new ErrorInfo("option untilAttach is invalid. \"true\" or \"false\" expected", 40000, 400));
+				}
+			}
+			else {
+				/* Add non-placeholder param as is */
+				params.add(param);
+			}
+		}
+
+		return params.toArray(new Param[params.size()]);
+	}
+
+
+	private static final String KEY_UNTIL_ATTACH = "untilAttach";
+	private static final String KEY_FROM_SERIAL = "fromSerial";
 	private List<QueuedMessage> queuedMessages;
 
 	/************************************
@@ -491,8 +525,10 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 	 * @throws AblyException
 	 */
 	public PaginatedResult<Message> history(Param[] params) throws AblyException {
+		params = replacePlaceholderParams(this, params);
+
 		BodyHandler<Message> bodyHandler = MessageSerializer.getMessageResponseHandler(options);
-		return new PaginatedQuery<Message>(ably.http, basePath + "/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
+		return new PaginatedQuery<>(ably.http, basePath + "/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
 	}
 
 	/************************************
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -13,6 +13,7 @@ import io.ably.lib.types.PresenceSerializer;
 import io.ably.lib.types.ProtocolMessage;
 import io.ably.lib.util.Log;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -367,9 +368,11 @@ public class Presence {
 	 * @throws AblyException
 	 */
 	public PaginatedResult<PresenceMessage> history(Param[] params) throws AblyException {
+		params = Channel.replacePlaceholderParams(channel, params);
+
 		AblyRealtime ably = channel.ably;
 		BodyHandler<PresenceMessage> bodyHandler = PresenceSerializer.getPresenceResponseHandler(channel.options);
-		return new PaginatedQuery<PresenceMessage>(ably.http, channel.basePath + "/presence/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
+		return new PaginatedQuery<>(ably.http, channel.basePath + "/presence/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -475,6 +477,24 @@ private void failQueuedMessages(ErrorInfo reason) {
...
+				}
+
+				params[i] = new Param(KEY_FROM_SERIAL, attachSerial);
+			}
+		}
+	}
+
+
+	private static final String KEY_UNTIL_ATTACH = "untilAttach";
+	private static final String KEY_FROM_SERIAL = "from_serial";
```
paddybyers: We should use `fromSerial`.

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -434,6 +440,25 @@ private void failQueuedMessages(ErrorInfo reason) {
 			
 	}
 
+	private void replacePlaceholderParams(Param[] params) throws AblyException {
```
paddybyers: Please don't duplicate this. It can be a static of `Channel`.

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -475,6 +477,24 @@ private void failQueuedMessages(ErrorInfo reason) {
 		queuedMessages.clear();
 	}
 
+	private void replacePlaceholderParams(Param[] params) throws AblyException {
+		Param param;
+		for (int i = 0; i < params.length; i++) {
+			param = params[i];
+
+			if (KEY_UNTIL_ATTACH.equals(param.key) && "true".equals(param.value)) {
```
mattheworiordan: I would argue that if you detect `untilAttach`, then it should be removed in all cases regardless of the value, and replaced with the serial when the value equals `true` (perhaps ignore case for convenience).  I don't see any value in sending the `untilAttach` param to the Ably service ever. Also, if we are raising an exception if `untilAttach` is called when the channel is attached, why don't we raise an exception if `untilAttach` is anything other than `true` / `false` (case-insensitive) with an error something along the lines `option untilAttach is invalid. "true" or "false" expected`

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -491,6 +511,11 @@ private void failQueuedMessages(ErrorInfo reason) {
 	 * @throws AblyException
 	 */
 	public PaginatedResult<Message> history(Param[] params) throws AblyException {
+		/* We don't want to mess with users data. Acquire a copy of params array */
+		params = Arrays.copyOf(params, params.length);
```
mattheworiordan: Instead of doing a copy and then replacing the param why not let replacePlaceholderParams return a new Params object as it is iterating over the Params anyway and could easily build a new Array.  One less step?

lib/src/main/java/io/ably/lib/realtime/Channel.java
mattheworiordan: FYI @gokhanbarisaker we changed all query string params from under_score case to camelCase a little while ago but did not roll it out everywhere, hence this comment...

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelHistoryTest.java
```diff
@@ -1355,4 +1355,108 @@ public void run() {
...
+			/* create a channel */
+			final Channel txChannel = txAbly.channels.get(channelName);
+			final Channel rxChannel = rxAbly.channels.get(channelName);
+
+			/* attach sender */
+			txChannel.attach();
+			(new ChannelWaiter(txChannel)).waitFor(ChannelState.attached);
+			assertEquals("Verify attached state reached", txChannel.state, ChannelState.attached);
+
+			/* publish messages to the channel */
```
mattheworiordan: Might be worth modifying the comment to state that messages are published over a period of 5 seconds, and the attach request on the receiving channel is made at 2s.  However, this does seem like we're relying too much on everything to be synchronised.  Why not simply publish 25 messages before rxChannel is attached and wait for the 25 publishes to be completed, then publish 25 messages after rxChannel is attached, in this way there is NO chance that an attach delay for example could cause this test to fail.  Then once the 25 messages are received on rxChannel, do a history request with untilAttach and expect exactly 25 messages with IDs 0 -> 25.  This seems simpler to me and certainly a lot less brittle.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceHistoryTest.java
```diff
@@ -1292,4 +1292,124 @@ public void run() {
...
+	/**
+	 * Connect twice to the service, each using the default (binary) protocol.
+	 * Publish messages on one connection to a given channel; while in progress,
+	 * attach the second connection to the same channel and verify a message
+	 * history up to the point of attachment can be obtained.
+	 */
+	@Test
+	public void presencehistory_until_attach() {
+		AblyRealtime txAbly = null, rxAbly = null;
+		try {
```
mattheworiordan: As above, I think we should make this test more deterministic and less effected by network conditions

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceHistoryTest.java
```diff
@@ -1292,4 +1292,124 @@ public void run() {
...
+	}
+
+	/**
+	 * Verifies an Exception is thrown, when a channel history is requested
+	 * with parameter {"untilAttach":"true}" before client is attached to the channel
+	 *
+	 * @throws AblyException
+	 */
+	@Test(expected=AblyException.class)
+	public void channelhistory_until_attach_before_attached() throws AblyException {
```
mattheworiordan: Should we add one for presence as well?

lib/src/main/java/io/ably/lib/realtime/Channel.java
gokhanbarisaker: Done

lib/src/main/java/io/ably/lib/realtime/Presence.java
gokhanbarisaker: Done

lib/src/main/java/io/ably/lib/realtime/Channel.java
gokhanbarisaker: Done

lib/src/main/java/io/ably/lib/realtime/Channel.java
gokhanbarisaker: Done

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelHistoryTest.java
gokhanbarisaker: This is a copy of previous test case. We can create a new task for fixing both of them.

mattheworiordan: Ok, fine, please do that and we can then merge

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -491,8 +525,10 @@ private void failQueuedMessages(ErrorInfo reason) {
 	 * @throws AblyException
 	 */
 	public PaginatedResult<Message> history(Param[] params) throws AblyException {
+		params = replacePlaceholderParams(this, params);
+
 		BodyHandler<Message> bodyHandler = MessageSerializer.getMessageResponseHandler(options);
-		return new PaginatedQuery<Message>(ably.http, basePath + "/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
+		return new PaginatedQuery<>(ably.http, basePath + "/history", HttpUtils.defaultAcceptHeaders(ably.options.useBinaryProtocol), params, bodyHandler).get();
```
mattheworiordan: This is I am sure my lack of knowledge of Java syntax here, but why did we drop the `Message` here?

lib/src/main/java/io/ably/lib/types/Param.java
```diff
@@ -25,4 +25,23 @@ public static boolean containsKey(Param[] params, String key) {
...
+		return value != null ? value.equals(param.value) : param.value == null;
+
+	}
+
+	@Override
+	public int hashCode() {
+		int result = key != null ? key.hashCode() : 0;
+		result = 31 * result + (value != null ? value.hashCode() : 0);
+		return result;
+	}
```
mattheworiordan: Why is `equals` and `hashCode` overriden?

gokhanbarisaker: I am using them within HashSet, that requires all values to override both methods to ensure uniqueness.

https://github.com/ably/ably-java/pull/102/files#diff-db9d112187c909d918e250dd2544de78R482

lib/src/main/java/io/ably/lib/realtime/Channel.java
gokhanbarisaker: It is automatically using the generic passed down with `BodyHandler`. Therefore, we don't need to explicitly specify the generic again.

paddybyers:
This looks ok to me now. @mattheworiordan ?

mattheworiordan:
LGTM

Issue #103, opened 2016-02-19
force is an attribute of AuthOptions, not an argument


mattheworiordan:
`force` is an attribute of `AuthOptions`, not an [argument to the `authorise` method](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/rest/Auth.java#L322)

PR #107, opened 2016-02-28
Move force argument to AuthOptions as a variable

gokhanbarisaker:
Resolves: #103

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -189,7 +189,13 @@ public class Http {
 		if(auth.getAuthMethod() == AuthMethod.basic) {
 			authHeader = "Basic " + Base64Coder.encodeString(auth.getBasicCredentials());
 		} else {
-			auth.authorise(null, null, renew);
+			Auth.AuthOptions options = null;
+			if (renew) {
+				options = new Auth.AuthOptions();
+				options.force = true;
+			}
+
+			auth.authorise(options, null);
 			authHeader = "Bearer " + auth.getTokenAuth().getEncodedToken();
 		}
 		return authHeader;
--- a/lib/src/main/java/io/ably/lib/http/TokenAuth.java
+++ b/lib/src/main/java/io/ably/lib/http/TokenAuth.java
@@ -36,11 +36,11 @@ public class TokenAuth {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params, boolean force) throws AblyException {
+	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
 		Log.i("TokenAuth.authorise()", "");
 		if(tokenDetails != null) {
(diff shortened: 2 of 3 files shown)


lib/src/main/java/io/ably/lib/http/TokenAuth.java
```diff
@@ -36,11 +36,11 @@ public void setTokenDetails(TokenDetails tokenDetails) {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params, boolean force) throws AblyException {
+	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
 		Log.i("TokenAuth.authorise()", "");
 		if(tokenDetails != null) {
 			if(tokenDetails.expires == 0 || tokenValid(tokenDetails)) {
-				if(!force) {
+				if(!options.force) {
```
paddybyers: Don't you need to check here that `options != null` ?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -189,7 +189,9 @@ private String getAuthorizationHeader(boolean renew) throws AblyException {
 		if(auth.getAuthMethod() == AuthMethod.basic) {
 			authHeader = "Basic " + Base64Coder.encodeString(auth.getBasicCredentials());
 		} else {
-			auth.authorise(null, null, renew);
+			Auth.AuthOptions options = new Auth.AuthOptions();
```
paddybyers: I think you should do this only in the `renew == true` case.

lib/src/main/java/io/ably/lib/http/TokenAuth.java
gokhanbarisaker: I thought about that too, and decided not to take an action since test cases under `RestAuthTest` didn't failed. Now that, Travis reported `RealtimeConnectFailTest`'s `connect_token_expire_disconnected` failing because of that reason, I will fix it now

lib/src/main/java/io/ably/lib/http/Http.java
gokhanbarisaker: It doesn't seem like there is a difference between using,
  - null argument, and letting system use default options
  - using an empty authoptions and letting system merge it with default options

i.e.,

https://github.com/ably/ably-java/blob/5987ee1292a72dc2ba8415b2bb0e705aac139bd6/lib/src/main/java/io/ably/lib/rest/Auth.java#L334

https://github.com/ably/ably-java/blob/5987ee1292a72dc2ba8415b2bb0e705aac139bd6/lib/src/main/java/io/ably/lib/rest/Auth.java#L455

paddybyers: I just think there is no point in creating the AuthOptions and for it to be merged when you know it only contains an option that is already the default.

gokhanbarisaker: Got it

paddybyers:
LGTM

mattheworiordan:
LGTM

Issue #84, opened 2016-02-03
Attach CompletionListener


mattheworiordan:
The Java library is missing a callback for completion of `Channel#attach`

See http://docs.ably.io/client-lib-development-guide/features/#RTL4d and

mattheworiordan:
When this is fixed, please check all references to attach in the documentation, for example realtime history

PR #108, opened 2016-02-28
Add CompletionListener to Channel's attach API

gokhanbarisaker:
Resolves: #84

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Channel.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Channel.java
@@ -118,7 +118,7 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 		switch(state) {
 			case attaching:
 				if (listener != null) {
-					on(new AttachListener(listener));
+					on(new ChannelStateCompletionListener(listener, ChannelState.attached, ChannelState.failed));
 				}
 
 				return;
@@ -137,7 +137,7 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 		ProtocolMessage attachMessage = new ProtocolMessage(Action.attach, this.name);
 		try {
 			if (listener != null) {
-				on(new AttachListener(listener));
+				on(new ChannelStateCompletionListener(listener, ChannelState.attached, ChannelState.failed));
 			}
 
 			setState(ChannelState.attaching, null);
@@ -154,13 +154,30 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 	 * @throws AblyException
 	 */
 	public void detach() throws AblyException {
+		detach(null);
+	}
+
+	/**
+	 * Detach from this channel.
+	 * This call initiates the detach request, and the response
+	 * is indicated asynchronously in the resulting state change.
+	 * @throws AblyException
+	 */
+	public void detach(CompletionListener listener) throws AblyException {
 		Log.v(TAG, "detach(); channel = " + name);
 		/* check preconditions */
 		switch(state) {
 			case initialized:
+			case detached: {
+				if(listener != null) {
+					listener.onSuccess();
+				}
+				return;
+			}
 			case detaching:
-			case detached:
-				/* nothing to do */
+				if (listener != null) {
+					on(new ChannelStateCompletionListener(listener, ChannelState.detached, ChannelState.failed));
+				}
 				return;
 			default:
 		}
@@ -171,6 +188,10 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 		/* send detach request */
 		ProtocolMessage detachMessage = new ProtocolMessage(Action.detach, this.name);
 		try {
+			if (listener != null) {
+				on(new ChannelStateCompletionListener(listener, ChannelState.detached, ChannelState.failed));
+			}
+
 			setState(ChannelState.detaching, null);
 			connectionManager.send(detachMessage, true, null);
 		} catch(AblyException e) {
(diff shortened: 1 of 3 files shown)


paddybyers:
This isn't going to work because it assumes that there's only ever a single caller waiting for `attach()` to complete. Either have an array or a Multicaster, or register the listener for the attached event.

mattheworiordan:
>  Either have an array or a Multicaster, or register the listener for the attached event.

Given we have support for `once` on the EventEmitter, surely it would be easier to register the listener using `once` and leave the rest of the functionality as it is?

paddybyers:
> Given we have support for once on the EventEmitter, surely it would be easier to register the listener using once for success conditions and once for failed conditions and leave the rest of the functionality as it is?

Yes, assuming you remove the unused listener in each case.

gokhanbarisaker:
> Given we have support for once on the EventEmitter, surely it would be easier to register the listener using once for success conditions and once for failed conditions and leave the rest of the functionality as it is?

Updated. `once` definitely worked better here.

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -94,12 +94,43 @@ private void setState(ChannelState newState, ErrorInfo reason) {
...
+	 * @param listener When the channel is attached successfully or the attach fails and
+	 * the ErrorInfo error is passed as an argument to the callback
+	 * @throws AblyException
+	 */
+	public void attach(final CompletionListener listener) throws AblyException {
 		Log.v(TAG, "attach(); channel = " + name);
 		/* check preconditions */
 		switch(state) {
 			case attaching:
+				once(ChannelState.attached, new ChannelStateListener() {
```
paddybyers: You're not handling the possibility that the attach fails.

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -94,12 +94,43 @@ private void setState(ChannelState newState, ErrorInfo reason) {
...
+					@Override
+					public void onChannelStateChanged(ChannelState state, ErrorInfo reason) {
+						if(reason == null) {
+							listener.onSuccess();
+						}
+						else {
+							listener.onError(reason);
```
mattheworiordan: I am not sure in what situations we would want to consider the attach a failure even though the channel is attached.  I would imagine in all instances we call the `onSuccess`

paddybyers: > I am not sure in what situations we would want to consider the attach a failure even though the channel is attached

I mean if the next state is `failed`, then nothing calls the listener with the error, and the once listener for `attached` is not removed.

mattheworiordan: I know, but that's a separate issue because with this code if the attach succeeds but includes a reason, then the failed callback is called.

gokhanbarisaker: I got both of your concerns. 

  - I will add listener for both `attached` state and `failed` state.
  - Attached state will only trigger `onSuccess`

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -362,4 +363,75 @@ public void attach_fail() {
...
+			e.printStackTrace();
+			fail("init0: Unexpected exception instantiating library");
+		} finally {
+			if(ably != null)
+				ably.close();
+		}
+	}
+
+	/**
+	 * When client attaches to a channel successfully, verify
```
mattheworiordan: Test description is wrong, says successful

mattheworiordan:
One comment, otherwise LGTM

gokhanbarisaker:
Done

mattheworiordan:
@paddybyers you happy with this?

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -94,12 +94,51 @@ private void setState(ChannelState newState, ErrorInfo reason) {
...
+	 * the ErrorInfo error is passed as an argument to the callback
+	 * @throws AblyException
+	 */
+	public void attach(final CompletionListener listener) throws AblyException {
 		Log.v(TAG, "attach(); channel = " + name);
 		/* check preconditions */
 		switch(state) {
 			case attaching:
+				if (listener != null) {
+					on(new ChannelStateListener() {
```
paddybyers: Can't we remove this duplication by having a `private static class AttachListener implements ChannelStateListener { .... }` ?

paddybyers:
LGTM except for 2 comments:
1) I'd prefer that we avoid duplication of the anonymous listener class;
2) don't we also have to do the same for `detach()` ?

gokhanbarisaker:
@paddybyers 

> I'd prefer that we avoid duplication of the anonymous listener class;

Sure, I am on it.

> don't we also have to do the same for detach() ?

As per [RTL5e](http://docs.ably.io/client-lib-development-guide/features/#RTL5e), yes we do

Would you like to create another issue for RTL5e? or maybe modify #84 title?

paddybyers:
> Would you like to create another issue for RTL5e? or maybe modify #84 title?

The same PR is fine, and I don't really care about the title :)

mattheworiordan:
LGTM.  I would have liked to have seen a detach fail callback test, but probably not necessary.

gokhanbarisaker:
@mattheworiordan

> I would have liked to have seen a detach fail callback test, but probably not necessary.

I couldn't figure out how to implement that. Do you have any suggestion, or a reference test case from any other ably client libraries?

mattheworiordan:
> I couldn't figure out how to implement that. Do you have any suggestion, or a reference test case from any other ably client libraries?

Not worth the hassle for now, let's leave as is.  It require some stubbing which I believe there is little of in the lib at present. LGTM :+1:

Issue #33, opened 2015-08-10
PaginatedResult#isLast method missing


mattheworiordan:
See https://goo.gl/2aoKPg, and the spec `TG7` at http://docs.ably.io/client-lib-development-guide/features/, `isLast` method is missing.

paddybyers:
I thought that last time we reviewed that spec we agreed to remove `isLast()` ?

mattheworiordan:
Not as I recall, only `isFirst` because of the issue relating to the cursor and current not necessarily reflecting the first position.  I don't see any agreement on that.

gokhanbarisaker:
is `!hasNext()` and `isLast()` has a difference?

gokhanbarisaker:
Also, docs for TG7 seems confusing, where `isLast` returns true for non-last pages

> (TG7) PaginatedResult#isLast function returns true if this page is **not** the last page i.e. !hasNext

paddybyers:
> !hasNext() and isLast() are same things, aren't they?

Yes

> docs for TG7 seems confusing

You're right, and the doc is wrong.

mattheworiordan:
Typo in TG7 fixed at https://github.com/ably/docs/commit/08de08e7b75d21bf455f68111f0afaf3e2ba1999

PR #111, opened 2016-02-29
Add isLast API to PaginatedResult

gokhanbarisaker:
Resolves: #33

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelHistoryTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelHistoryTest.java
@@ -24,7 +24,9 @@ import io.ably.lib.types.PaginatedResult;
 import io.ably.lib.types.Param;
 
 import java.util.HashMap;
+import java.util.Locale;
 
+import org.hamcrest.Matchers;
 import org.junit.AfterClass;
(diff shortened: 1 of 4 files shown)


mattheworiordan:
I would like to see a test please

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelHistoryTest.java
```diff
@@ -1465,6 +1467,7 @@ public void run() {
 	}
 
 	/**
+<<<<<<< 97c43f880519f2ae4d9b1a160b778c5bca683858
```
mattheworiordan: Merge conflict

mattheworiordan:
Thanks @gokhanbarisaker, I didn't realise you were waiting on us, in future please post a message to re-review when you can.  Also, our preferred approach is to [use fixups](http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) when updating an already up PR. This way we can review the fixes you submit, and when ready, you can rebase by squashing the commits before merge.  Is that OK?

gokhanbarisaker:
I will start doing as instructed :)

paddybyers:
LGTM

Issue #45, opened 2015-11-02
Implicit attach on Publish or Subscribe


mattheworiordan:
The client library should implicitly attach on Publish or Subscribe on a channel, but does not appear to be doing this.

Please check that all other implicit operations on Channel & Presence work too, and add some tests to cover this in future.

paddybyers:
https://github.com/ably/ably-java/pull/48

PR #115, opened 2016-03-07
Add implicit attach test cases for channel publish and subscribe

gokhanbarisaker:
Resolves: #45 

- [x] RTL6c3
- [x] RTL7c
- [x] RTP6c
- [x] RTP8d
- [x] RTP11b
- [x] RTP15e

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -41,6 +41,10 @@ public class Presence {
 	 * @throws AblyException
 	 */
 	public synchronized PresenceMessage[] get() throws AblyException {
+		if (channel.state == ChannelState.failed) {
+			throw AblyException.fromErrorInfo(new ErrorInfo("channel operation failed (invalid channel state)", 90001));
+		}
+
 		channel.attach();
 		Collection<PresenceMessage> values = presence.values();
 		return values.toArray(new PresenceMessage[values.size()]);
@@ -54,6 +58,10 @@ public class Presence {
 	 * @throws AblyException
 	 */
 	public synchronized PresenceMessage[] get(boolean wait) throws InterruptedException, AblyException {
+		if (channel.state == ChannelState.failed) {
+			throw AblyException.fromErrorInfo(new ErrorInfo("channel operation failed (invalid channel state)", 90001));
+		}
+
 		channel.attach();
 		Collection<PresenceMessage> values = presence.values(wait);
 		return values.toArray(new PresenceMessage[values.size()]);
@@ -69,6 +77,10 @@ public class Presence {
 	 * @throws AblyException
 	 */
 	public synchronized PresenceMessage[] get(String clientId, boolean wait) throws InterruptedException, AblyException {
+		if (channel.state == ChannelState.failed) {
+			throw AblyException.fromErrorInfo(new ErrorInfo("channel operation failed (invalid channel state)", 90001));
+		}
+
 		channel.attach();
 		Collection<PresenceMessage> values = presence.getClient(clientId, wait);
 		return values.toArray(new PresenceMessage[values.size()]);
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -34,32 +34,42 @@ public class Presence {
 	 ************************************/
 
 	/**
-	 * Get the presence state for this Channel.
+	 * Get the presence state for this Channel. Implicitly attaches the
+	 * Channel. However, if the channel is in or moves to the FAILED
+	 * state before the operation succeeds, it will result in an error
 	 * @return: the current present members.
 	 * @throws AblyException
 	 */
-	public synchronized PresenceMessage[] get()  {
+	public synchronized PresenceMessage[] get() throws AblyException {
+		channel.attach();
 		Collection<PresenceMessage> values = presence.values();
 		return values.toArray(new PresenceMessage[values.size()]);
 	}
 
 	/**
 	 * Get the presence state for this Channel, optionally waiting for sync to complete.
+	 * Implicitly attaches the Channel. However, if the channel is in or moves to the FAILED
+	 * state before the operation succeeds, it will result in an error
 	 * @return: the current present members.
 	 * @throws AblyException
 	 */
-	public synchronized PresenceMessage[] get(boolean wait) throws InterruptedException {
+	public synchronized PresenceMessage[] get(boolean wait) throws InterruptedException, AblyException {
+		channel.attach();
 		Collection<PresenceMessage> values = presence.values(wait);
 		return values.toArray(new PresenceMessage[values.size()]);
 	}
 
 	/**
-	 * Get the presence state for a given clientId
+	 * Get the presence state for a given clientId. Implicitly attaches the
+	 * Channel. However, if the channel is in or moves to the FAILED
+	 * state before the operation succeeds, it will result in an error
 	 * @param wait
 	 * @return
 	 * @throws InterruptedException
+	 * @throws AblyException
 	 */
-	public synchronized PresenceMessage[] get(String clientId, boolean wait) throws InterruptedException {
+	public synchronized PresenceMessage[] get(String clientId, boolean wait) throws InterruptedException, AblyException {
+		channel.attach();
 		Collection<PresenceMessage> values = presence.getClient(clientId, wait);
 		return values.toArray(new PresenceMessage[values.size()]);
 	}
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
@@ -980,4 +980,40 @@ public class RealtimeChannelTest {
 				ably.close();
 		}
 	}
+
+	/**
+	 * <p>
+	 * Validate subscribe will result in an error, when the channel is
+	 * in or moves to the FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTL7c
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
+	public void attach_implicit_subscribe_fail() throws AblyException {
+		AblyRealtime ably = null;
+		try {
+			TestVars testVars = Setup.getTestVars();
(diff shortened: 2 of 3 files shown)


mattheworiordan:
LGTM, a few small comments

gokhanbarisaker:
What would you like me to use for error code? Is [90001](https://github.com/ably/ably-common/blob/5b2c36c7b282f44d37879e995bc70de28c20c46a/protocol/errors.json#L97) fine for Presence#get(...)?

mattheworiordan:
LGTM, @paddybyers you happy for this to be merged now?

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -34,32 +34,46 @@
...
-	 * Get the presence state for this Channel.
+	 * Get the presence state for this Channel. Implicitly attaches the
+	 * Channel. However, if the channel is in or moves to the FAILED
+	 * state before the operation succeeds, it will result in an error
 	 * @return: the current present members.
 	 * @throws AblyException
 	 */
-	public synchronized PresenceMessage[] get()  {
+	public synchronized PresenceMessage[] get() throws AblyException {
+		if (channel.state == ChannelState.failed) {
```
paddybyers: Why is this check here, and not on the other methods below?

paddybyers:
> @paddybyers you happy for this to be merged now?

See my comment above.

lib/src/main/java/io/ably/lib/realtime/Presence.java
gokhanbarisaker: My current focus is on Presence with this pull request. If other methods have features like  [RTP11b](http://docs.ably.io/client-lib-development-guide/features/#RTP11b), I am happy to implement that for them also

paddybyers: If the channel is in the failed state, does the `attach()` throw an exception? If so, the check above (l44) can be removed.

gokhanbarisaker: Currently, `channel.attach()` does not throw an exception for it. Would you like me to move the state check to `channel.attach()`, instead of `presence.get()`?

mattheworiordan: @paddybyers?

paddybyers: > Would you like me to move the state check to channel.attach(), instead of presence.get()?

No, we think that the state should be checked in `Presence.get()`, only calling `attach()` if not `failed`.

So please add that check to the other `get()` variants below.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -943,4 +943,78 @@ public void detach_success_callback_detached() throws AblyException {
...
+	public void attach_implicit_publish_fail() throws AblyException {
+		AblyRealtime ably = null;
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = testVars.createOptions(testVars.keys[1].keyStr);
+			ably = new AblyRealtime(opts);
+
+			/* wait until connected */
+			new ConnectionWaiter(ably.connection).waitFor(ConnectionState.connected);
+			assertEquals("Verify failed state reached", ably.connection.state, ConnectionState.connected);
```
paddybyers: This is verifying that the connected state is reached, not failed as in the message. (This problem also exists in the test cases below.)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -943,4 +943,78 @@ public void detach_success_callback_detached() throws AblyException {
...
+	 * Validate publish will result in an error, when the channel is
+	 * in or moves to the FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTL6c3
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
```
paddybyers: I don't think this test will test the things it says here.

The only assertions are checking for channel state, not for any exception from `publish()`. Also, if there is an exception,  the finally block will be executed and there are no assertions - the test will just pass.

This test should call `publish()`, catch the expected exception, and may also then check the expected channel state.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -943,4 +943,78 @@ public void detach_success_callback_detached() throws AblyException {
...
+	 * Validate subscribe will result in an error, when the channel is
+	 * in or moves to the FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTL7c
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
```
paddybyers: This test suffers from the same problem as the `publish()` test above: it does not test what it says, and all of the assertions will simply be bypassed by the exception thrown by `subscribe()`.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1664,4 +1663,274 @@ public void onPresenceMessage(PresenceMessage message) {
...
+	 * registered and an error being indicated, when the channel is in or moves to the
+	 * FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTP6c
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
```
paddybyers: Same problem as above.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1664,4 +1663,274 @@ public void onPresenceMessage(PresenceMessage message) {
...
+	 * registered and an error being indicated, when the channel is in or moves
+	 * to the FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTP8d
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
```
paddybyers: Same problem as above.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1664,4 +1663,274 @@ public void onPresenceMessage(PresenceMessage message) {
...
+	 * Validate {@code Presence#get(...)} will result in an error, when the channel
+	 * is in or moves to the FAILED state before the operation succeeds
+	 * </p>
+	 * <p>
+	 * Spec: RTP11b
+	 * </p>
+	 *
+	 * @throws AblyException
+	 */
+	@Test
```
paddybyers: Same problem as above.

paddybyers:
See my comments on the tests. I stopped adding comments because I think they all have the same problem.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
gokhanbarisaker: > The only assertions are checking for channel state, not for any exception from publish(). Also, if there is an exception, the finally block will be executed and there are no assertions - the test will just pass.

Any uncatched exception will make our test fail. No worries there.


> This test should call publish(), catch the expected exception, and may also then check the expected channel state.

I am doing that on [line 975](https://github.com/gokhanbarisaker/ably-java/blob/channel-test-implicitattach/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java#L975)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
gokhanbarisaker: holy copy & paste, my bad. fixing now

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
paddybyers: > I am doing that on line 975

How is that possible? The thing we're expecting is that the publish call (https://github.com/gokhanbarisaker/ably-java/blob/channel-test-implicitattach/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java#L972) throws an exception. How will it reach line 975?

gokhanbarisaker: The test case verifies that,
  In case of fail prune channel (in our case invalid test token) implicit attach call made by `publish(...)` with fail in order `attaching` --> `failed` with ErrorInfo. 

Please correct me if I am wrong. Publish won't throw an exception, but will just fail with `ErrorInfo`

gokhanbarisaker: The `throws AblyException` on publish method signature is for other possible exceptions. I think we need to fail with unexpected exceptions like that.

paddybyers: > Publish won't throw an exception, but will just fail with ErrorInfo

So what does this line do then? https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/realtime/Channel.java#L541

gokhanbarisaker: It gets executed, if we call publish on already failed channel. (post-failed-channel)

Would you like me to add test cases for that also?

paddybyers: > It gets executed, if we call publish on already failed channel.

ok, my mistake.

> Would you like me to add test cases for that also?

Not in this PR, thanks :)

However ... lets look at the comments at the top of this test:

````
Validate publish will result in an error, when the channel is
in or moves to the FAILED state before the operation succeeds
````

The `... when the channel is in ... the FAILED state` part is not part of this test, so please update the comment to reflect that.

Also, this test does not verify that the `publish()` results in an error. If verifies that the channel implicit attach is triggered, and the channel reaches the `failed` state, but it does not look at the result of the `publish()`. Please add a `CompletionListener` to the `publish()` call, and verify that there was an error. Also please review the tests below for the same issue.

paddybyers:
Thanks

PR #117, opened 2016-03-10
Avoid stalled state from previous connection when reusing Realtime.

tcard:
After the connection transitions to CLOSED, some state has to be
reset. The changed test fails if channels are not purged and if
the pending message queue's last processed serial isn't resetted.


mattheworiordan:
LGTM. I wonder if this would effect other libraries as well, I'll raise an issue in Ruby at least.

paddybyers:
LGTM

PR #118, opened 2016-03-11
Update changelog for v0.8.2 release

gokhanbarisaker:


mattheworiordan:
We use semantic versioning and currently our spec is at `0.8.*`, see http://docs.ably.io/client-lib-development-guide/features/.  `0.9.0` is not valid, we are only making patch updates currently and are stick with `0.8.*`

mattheworiordan:
THanks, LGTM

PR #122, opened 2016-03-14
RTN17

gokhanbarisaker:


lib/src/test/java/io/ably/lib/transport/ConnectionManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+		/* Verify that,
+		 *   - connectionManager is connected
+		 *   - connectionManager is connected to the host without any fallback
+		 */
+		assertThat(connectionManager.getConnectionState().state, is(ConnectionState.connected));
+		assertThat(connectionManager.getHost(), is(equalTo(opts.realtimeHost)));
+	}
+
+	@Test
```
paddybyers: This test looks identical to the one above.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -489,10 +488,16 @@ private StateIndication checkSuspend(StateIndication stateChange) {
 
 		/* FIXME: we might want to limit this behaviour to only a specific
 		 * set of error codes */
-		if(pendingConnect != null && checkConnectivity()) {
+		if(pendingConnect != null) {
+			if (!Hosts.isFallback(pendingConnect.host)) {
+				if (!checkConnectivity()) {
+					return new StateIndication(ConnectionState.failed, new ErrorInfo("connection failed", 80000));
```
mattheworiordan: Not sure I understand why we move to a failed state, unless there is an unrecoverable error, the connection should always stay in the suspended state and keep retrying.

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -6,51 +6,61 @@
  * Created by gokhanbarisaker on 2/1/16.
  */
 public class Hosts {
-    private static final List<String> FALLBACKS = Defaults.HOST_FALLBACKS;
-    private static final String PATTERN_REST_PROD = "^rest.ably.io$";
-    private static final String PATTERN_REALTIME_PROD = "^realtime.ably.io$";
+	private static final List<String> FALLBACKS = Defaults.HOST_FALLBACKS;
+	private static final String PATTERN_REST_PROD = "^rest.ably.io$";
```
mattheworiordan: I realise this is old code, but if we're just checking for this pattern why don't we simply use a string and check `host.equals(PATTERN_REST_PROD)`

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -6,51 +6,61 @@
...
+	 * </p>
+	 * <p>
+	 * Spec: RSC15b
+	 * </p>
+	 *
+	 * @param host
+	 * @return true, if the given host is qualified for a retry against a fallback host. Otherwise, false.
+	 */
+	public static boolean isRestFallbackSupported(String host) {
+		return host.matches(PATTERN_REST_PROD);
```
mattheworiordan: as above could be simply `host.equals(REST_PROD_HOST)`

gokhanbarisaker:
rebased

mattheworiordan:
One comment, but otherwise LGTM. @paddybyers?

lib/src/main/java/io/ably/lib/transport/Hosts.java
paddybyers: I agree with this comment - was there a reason not to change it?

gokhanbarisaker: updated the commit

mattheworiordan:
@gokhanbarisaker in future can I ask that you only rebase when the PR is ready to be merged. The reason is simply so that when you address our comments, we can look at the commits that address it easily, whereas now I have re-review the entire PR. See http://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html for more info

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -6,51 +6,68 @@
...
+	 * </p>
+	 * <p>
+	 * Spec: RSC15b
+	 * </p>
+	 *
+	 * @param host
+	 * @return true, if the given host is qualified for a retry against a fallback host. Otherwise, false.
+	 */
+	public static boolean isRestFallbackSupported(String host) {
+		return host.equals(REST_PROD_HOST);
```
mattheworiordan: Sorry to be a pain, but is it worth ignoring the case here?  It would be silly if it didn't fallback over that

mattheworiordan:
LGTM

gokhanbarisaker:
> in future can I ask that you only rebase when the PR is ready to be merged. The reason is simply so that when you address our comments, we can look at the commits that address it easily, whereas now I have re-review the entire PR

C'mon, I have asked a question before about it and you didn't replied :). https://ably-real-time.slack.com/archives/D0HGBQ0F9/p1458313433000050

I will apply rebasing and autosquashing just before I got 2-way LGTM authorization from you and @paddybyers

lib/src/main/java/io/ably/lib/transport/Hosts.java
gokhanbarisaker: Not a problem. done!

mattheworiordan:
> C'mon, I have asked a question before about it and you didn't replied :).

Sorry, there's clearly too much noise anyway.  I am going to fix that partially now by moving Github alerts out of `#client-libraries`.  It does mean client-library developers won't then see updates on other client libraries, but I don't think it matters anyway to them / you.  However, if you then ask a question with an `@` tag, then at least I'll be notified and know I need to go in and respond.

> I will apply rebasing and autosquashing just before I got 2-way LGTM authorization from you and

Thanks, sorry I didn't see your message.  You could have asked again though :)

PR #125, opened 2016-04-03
RTN20 os connectivity events

gokhanbarisaker:
- [x] RTN20a
- [x] RTN20b


gokhanbarisaker:
@mattheworiordan, I couldn't found a way simulating connectivity changes on test cases with this one.

Also, travis had some internal connection problems during [latest build](https://travis-ci.org/ably/ably-java/builds/120835366)

[Here](https://travis-ci.org/gokhanbarisaker/ably-java/builds/120835353) you can see the passing build for the same commit

maven.gradle
```diff
@@ -1,26 +1,43 @@
...
 def groupId = "io.ably"
-def artifactId = "ably-java"
-def localReleaseDest = "${buildDir}/release/${version}"
+def localReleaseDest = "${buildDir}/release/${ably.version}"
 
-allprojects {
-    tasks.withType(Javadoc) {
-        // To prevent javadoc warnings with Java 8
-        options.addStringOption('Xdoclint:none', '-quiet')
+//allprojects {
```
paddybyers: Why commented out?

settings.gradle
```diff
@@ -1,6 +1 @@
-rootProject.name = 'ably-java'
-include 'lib'
-
-if(System.getenv('ANDROID_HOME')) {
-  include 'android-test'
-}
+include ':core', ':android', ':java'
```
paddybyers: Do you think `java` is a better name than `jre` ?

gokhanbarisaker: I wanted to go with artifact prefix as a module name after all.
  - `core` for `ably-core`
  - `java` for `ably-java`
  - `android` for `ably-android`

paddybyers: > I wanted to go with artifact prefix as a module name after all.

fair enough

mattheworiordan:
What is the status on this @gokhanbarisaker?

gokhanbarisaker:
> What is the status on this @gokhanbarisaker?

We want to run current tests on android environment, as well as local java environment that we run current tests on. I have succeeded in doing with following changes,

  - I have updated [travis-ci script](https://github.com/ably/ably-java/pull/125/commits/10aed3072b84817aee09680dee6db455385ee42f#diff-354f30a63fb0907d4ad57269548329e3R6) to make this possible
  - Added a proof of concept [test class](https://github.com/ably/ably-java/pull/125/commits/a56e4bea409286a69eec8c49a2fd924b4e7d820a) with 1 passing and 1 failing test cases


Also we wanted to run our core module tests on android environment, as well as local java environment that we run current tests on. Unfortunately, I couldn't make it happen. Among many other failing attempts, I have also tried our [old solution](https://github.com/ably/ably-java/blob/master/android-test/build.gradle#L14). That gives me following error,

`No tests found. This usually means that your test classes are not in the form that your test runner expects (e.g. don't inherit from TestCase or lack @Test annotations).
:android:connectedDebugAndroidTest FAILED`


Probably I didn't had the enough experience & luck to make it happen, or we were under illusion of successfully running tests on android environment since there was no test to run.


I will suggest not making core module tests run on android environment for now.

gokhanbarisaker:
Also, I have updated the commits for I figured out there has been a code mixup during autosquash

paddybyers:
> or we were under illusion of successfully running tests on android environment since there was no test to run

Previously, explicitly running the `connectedAndroidTest` target would definitely run the tests on the target.  I don't think they were running with `android-test:test` but we should be able to make that happen.

> I will suggest not making core module tests run on android environment for now.

We do need to run tests on the target, because there are too many potential differences in behaviour from the underlying API. Even if `android:test` doesn't work, we definitely need to be able to run them on the target.

gokhanbarisaker:
> Previously, explicitly running the connectedAndroidTest target would definitely run the tests on the target

Then, it is obvious I am lacking experience & luck part :). I have tried both defining `testInstrumentationRunnerArgument` as we did before & importing tests jar with `compile project(path: ':core', configuration: 'testsConfiguration')`

mattheworiordan:
@paddybyers is this something you can help @gokhanbarisaker with then?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -140,6 +140,18 @@ public ConnectionManager(final AblyRealtime ably, Connection connection) {
 			Log.e(getClass().getName(), msg, e);
 			throw new RuntimeException(msg, e);
 		}
+
+		try {
+			connectivityStateChangeEmitter = (ConnectivityStateChangeEmitter) Class.forName("io.ably.lib.transport.ConnectivityStateChangeEmitterImpl").newInstance();
+			connectivityStateChangeEmitter.on(this);
+		} catch (ClassNotFoundException e) {
+			e.printStackTrace();
```
mattheworiordan: Should this not go to the logger?

paddybyers: > Should this not go to the logger?

It's an error that formally must be handled, but it indicates a packaging error with the library rather than any error that can realistically occur in practice. So either we know for sure from our build/CI process that it will never happen, and we just have an empty catch block, or we call the logger. On Android you won't necessarily see `System.err` in logcat, so if you're relying on this message to realise there's a problem in development, it should go to the logger.

mattheworiordan: OK

paddybyers:
Closing as superseded

PR #131, opened 2016-07-15
Ensure graceful handling of DETACH and DISCONNECT.

### LINKED ISSUE #130: Important: Ensure DETACHED or DISCONNECTED with error is non-fatal
See a bug discovered today at https://github.com/ably/ably-js/issues/312 for ably-js.

This is quite problematic because if we explicitly have to DETACH a channel (due to a timeout for example), then the client now will move the channel into the failed state.  This means most customers will never re-attach that channel automatically, see https://www.ably.io/documentation/realtime/connection#handling-failures.

With some urgency, we need to add test coverage to ensure that:

* A server initiated `DETACH` with or without an error moves the channel to the `DETACHED` state and not a failed state
* A server initiated `DISCONNECT` with or without an error moves the connection to the `DISCONNECTED` state and not a failed state.

tcard:
Fixes #130.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -269,7 +269,7 @@ public class ConnectionManager implements Runnable, ConnectListener {
 	 * transport events/notifications
 	 ***************************************/
 
-	void onMessage(ProtocolMessage message) throws AblyException {
+	public void onMessage(ProtocolMessage message) throws AblyException {
 		try {
 			if(protocolListener != null)
 				protocolListener.onRawMessage(message);
@@ -293,6 +293,7 @@ public class ConnectionManager implements Runnable, ConnectListener {
 				case connected:
 					onConnected(message);
 					break;
+				case disconnect:
 				case disconnected:
 					onDisconnected(message);
 					break;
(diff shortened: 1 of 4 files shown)


mattheworiordan:
Nice :) Thanks @tcard

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -269,7 +269,7 @@ public void run() {
 	 * transport events/notifications
 	 ***************************************/
 
-	void onMessage(ProtocolMessage message) throws AblyException {
+	public void onMessage(ProtocolMessage message) throws AblyException {
```
paddybyers: It's a shame we need to make this public just in order to be able to call it from a test. I can't think of an alternative at the moment without mocking the transport so the message comes from there.

paddybyers:
LGTM, thanks

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: > It's a shame we need to make this public just in order to be able to call it from a test. I can't think of an alternative at the moment without mocking the transport so the message comes from there.

I don't know if there is a convention if Java, but in Ruby you can add `@api private` to the method so that at least it's documented clearly that the method is public, but not as part of the public API

PR #132, opened 2016-07-30
Added new header "X-Ably-Lib"

lucky-dev:
Fixed http://docs.ably.io/client-lib-development-guide/features/#RSC7b

To run test: `./gradlew testUtilsSuite`

### DIFF
--- a/lib/src/main/java/io/ably/lib/GlobalConstants.java
+++ b/lib/src/main/java/io/ably/lib/GlobalConstants.java
@@ -1,8 +1,10 @@
 package io.ably.lib;
 
+import io.ably.BuildConfig;
+
 public class GlobalConstants {
 
-    public static final String LIB_VERSION = "0.8.2";
+    public static final String LIB_VERSION = BuildConfig.VERSION;
     public static final String LIB_TYPE = "java";
 
     private GlobalConstants() {
--- a/lib/src/main/java/io/ably/lib/GlobalConstants.java
+++ b/lib/src/main/java/io/ably/lib/GlobalConstants.java
@@ -1,0 +1,11 @@
+package io.ably.lib;
+
+public class GlobalConstants {
+
+    public static final String LIB_VERSION = "0.8.2";
+    public static final String LIB_TYPE = "java";
+
+    private GlobalConstants() {
+    }
+
+}
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -753,5 +753,5 @@ public class Http {
 	private static final String PROXY_AUTHENTICATE  = "Proxy-Authenticate";
 	private static final String AUTHORIZATION       = "Authorization";
 	private static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
-	private static final String X_ABLY_LIB 			= "X-Ably-Lib";
+	private static final String X_ABLY_LIB          = "X-Ably-Lib";
 }
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -424,6 +424,8 @@ public class Http {
 			}
 			if(!acceptSet) { conn.setRequestProperty(ACCEPT, JSON); }
 
+			conn.setRequestProperty(X_ABLY_LIB, headerXAblyLib);
+
 			/* send request body */
 			if(requestBody != null) {
 				writeRequestBody(requestBody, conn);
@@ -711,6 +713,8 @@ public class Http {
 	 *     Private state
 	 *************************/
 
+	static final String headerXAblyLib;
+
 	static {
 		/* if on Android, check version */
 		Field androidVersionField = null;
@@ -723,6 +727,8 @@ public class Http {
 			/* HTTP connection reuse which was buggy pre-froyo */
 			System.setProperty("http.keepAlive", "false");
 		}
+
+		headerXAblyLib = HttpUtils.getHeaderXAblyLib(androidVersionField != null ? "android" : null);
 	}
 
 	final String scheme;
@@ -747,4 +753,5 @@ public class Http {
 	private static final String PROXY_AUTHENTICATE  = "Proxy-Authenticate";
 	private static final String AUTHORIZATION       = "Authorization";
 	private static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
+	private static final String X_ABLY_LIB 			= "X-Ably-Lib";
 }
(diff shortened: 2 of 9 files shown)


lib/src/main/java/io/ably/lib/GlobalConstants.java
```diff
@@ -0,0 +1,11 @@
+package io.ably.lib;
+
+public class GlobalConstants {
+
+    public static final String LIB_VERSION = "0.8.2";
```
mattheworiordan: Perhaps update https://github.com/ably/ably-java/blob/master/README.md as it states that there are version references but does not include a mention of this new one now

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -63,4 +64,10 @@ public static String encodeURIComponent(String input) {
 		} catch (UnsupportedEncodingException e) {}
 		return null;
 	}
+
+	public static String getHeaderXAblyLib(String platform) {
+		return String.format("%s%s-%s", GlobalConstants.LIB_TYPE,
+										((platform == null) || (platform.length() == 0)) ? "" : "." + platform,
```
mattheworiordan: Not sure if this is a Github issue, but tabbing looks very odd here, is it consistent with current tab?

lib/src/test/java/io/ably/lib/test/other/HttpUtilsTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        headerNoPlatform = GlobalConstants.LIB_TYPE + "-" + GlobalConstants.LIB_VERSION;
+        headerAndroidPlatform = GlobalConstants.LIB_TYPE + ".android-" + GlobalConstants.LIB_VERSION;
+    }
+
+    @Test
+    public void testHeaderXAblyLyb() {
+        assertTrue(HttpUtils.getHeaderXAblyLib(null).equals(headerNoPlatform));
+        assertTrue(HttpUtils.getHeaderXAblyLib("android").equals(headerAndroidPlatform));
```
mattheworiordan: Looks good, but is there a way to do a test to check the header that is actually sent in a request?

mattheworiordan:
Thanks, few comments, but looks good

lib/src/main/java/io/ably/lib/http/HttpUtils.java
lucky-dev: It's my tabs. I will remove.

lib/src/test/java/io/ably/lib/test/other/HttpUtilsTest.java
lucky-dev: Need to think about this.

lucky-dev: I think that it can be done in `HttpTest.java`.

lucky-dev:
I added a special gradle plugin to generate the file `BuildConfig.java`. It gives opportunity to store version of the library in one place. E.g. https://github.com/lucky-dev/ably-java/blob/63f51e27b8cbfa842e493471e83efacee23e3c3f/lib/build.gradle#L17

mattheworiordan:
Thanks @lucky-dev seems good.  Did you have any ideas way to test the header is present on the HTTP requests?

lucky-dev:
I think that need to improve class `Http.java`. It will be good to create some listener to get headers before each request. In this method [httpExecute()](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/http/Http.java#L400) call some callback method.

mattheworiordan:
> Maybe in this method httpExecute() call some callback method to notify a client program about headers.

Sure, whatever you think is practical.  The other option is simply to set up a local HTTP server as part of the test and confirm the header is received when setting a custom host to the local HTTP server.

paddybyers:
> It will be good to create some listener to get headers before each request. Maybe in this method httpExecute() call some callback method to notify a client program about headers.

Even if there is a listener that is told about the headers that the client thinks it is about to send, that won't be confirmation that they are actually sent. I would prefer that we do something with `HttpTest` that echoes back the headers received, or something else that confirms they were sent.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -747,4 +753,5 @@ private Proxy getProxy(String host) {
 	private static final String PROXY_AUTHENTICATE  = "Proxy-Authenticate";
 	private static final String AUTHORIZATION       = "Authorization";
 	private static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
+	private static final String X_ABLY_LIB 			= "X-Ably-Lib";
```
paddybyers: Something wrong with whitespace here?

lib/src/test/java/io/ably/lib/test/other/HttpUtilsTest.java
```diff
@@ -0,0 +1,28 @@
...
+    private static String headerAndroidPlatform = null;
+
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        headerNoPlatform = GlobalConstants.LIB_TYPE + "-" + GlobalConstants.LIB_VERSION;
+        headerAndroidPlatform = GlobalConstants.LIB_TYPE + ".android-" + GlobalConstants.LIB_VERSION;
+    }
+
+    @Test
+    public void testHeaderXAblyLyb() {
```
paddybyers: Typo? (lyb)

lib/src/test/java/io/ably/lib/test/other/OtherSuite.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.runner.Result;
+import org.junit.runner.RunWith;
+import org.junit.runner.notification.Failure;
+import org.junit.runners.Suite;
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+        HttpUtilsTest.class,
+})
+public class OtherSuite {
```
paddybyers: I think it would be ok for these tests to be part of the rest suite, since they're tests for the http primitive. If it is a separate suite, the lets call it `UtilsSuite` and not `OtherSuite`.

lib/src/test/java/io/ably/lib/test/other/HttpUtilsTest.java
lucky-dev: Yes

lib/src/test/java/io/ably/lib/test/other/OtherSuite.java
lucky-dev: > If it is a separate suite, the lets call it UtilsSuite and not OtherSuite.
It's a good idea.

lib/src/main/java/io/ably/lib/http/Http.java
lucky-dev: I will fix.

PR #135, opened 2016-08-08
Move all channels to FAILED state if connection failed: 

saiber:
Issue (RTN15i): http://docs.ably.io/client-lib-development-guide/features/#RTN15i
If connection state receive failed state for particular connection move all connected channel to failed state.

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -72,10 +72,19 @@ public class AblyRealtime extends AblyRest {
 	public class Channels extends HashMap<String, Channel> {
 		public Channels() {
 			/* remove all channels when the connection is closed, to avoid stalled state */
-			connection.on(ConnectionState.closed, new ConnectionStateListener() {
+			connection.on(new ConnectionStateListener() {
 				@Override
 				public void onConnectionStateChanged(ConnectionStateListener.ConnectionStateChange state) {
-					Channels.this.clear();
+					if (state.current == ConnectionState.closed) {
+						Channels.this.clear();
+					} else if (state.current == ConnectionState.failed) {
+						Iterator iterator = channels.entrySet().iterator();
+						while (iterator.hasNext()) {
+							ProtocolMessage protocolMessage = new ProtocolMessage(ProtocolMessage.Action.error);
+							protocolMessage.error = state.reason;
+							((Channel)((Map.Entry)iterator.next()).getValue()).onChannelMessage(protocolMessage);
+						}
+					}
 				}
 			});
 		}
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
@@ -1,0 +1,8 @@
+package io.ably.lib.test.realtime;
+
+/**
+ * Created by kostiantyn on 8/9/2016.
+ */
+
+public class RealtimeChannelFailedTest {
+}
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
@@ -1,8 +1,60 @@
 package io.ably.lib.test.realtime;
 
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import io.ably.lib.realtime.AblyRealtime;
+import io.ably.lib.realtime.Channel;
+import io.ably.lib.realtime.ChannelState;
+import io.ably.lib.realtime.ChannelStateListener;
+import io.ably.lib.realtime.ConnectionState;
+import io.ably.lib.realtime.ConnectionStateListener;
+import io.ably.lib.test.common.Helpers;
+import io.ably.lib.test.common.Setup;
+import io.ably.lib.types.AblyException;
+import io.ably.lib.types.ClientOptions;
+import io.ably.lib.types.ErrorInfo;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
 /**
  * Created by kostiantyn on 8/9/2016.
  */
 
 public class RealtimeChannelFailedTest {
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        Setup.getTestVars();
+    }
+
+    @AfterClass
+    public static void tearDownAfterClass() throws Exception {
+        Setup.clearTestVars();
+    }
+
+    @Test
+    public void channel_filed_state_test(){
+        AblyRealtime ablyRealtime = null;
+        try {
+            Setup.TestVars testVars = Setup.getTestVars();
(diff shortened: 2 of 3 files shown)


mattheworiordan:
Thanks for this. Before I comment on the actual implementation, how am I to know if this works?

saiber:
When the connection state is changed and server send a message with FAILED state it move all attached channels to "detached" state as well, as I understood FAILED state should appears before channel are moved to "detached" state. I'm still working and try to write some test

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -72,10 +72,19 @@ public void close() {
 	public class Channels extends HashMap<String, Channel> {
 		public Channels() {
 			/* remove all channels when the connection is closed, to avoid stalled state */
-			connection.on(ConnectionState.closed, new ConnectionStateListener() {
+			connection.on(new ConnectionStateListener() {
```
tcard: This was also a problem with what was there before so probably can be fixed in a new PR, but if you call `off()` (without arguments) on the `Connection` ([which calls this](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/util/EventEmitter.java#L24-L27)) this listener will be removed and thus this logic will be broken. I fixed this in iOS [by having an internal EventEmitter](https://github.com/ably/ably-ios/blob/master/Source/ARTRealtime.m#L68).

saiber: Thank for your comment, but I'm not sure that I'm well understand, that the listener will be removed, do you mean I should remove all listeners?

mattheworiordan: @tcard sorry, also not too clear what you mean here?

tcard: Let's say you do this:

```
AblyRealtime realtime = new AblyRealtime(...);
realtime.connection.on(someListener);
realtime.connection.on(someOtherListener);

// Now I disable all my listeners, for example because my app goes to the background or whatever.
realtime.connection.off();
```

As a side effect, I've also disabled the listener registered [here](https://github.com/ably/ably-java/pull/135/files/c5a158e96d7dd0d73e35cbdf18d920ef30d2cd8b#diff-b8b1ac0a34230df724c2576f32005521R75), so if the connection state transitions to FAILED, channels won't transition to FAILED anymore.

What I've done is to have a public EventEmitter, the one that gets called when you do `realtime.connection.on(...)`, and a private/internal EventEmitter for internal logic like this, so that operations on the public one don't affect internal functionality.

mattheworiordan: Ah, that's interesting.  I suspect that affects a lot of libraries in fact.

I will raise a separate issue to cover that in docs.  @saiber I don't think you should worry about that though for this PR as it's a separate issue that needs addressing on its own.

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -72,10 +72,19 @@ public void close() {
...
 				public void onConnectionStateChanged(ConnectionStateListener.ConnectionStateChange state) {
-					Channels.this.clear();
+					if (state.current == ConnectionState.closed) {
+						Channels.this.clear();
+					} else if (state.current == ConnectionState.failed) {
+						Iterator iterator = channels.entrySet().iterator();
+						while (iterator.hasNext()) {
+							ProtocolMessage protocolMessage = new ProtocolMessage(ProtocolMessage.Action.error);
+							protocolMessage.error = state.reason;
+							((Channel)((Map.Entry)iterator.next()).getValue()).onChannelMessage(protocolMessage);
```
paddybyers: I think we should call `Channel.setState()` here instead of this.

saiber: private void setState(ChannelState newState, ErrorInfo reason)

saiber:
The Channel.setState() is private method, a normal way to set channel in different state is a call onChannelMessage, am I right? thank

mattheworiordan:
> The Channel.setState() is private method, a normal way to set channel in different state is a call onChannelMessage, am I right? 

Not sure, that's a Java specific implementation detail.  How are other parts of the system doing it now?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
```diff
@@ -1,8 +1,60 @@
...
+        Setup.getTestVars();
+    }
+
+    @AfterClass
+    public static void tearDownAfterClass() throws Exception {
+        Setup.clearTestVars();
+    }
+
+    @Test
+    public void channel_filed_state_test(){
```
mattheworiordan: What does this name mean?  I expect it's the typo that is confusing things, but it's still not describing what about the failed state you are testing.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
```diff
@@ -1,8 +1,60 @@
...
+        AblyRealtime ablyRealtime = null;
+        try {
+            Setup.TestVars testVars = Setup.getTestVars();
+            ClientOptions clientOptions = testVars.createOptions("not_an_app.invalid_key_id:invalid_key_value");
+            ablyRealtime = new AblyRealtime(clientOptions);
+            final Channel channel = ablyRealtime.channels.get("test_channel");
+
+            channel.on(new ChannelStateListener() {
```
mattheworiordan: You have not asserted that the connection failing has triggered this state, which is part of the spec.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
```diff
@@ -1,8 +1,60 @@
...
+                @Override
+                public void onChannelStateChanged(ChannelState state, ErrorInfo reason) {
+                    assertEquals("Verify filed state on channel ", channel.state, ChannelState.failed);
```
mattheworiordan: More typos

mattheworiordan:
Why did you create a whole new test suite?
Also, it would be useful to have some reference of what spec this test is actually solving.

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
mattheworiordan: You are sending fabricated failed ProtocolMessages to all channels, even if they are initialized or failed.  This is wrong.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelFailedTest.java
saiber: Yes it is typo, about Channel state, there is one possible way to set channel in different state is to call "onChannelMessage" method in Channel class, all other method in Channel class like "setState", "setFailed" etc can not be called outside the class instance, by the way the "ChannelState state" member of the Channel class is a public I think it is wrong implementation, and "ChannelState state" should be private.

saiber:
As example for connection state, but of course can be moved to RealtimeChannelTest

mattheworiordan:
As mentioned to Vlad, I don't think we'll proceed with this.  Thanks @saiber so far.

PR #137, opened 2016-08-09
Ensure that messages with invalid data type are rejected.

### LINKED ISSUE #112: Library doesn't seem to serialise Map objects properly
I'm using the following code to send a presence update:

    Map<String,Boolean> payload = new HashMap<>();
    payload.put("isTyping",false);
    try {
        sessionChannel.presence.update(payload,null);
    } catch (AblyException e) {
        e.printStackTrace();
    }

This translates to the following MsgPack hex:
`89A6616374696F6E0EA26964AD4D3754536F6A4A5A35532D3331A5636F756E7415A76368616E6E656CAB6D6F62696C653A63686174AD6368616E6E656C53657269616CB0396335333838303037353736313A3733AC636F6E6E656374696F6E4964AA4D3754536F6A4A5A3553B0636F6E6E656374696F6E53657269616C2CA974696D657374616D70CF000001533354B1D2A870726573656E63659184A8636C69656E744964A573616E6479AC636F6E6E656374696F6E4964AA4D3754536F6A4A5A3553A6616374696F6E04A464617461B07B6973547970696E673D66616C73657D`

Which translates to:
`{"action":14,"id":"M7TSojJZ5S-31","count":21,"channel":"mobile:chat","channelSerial":"9c53880075761:73","connectionId":"M7TSojJZ5S","connectionSerial":44,"timestamp":null,"presence":[{"clientId":"sandy","connectionId":"M7TSojJZ5S","action":4,"data":"{isTyping=false}"}]}
`.

However, when ably.js does a similar thing (I'm testing via the Heroku app), the MsgPack hex is:

`89A6616374696F6E0EA26964AD6E554158777A664E33522D3232A5636F756E7412A76368616E6E656CAB6D6F62696C653A63686174AD6368616E6E656C53657269616CB0396335333838303037353736313A3737AC636F6E6E656374696F6E4964AA6E554158777A664E3352B0636F6E6E656374696F6E53657269616C30A974696D657374616D70CF000001533359AAE1A870726573656E63659185A8636C69656E744964A27363AC636F6E6E656374696F6E4964AA6E554158777A664E3352A6616374696F6E04A8656E636F64696E67A46A736F6EA464617461B27B226973547970696E67223A66616C73657D`

Which translates to:
`{"action":14,"id":"nUAXwzfN3R-22","count":18,"channel":"mobile:chat","channelSerial":"9c53880075761:77","connectionId":"nUAXwzfN3R","connectionSerial":48,"timestamp":null,"presence":[{"clientId":"sc","connectionId":"nUAXwzfN3R","action":4,"encoding":"json","data":"{\"isTyping\":false}"}]}
`

The Java lib doesn't seem to create a valid JSON (IIRC JSON doesn't contain equals signs), that the JS lib can successfully parse. I set breakpoints in the JS presence handler and they do not trigger.
As a consequence of these differences, the Cordova sample app does not recognise presence messages sent from the Android sample app. 
The inverse is true - the Android lib successfully parses and displays "isTyping" presence updates coming from JS.

@mattheworiordan @paddybyers @gokhanbarisaker

tcard:
Fixes #112.

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/BaseMessage.java
+++ b/lib/src/main/java/io/ably/lib/types/BaseMessage.java
@@ -17,6 +17,7 @@ import com.google.gson.JsonSerializationContext;
 
 import io.ably.lib.util.Base64Coder;
 import io.ably.lib.util.Crypto.ChannelCipher;
+import io.ably.lib.util.Log;
 import io.ably.lib.util.Serialisation;
 
 public class BaseMessage implements Cloneable {
@@ -115,7 +116,11 @@ public class BaseMessage implements Cloneable {
 				encoding = ((encoding == null) ? "" : encoding + "/") + "utf-8";
 			}
 		} else if(!(data instanceof byte[])) {
-			throw AblyException.fromErrorInfo(new ErrorInfo("Unable to encode message data (incompatible type)", 400, 40000));
+			if (opts != null && opts.encrypted) {
+				throw AblyException.fromErrorInfo(new ErrorInfo("Invalid message data or encoding", 400, 40013));
+			} else {
+				Log.e(TAG, "Message data must be either `byte[]`, `String` or `JSONElement`; implicit coercion of other types to String is deprecated and throws from v.0.9 on.\nPlease check the documentation (https://www.ably.io/documentation/realtime/types#message).");
+			}
 		}
 		if (opts != null && opts.encrypted) {
 			ChannelCipher cipher = opts.getCipher();
@@ -220,4 +225,6 @@ public class BaseMessage implements Cloneable {
 			}
 		}
 	}
+
+	private static final String TAG = BaseMessage.class.getName();
 }
--- a/lib/src/main/java/io/ably/lib/types/BaseMessage.java
+++ b/lib/src/main/java/io/ably/lib/types/BaseMessage.java
@@ -109,16 +109,17 @@ public class BaseMessage implements Cloneable {
 			data = Serialisation.gson.toJson((JsonElement)data);
 			encoding = ((encoding == null) ? "" : encoding + "/") + "json";
 		}
-		if(opts != null && opts.encrypted) {
-			if(data instanceof String) {
+		if(data instanceof String) {
+			if (opts != null && opts.encrypted) {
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -109,16 +109,17 @@ public void encode(ChannelOptions opts) throws AblyException {
...
+		if(data instanceof String) {
+			if (opts != null && opts.encrypted) {
 				try { data = ((String)data).getBytes("UTF-8"); } catch(UnsupportedEncodingException e) {}
 				encoding = ((encoding == null) ? "" : encoding + "/") + "utf-8";
 			}
-			if(!(data instanceof byte[])) {
-				throw AblyException.fromErrorInfo(new ErrorInfo("Unable to encode message data (incompatible type)", 400, 40000));
-			}
+		} else if(!(data instanceof byte[])) {
+			throw AblyException.fromErrorInfo(new ErrorInfo("Unable to encode message data (incompatible type)", 400, 40000));
```
mattheworiordan: Should this not be "Invalid message data or encoding", see https://github.com/ably/ably-common/blob/master/protocol/errors.json#L19

mattheworiordan:
One issue but otherwise LGTM 👍

paddybyers:
LGTM but I'm concerned that this will cause compatibility issues if anyone is unintentionally depending on the library converting the given data to a string.

Should we consider perhaps instead of an exception we might coerce to a string and log an error message saying

```
ably-java: Message data must be either `byte[]`, `String` or `JSONElement`; implicit coercion of other types to String is deprecated and throws from v.0.9 on.
Please check the documentation (https://www.ably.io/documentation/realtime/types#message).
```

Note also that the documentation at that link says is a bit messed up as it says the data can be a StringBuffer in Javascript, a JSON Object in Javascript (it should be Object I think).

tcard:
PTAL

paddybyers:
LGTM/ @mattheworiordan you ok with this?

mattheworiordan:
Sorry, forgot to 👍 

@tcard please can you add an issue to remove the deprecation notice in 0.9 and reject the message when you merge this though?

tcard:
@mattheworiordan Done: #139

PR #138, opened 2016-08-10
Update README.md

hauleth:
Execute Gradle with custom environment in shell-independent way (this will also work in fish) and without polluting global environment.


paddybyers:
LGTM

mattheworiordan:
Nice, thanks @hauleth

PR #140, opened 2016-08-11
Ensure interoperability with other libraries over JSON.

tcard:
We were testing MsgPack only as protocol encoding.


mattheworiordan:
👍

PR #141, opened 2016-08-16
Added custom fallback hosts support

piotrekinventica:
Added missing functionality and tests for http://docs.ably.io/client-lib-development-guide/features/#RSC15a 
run tests with `gradle testTransportSuite`

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -11,6 +11,7 @@ import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -327,6 +328,7 @@ public class Http {
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
 		String candidateHost = this.host;
+		Collections.shuffle(options.fallbackHosts);
 		URL url;
 
 		while(true) {
@@ -338,7 +340,7 @@ public class Http {
 					throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
 				}
 				Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
-				candidateHost = Hosts.getFallback(candidateHost);
+				candidateHost = Hosts.getFallback(candidateHost, options.fallbackHosts);
 				Log.d(TAG, "Switched to `" + candidateHost + "`.");
 			}
 		}
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -328,7 +328,6 @@ public class Http {
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
 		String candidateHost = this.host;
-		Collections.shuffle(options.fallbackHosts);
 		URL url;
 
 		while(true) {
@@ -340,7 +339,7 @@ public class Http {
 					throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
 				}
 				Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
-				candidateHost = Hosts.getFallback(candidateHost, options.fallbackHosts);
+				candidateHost = Hosts.getFallback(candidateHost, options.getFallbackHosts());
 				Log.d(TAG, "Switched to `" + candidateHost + "`.");
 			}
 		}
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -327,6 +327,8 @@ public class Http {
 	 */
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
+		if(options.fallbackHosts != null)
+			Collections.shuffle(options.fallbackHosts);
 		String candidateHost = this.host;
 		URL url;
 
@@ -339,8 +341,10 @@ public class Http {
 					throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
 				}
 				Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
-				candidateHost = Hosts.getFallback(candidateHost, options.getFallbackHosts());
+				candidateHost = Hosts.getFallback(candidateHost, options.fallbackHosts);
 				Log.d(TAG, "Switched to `" + candidateHost + "`.");
+				if(candidateHost == null)
+					throw AblyException.fromErrorInfo(new ErrorInfo("Connection failed; no host available", 404, 80000));
 			}
 		}
 	}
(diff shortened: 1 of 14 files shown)


lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -8,7 +8,7 @@
 
 public class Defaults {
 	public static final int PROTOCOL_VERSION        = 1;
-	static final List<String> HOST_FALLBACKS        = Arrays.asList("A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com");
+	public static final List<String> HOST_FALLBACKS        = Arrays.asList("A.ably-realtime.com", "B.ably-realtime.com", "C.ably-realtime.com", "D.ably-realtime.com", "E.ably-realtime.com");
```
mattheworiordan: Minor point, but spacing is wrong here

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -2,11 +2,13 @@
...
 import java.util.List;
 
+import io.ably.lib.types.ClientOptions;
+
 /**
  * Created by gokhanbarisaker on 2/1/16.
  */
 public class Hosts {
-	private static final List<String> FALLBACKS = Defaults.HOST_FALLBACKS;
+	//private static final List<String> FALLBACKS = Defaults.HOST_FALLBACKS;
```
mattheworiordan: Any reason to commit commented out code?  We prefer to avoid this

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -19,16 +21,23 @@ private Hosts() { /* Restrict new instance creation */ }
...
 	 */
-	public static String getFallback(String host) {
-		int size = FALLBACKS.size();
-		int indexCurrent = FALLBACKS.indexOf(host);
+	public static String getFallback(String host, List<String> fallbackHosts) {
+
+		if(fallbackHosts == null)
+			return null;
+
+		int size = fallbackHosts.size();
```
mattheworiordan: I don't see where the hosts are randomised, `RSC15a` stats `try fallback hosts in random order`.  In other libraries, at the time the instance is created, the hosts are randomised once and persisted in random order.  If that is done, then all other logic will continue to work as is.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -65,7 +66,7 @@ public void hosts_fallback_null_array(){
 
     @Test
     public void hosts_fallback_empty_array(){
-        String host = Hosts.getFallback(null, new ArrayList<>());
+        String host = Hosts.getFallback(null, Collections.<String>emptyList());
```
mattheworiordan: This code was written in the previous commit so should be squished unless this introduces some new concept I believe.

lib/src/test/java/io/ably/lib/transport/TransportSuite.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Created by piotrkazmierczak on 16.08.16.
+ */
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+        ConnectionManagerTest.class,
+        HostsTest.class
+})
+public class TransportSuite {
```
mattheworiordan: Why a new suite?

mattheworiordan:
Not sure what's happened, but you have two commits ec50456 and 574a0ca which appear (at least in Github) to be the same?

lib/src/test/java/io/ably/lib/transport/TransportSuite.java
```diff
@@ -9,7 +9,7 @@
 /**
  * Created by piotrkazmierczak on 16.08.16.
  */
-@RunWith(Suite.class)
+@RunWith(Suite.class)git
```
mattheworiordan: Invalid commit?

mattheworiordan:
c3b4662 is an empty commit as far as I can see?   Why is it included?

mattheworiordan:
What is merge commit 07278d6 for?

mattheworiordan:
Thanks for the PR, I've added various comments.  Please can you reply so that I can understand some of thinking behind it.

mattheworiordan:
In `RSC15a`, the spec states `If an array of custom fallback hosts are provided in ClientOptions#fallbackHosts, then they will be used instead. If an empty array of fallback hosts is provided, then fallback host functionality is disabled`.  The tests you have added are unit tests against an effectively private object (inner workings) of the client library. As such, there is no test coverage that confirms the new requirements of `RSC15a` is met?

piotrekinventica:
@mattheworiordan I've cleared commit history, so there won't be any confusion about merges and unnecessary commits, I completely forgot to squash commits before creating PR.
I also added randomization using `Collections.shuffle()`, I thought I saw it the first time I was implementing my solution in Defaults.java but it was only for default hosts and done once. 
I corrected the spacing and added test for custom fallback hosts, to check if the output list is contained in any order in the original list.
And I agree, new Suite might be an overkill, I originally also added ConnectionManagerTest to the Suite but second test from that class would not complete.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -327,6 +328,7 @@ public void finalize() {
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
 		String candidateHost = this.host;
+		Collections.shuffle(options.fallbackHosts);
```
mattheworiordan: As above, I think we should shuffle just once, probably best when `options.fallbackHosts` is configured

lib/src/main/java/io/ably/lib/transport/Hosts.java
mattheworiordan: Why is this still committed?

mattheworiordan:
> In RSC15a, the spec states If an array of custom fallback hosts are provided in ClientOptions#fallbackHosts, then they will be used instead. If an empty array of fallback hosts is provided, then fallback host functionality is disabled. The tests you have added are unit tests against an effectively private object (inner workings) of the client library. As such, there is no test coverage that confirms the new requirements of RSC15a is met?

Thanks @piotrekinventica but I still don't see a test that demonstrates thats providing a custom set of fallback hosts in `ClientOptions` results in the fallback hosts being used

piotrekinventica:
@mattheworiordan I've changed ClientOptions#fallbackHost field to private and added getters/setters to track if it's being changed, so it'll shuffle only once. I also added tests for getting fallback hosts using ClientOptions

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -83,4 +85,58 @@ public void hosts_fallback_custom_hosts() {
...
+        assertThat(host, is(equalTo(null)));
+    }
+
+    /**
+     * Expect that returned host is null when setting null as fallback host list
+     */
+    @Test
+    public void hosts_fallback_null_custom_hosts_options(){
+        ClientOptions options = new ClientOptions();
+        options.setFallbackHosts(null);
```
mattheworiordan: Sorry, but seems we are not communicating clearly, perhaps my fault.

Adding a setter / getter is not the solution given the IDL explicitly defines the attribute as `fallbackHosts`, see http://docs.ably.io/client-lib-development-guide/features/#TO3k6.  Unless there is a very good reason to deviate from the API that all client libraries use, we do not do this.  Also, adding a getter & setter is unnecessary given you could simply shuffle the `fallbackHosts` after the Client is initialised.  Additionally unfortunately with these changes the shuffling does not apply to the default fallback hosts now, see https://github.com/ably/ably-java/pull/141/commits/76d931f545c6cc588f67b42d0554c0e9a2b13677#diff-f111f958256bd5375e3c22f2b1fed74bR142.  Finally, I still do not think this test shows an end-to-end type test that we expect.  For example, how do I know that the client library actually uses those fallback hosts.  You are still testing the use of `Host.getFallback` to check if a fallback host is used, yet that is an internal implementation API.  So there is no guarantee that the fallback host is ever used unfortunately when the client library is actually unable to use the primary host.

piotrekinventica:
@mattheworiordan I reverted my changes with ClientOptions#fallbackHosts back to the public field and shuffling it before every usage, that way we can be sure that every time someone use this field, it'll be random. If we randomize it only on the init, it can be overwritten later in the code since it's a public field and host order won't be random anymore. It also ensures that default hosts will also be randomized. 
I added 404 exception thrown when null fallback host is returned (which means all fallback options are exhausted or we disabled them), I think there is no need to keep incrementing retry count, when candidate host 100% will not work.
I also added tests for fallbacks connected directly to API and fixed one of the existing ones in HttpTest.java. Tests contain ClientOptions#fallbackHosts set to null and set to array with custom hosts.

trenouf:
Superseded by https://github.com/ably/ably-java/pull/176

PR #143, opened 2016-08-18
Header X-Ably-Lib (RSC7b)

VOstopolets:


build.gradle
```diff
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.1'
+        classpath 'com.android.tools.build:gradle:2.1.3'
```
paddybyers: I have to say I would have preferred that changes like this to the build configuration are kept separate from PRs for other features.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -424,6 +424,9 @@ public void finalize() {
 			}
 			if(!acceptSet) { conn.setRequestProperty(ACCEPT, JSON); }
 
+			/* pass required headers */
+			conn.setRequestProperty(HttpUtils.X_ABLY_LIB_HEADER, HttpUtils.getAblyLibValue());
```
paddybyers: We should statically format the header string once and use that value each time the header is added, instead of formatting the string each time there is a request.

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -13,6 +14,9 @@
  *
  */
 public class HttpUtils {
+	// Headers
```
paddybyers: Please use a comment style consistent with the existing code, ie `/* ..... */`

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -63,4 +67,27 @@ public static String encodeURIComponent(String input) {
...
+	 * -[version] is the full client library version (ex.0.9.2).
+	 * <p/>
+	 * For example, Javascript library would use the header X-Ably-Lib: js-0.9.2
+	 * <p/>
+	 * <p/>
+	 * <p>
+	 * Spec: RSC7b
+	 * </p>
+	 */
+	public static Param getAblyLibHeader() {
```
paddybyers: Is this function used?

lib/src/test/java/io/ably/lib/http/HttpHeaderTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private static SessionHandlerNanoHTTPD server;
+
+    @BeforeClass
+    public static void setUp() throws IOException {
+        // create custom RouterNanoHTTPD class for getting session object
+        server = new SessionHandlerNanoHTTPD(27331);
+        server.start(NanoHTTPD.SOCKET_READ_TIMEOUT, true);
+
+        while (!server.wasStarted()) {
```
paddybyers: Can you not synchronise explicitly to do this instead of poll?

lib/src/test/java/io/ably/lib/http/HttpHeaderTest.java
```diff
@@ -0,0 +1,197 @@
...
+            try {
+                ably.http.httpExecuteWithRetry(
+                        url, /* Ignore */
+                        Http.GET, /* Ignore */
+                        new Param[0], /* Ignore */
+                        requestBody, /* Ignore */
+                        mock(Http.ResponseHandler.class), /* Ignore */
+                        true
+                );
+            } catch (AblyException ex) {
```
paddybyers: Shouldn't we fail the test if an exception is caught here?

lib/src/test/java/io/ably/lib/http/HttpHeaderTest.java
```diff
@@ -0,0 +1,197 @@
...
+        public Response serve(IHTTPSession session) {
+            result = session;
+            return super.serve(session);
+        }
+
+        public IHTTPSession getResult() {
+            return result;
+        }
+    }
+}
```
paddybyers: Newline needed

lib/src/test/java/io/ably/lib/test/common/Helpers.java
```diff
@@ -557,7 +557,7 @@ public synchronized void onSuccess(T result) {
 		}
 
 		@Override
-		public void onError(ErrorInfo error) {
+		public synchronized void onError(ErrorInfo error) {
```
paddybyers: Why was this added in this PR?

paddybyers:
Thanks - please see my comments.

build.gradle
mattheworiordan: @paddybyers but isn't this change specifically to address the requirement of using the version from `BuildConfig`, see https://github.com/ably/ably-java/pull/143/commits/10e3948d290d71b6af9add400e0f59c01a6d1a61#diff-5d450029acab7ff9e9ffca5408321058L20?

mattheworiordan:
Thanks @VOstopolets, this is much better, however I am still concerned that the test still does not demonstrate using a public API of the library that HTTP requests made for all REST operations use the header.  I appreciate you are using an internal API, which may seemingly prove that, but it doesn't because as the reader of the test I am unable to see that for example requesting a token or publishing a message will include the header.  Also, if the internal implementation changes in the future of the library, this test may still pass whilst the rest of the library starts using a different internal method.  Our tests are here to prove that the [public defined API](http://docs.ably.io/client-lib-development-guide/features/#idl) is doing its job correctly and also to prevent regressions.  Unfortunately I don't think this PR delivers on either of those goals.

lib/src/test/java/io/ably/lib/http/HttpHeaderTest.java
VOstopolets: What do you mean? I get this code from existing Http.class https://github.com/ably/ably-java/blob/master/lib/src/test/java/io/ably/lib/http/HttpTest.java#L46-L64

lib/src/test/java/io/ably/lib/test/common/Helpers.java
VOstopolets: I have caught java.lang.IllegalMonitorStateException when notify() invoked at this place. I think is correct implementation and must be used at this place.

mattheworiordan:
Thanks @VOstopolets, looking much better 👍

Issue #133, opened 2016-08-04
A post-suspend clean connection removes all channels instead of moving them to DETACHED


paddybyers:
https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L326

PR #146, opened 2016-08-24
Detach on suspend

paddybyers:
Fixes: https://github.com/ably/ably-java/issues/133


mattheworiordan:
Lgtm me @paddybyers. Wdyt about adding a new spec for 1.0 that simply ensures listeners work when connections and/or channels move through terminal states. It would be easy enough to add a simple test and will help ensure we don't get a gotcha like this in other libraries?

paddybyers:
> Wdyt about adding a new spec for 1.0 that simply ensures listeners work when connections and/or channels move through terminal states. It would be easy enough to add a simple test and will help ensure we don't get a gotcha like this in other libraries?

Yes we should do this.

PR #147, opened 2016-08-25
RTN2g: Param "Lib" with header value (RSC7b)

VOstopolets:


lib/src/test/java/io/ably/lib/transport/ConnectionManagerTest.java
```diff
@@ -198,4 +198,35 @@ public void connectionmanager_reconnect_default_endpoint() throws AblyException
...
+	public void connectionmanager_param_lib() throws AblyException {
+		Setup.TestVars testVars = Setup.getTestVars();
+		ClientOptions opts = testVars.createOptions(testVars.keys[0].keyStr);
+		AblyRealtime ably = new AblyRealtime(opts);
+		ConnectionManager connectionManager = ably.connection.connectionManager;
+
+		new Helpers.ConnectionManagerWaiter(connectionManager).waitFor(ConnectionState.connected);
+
+		/* Get X-Ably-Lib param */
+		Param[] params = connectionManager.getConnectParams();
```
mattheworiordan: I am not entirely comfortable that this "really" tests that the connection URL includes the param, it simply tests that `getConnectParams()` returns the expected params.  If for example someone later changed the wS connection mechanic to no longer us this method, the tests would continue to pass even if the lib param may not be used in the URL.  Is there no other way you can think of testing this?  For example, you could run a simple web server that accepts the connection request, it will fail because the upgrade won't happen, but you can inspect the HTTP request to make sure the URL includes the param for example.  WDYT?

VOstopolets: Do you mean that params should be checked in local test server like in RSC7b feature?

mattheworiordan: Well yes if that is the best way.  The point is that we're testing internals hear, not behaviour.  So if you can think of any other way to test the behaviour of the connection then please do that, but failing that, a local test server should work?

lib/src/test/java/io/ably/lib/transport/ConnectionParamTest.java
```diff
@@ -0,0 +1,123 @@
...
+        Map<String, String> params = server.getParams();
+
+        /* Verify that params exists */
+        assertNotNull("Expected params", params);
+
+        /* Get Lib param */
+        String ablyLibParam = params.get(ITransport.TransportParams.LIB_PARAM_KEY);
+
+		/* Verify that,
+         *   - Lib param exists
```
paddybyers: Indentation?

VOstopolets:

trenouf:
Don't merge yet; I can't run any tests on sandbox.

trenouf:
Tests all ok for me now (apart from ones that often fail intermittently anyway), so this is ready to go.

PR #149, opened 2016-08-31
Reauth capability

VOstopolets:
Reauth with token details

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -324,7 +324,16 @@ public class Auth {
 	 * @param callback (err, tokenDetails)
 	 */
 	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
-		return tokenAuth.authorise(options, params);
+		TokenDetails tokenDetails = tokenAuth.authorise(options, params);
+
+		/* RTC8
+		 *  If authorise is called with AuthOptions#force set to true
+		 *  the client will obtain a new token, disconnect the current transport
+		 *  and resume the connection
+		 */
+		if (options != null && options.force)
+			ably.onAuthUpdated();
+		return tokenDetails;
 	}
 
 	/**
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -271,6 +271,33 @@ public class ConnectionManager implements Runnable, ConnectListener {
 		}
 	}
 
+	/**
+	 * Authentication token changes for the current connection are possible when
+	 * the client is in the {@link ConnectionState#connected}, {@link ConnectionState#connecting}
+	 * or {@link ConnectionState#disconnected } states.
+	 *
+	 * In the current protocol version we are not able to update auth params on the fly;
+	 * so disconnect, and the new auth params will be used for subsequent reconnection
+	 * * <p>
+	 * Spec: RTC8 (reauthorise) (0.8 spec)
+	 * </p>
+	 *
+	 */
+	public void onAuthUpdated() {
+		switch (state.state){
+			case connected:
+			case connecting:
+				ITransport transport = this.transport;
+				if (transport != null) {
+					Log.v(TAG, "onAuthUpdated: closing transport");
+					transport.close(/*sendDisconnect=*/false);
+				}
+				break;
+			default:
+				break;
+		}
+	}
+
 	/***************************************
 	 * transport events/notifications
 	 ***************************************/
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
@@ -100,6 +100,13 @@ public class RealtimeReauthTest {
 			Auth.TokenDetails reauthTokenDetails = ablyRealtime.auth.authorise(authOptions, null);
 			assertNotNull("Expected token value", reauthTokenDetails.token);
 			System.out.println("done reauthorise");
+			/* Delay 2s to allow connection to go disconnected (and probably
+			 * then onto connecting and connected). This is a workaround for
+			 * https://github.com/ably/ably-java/issues/180 */
+			try {
+				Thread.sleep(2000);
+			} catch (Exception e) {
+			}
 
 			/* re-attach to the channel */
 			waiter = new Helpers.CompletionWaiter();
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
@@ -1,0 +1,118 @@
+package io.ably.lib.test.realtime;
+
+import org.junit.Test;
+
+import io.ably.lib.realtime.AblyRealtime;
+import io.ably.lib.realtime.Channel;
+import io.ably.lib.realtime.ConnectionState;
+import io.ably.lib.rest.AblyRest;
+import io.ably.lib.rest.Auth;
+import io.ably.lib.test.common.Helpers;
+import io.ably.lib.test.common.Setup;
+import io.ably.lib.types.AblyException;
+import io.ably.lib.types.Capability;
+import io.ably.lib.types.ClientOptions;
+import io.ably.lib.types.ErrorInfo;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+/**
+ * Created by VOstopolets on 8/26/16.
+ */
+public class RealtimeReauthTest {
+
+	/**
+	 * RTC8 (0.8 spec)
+	 *  If authorise is called with AuthOptions#force set to true
+	 *  the client will obtain a new token, disconnect the current transport
+	 *  and resume the connection
+	 *
+	 * use authorise({force: true}) to reauth with a token with a different set of capabilities
+	 */
+	@Test
+	public void reauth_tokenDetails() {
+		String wrongChannel = "wrongchannel";
+		String rightChannel = "rightchannel";
+		String testClientId = "testClientId";
+
+		try {
+			/* init ably for token */
+			final Setup.TestVars optsTestVars = Setup.getTestVars();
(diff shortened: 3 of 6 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealitimeReauthTest.java
```diff
@@ -0,0 +1,103 @@
...
+            Auth.TokenParams tokenParams = new Auth.TokenParams();
+            Capability capability = new Capability();
+            capability.addResource(wrongChannel, "*");
+            tokenParams.capability = capability.toString();
+            tokenParams.clientId = testClientId;
+
+            Auth.TokenDetails firstToken = ablyForToken.auth.requestToken(tokenParams, null);
+            assertNotNull("Expected token value", firstToken.token);
+
+			/* create ably realtime with tokenDetails and clientId */
```
mattheworiordan: Odd spacing here

lib/src/test/java/io/ably/lib/test/realtime/RealitimeReauthTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Created by VOstopolets on 8/26/16.
+ */
+public class RealitimeReauthTest {
+
+    /**
+     * use authorise({force: true}) to reauth with a token with a different set of capabilities
```
mattheworiordan: It would be useful, if possible, to additionally add the spec item in the tests so that later other people can easily find out which spec(s) the test covers.

mattheworiordan:
Nice, looks good to me. @paddybyers any comments (note that this is not 0.9 reauth, but intentionally 0.8 auth so that we get 0.8 spec done and then move onto 0.9)

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -174,6 +174,17 @@ public void close() {
 		requestState(ConnectionState.closing);
 	}
 
+	/**
```
paddybyers: This should not be here as part of the public API.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -265,6 +276,63 @@ public void run() {
...
+				connection.once(ConnectionState.failed, new ConnectionStateListener() {
+					@Override
+					public void onConnectionStateChanged(ConnectionStateChange state) {
+						synchronized (ConnectionManager.this) {
+							ConnectionManager.this.notify();
+						}
+
+					}
+				});
+				disconnect();
```
paddybyers: Instead of this there should be an explicit call to close the transport (without closing the connection), which will cause the transport to disconnect and be disposed. It should not be initiated by changing connection manager state.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -265,6 +276,63 @@ public void run() {
...
+				synchronized (this){
+					try {
+						wait();
+					} catch (InterruptedException e) {
+						e.printStackTrace();
```
paddybyers: If we think that this case is possible (is there anything in the code that would interrupt this wait?) then this case should include a log message via the `Logger` instead of an explicit `printStackTrace()`. If we don't think this is reachable (ie because nothing interrupts the `wait()`, or only does so in circumstances that mean this is not an error, then we should have an empty catch block with no log statement.

Afaik there may be an interrupt to unblock the waiting thread as part of the shutdown sequence of the library; or it might be by a `notify()` - I can't remember. Either way, I think you should find out and decide what to put here.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -265,6 +276,63 @@ public void run() {
...
+	 * Spec: RTC8 (reauthorise)
+	 * </p>
+	 *
+	 */
+	public void onAuthUpdated() {
+		switch (state.state){
+			case connected:
+			case connecting:
+			case disconnected:
+				connection.once(ConnectionState.failed, new ConnectionStateListener() {
```
paddybyers: I don't see why you're depending on events from the `Connection` object here. This code is the `ConnectionManager`, which is responsible to determining the connection state, and issuing events; the `Connection` class is just the client API that exposes those events and state to the application. This class knows what the state is and should be reacting events and changes in its own state, rather than listening for events on the `Connection`.

Also, I have no idea why you're handling this event and then doing a `notify()` on the `ConnectionManager`. Can you explain why?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -265,6 +276,63 @@ public void run() {
...
+				disconnect();
+				synchronized (this){
+					try {
+						wait();
+					} catch (InterruptedException e) {
+						e.printStackTrace();
+					}
+				}
+				connection.once(ConnectionState.connected, new ConnectionStateListener() {
+					@Override
```
paddybyers: Same comment as above.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -265,6 +276,63 @@ public void run() {
...
+						}
+
+					}
+				});
+				connect();
+				synchronized (this){
+					try {
+						wait();
+					} catch (InterruptedException e) {
+						e.printStackTrace();
```
paddybyers: Same comment as above.

lib/src/test/java/io/ably/lib/test/realtime/RealitimeReauthTest.java
```diff
@@ -0,0 +1,103 @@
...
+            ClientOptions opts = testVars.createOptions();
+            opts.clientId = testClientId;
+            opts.tokenDetails = firstToken;
+            AblyRealtime ablyRealtime = new AblyRealtime(opts);
+
+            /* wait for connected state */
+            Helpers.ConnectionWaiter connectionWaiter = new Helpers.ConnectionWaiter(ablyRealtime.connection);
+            connectionWaiter.waitFor(ConnectionState.connected);
+            assertEquals("Verify connected state is reached", ConnectionState.connected, ablyRealtime.connection.state);
+
```
paddybyers: Please only a single line of whitespace anywhere.

lib/src/test/java/io/ably/lib/test/realtime/RealitimeReauthTest.java
```diff
@@ -0,0 +1,103 @@
...
+            channel.attach(waiter);
+
+            /* verify onSuccess callback gets called */
+            waiter.waitFor();
+            assertThat(waiter.success, is(true));
+        } catch (AblyException e) {
+            e.printStackTrace();
+        }
+    }
+}
```
paddybyers: Trailing newline please.

paddybyers:
I added some comments. I don't think you've understood the different purposes of `ConnectionManager` and `Connection`. Please review my comments and come back if you have any questions.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
VOstopolets: If onAuthUpdated() invoked, I must do reconnect. In ConnectionManager, I use method requestState(ConnectionState state), it invokes close(),abort(),connect() methods for the Transport. So, I invoke disconnect(), lock the thread, and listen when connection state is changed to failed, than notify and do next step (connect). For connect used the same logic. In the end, onAuthUpdated() makes complete cycle of reconnect.

VOstopolets: Where have I made a mistake?

mattheworiordan:
@VOstopolets there are lots of comments here but you've not replied to them. Can you please reply today @VOstopolets ?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: > There's no point doing that in disconnected state is there?

correct

trenouf:
This is ready to go now if nothing too bad happens in travis.

trenouf:
Nothing unexpected failed on travis jdk8.
On jdk7 it got a buffer overflow error on one of the tests that normally fails. That must be a JDK bug.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -324,7 +325,17 @@ public JsonObject asJSON() {
...
 	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
-		return tokenAuth.authorise(options, params);
+		TokenDetails tokenDetails = tokenAuth.authorise(options, params);
+
+		/* RTC8
+		 *  If authorise is called with AuthOptions#force set to true
+		 *  the client will obtain a new token, disconnect the current transport
+		 *  and resume the connection
+		 */
+		if (options != null && options.force && ably instanceof AblyRealtime) {
```
paddybyers: Does this impact our ability to compile a REST-only library? I know we don't do that right now but I don't really like the pollution.

I'm not sure what the cleanest way is to do this, but at a minimum we could just do `ably.getClass().getName().equals( ....) `

lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
```diff
@@ -0,0 +1,126 @@
...
+			opts.tokenDetails = firstToken;
+			AblyRealtime ablyRealtime = new AblyRealtime(opts);
+			System.out.println("done create ably");
+
+			/* wait for connected state */
+			Helpers.ConnectionWaiter connectionWaiter = new Helpers.ConnectionWaiter(ablyRealtime.connection);
+			connectionWaiter.waitFor(ConnectionState.connected);
+			assertEquals("Verify connected state is reached", ConnectionState.connected, ablyRealtime.connection.state);
+			System.out.println("connected");
+
```
paddybyers: Extra whitespace. 1 line only pls.

trenouf: Will be fixed in next push.

lib/src/main/java/io/ably/lib/rest/Auth.java
trenouf: Done in new push.

trenouf:
New push is ready to go.

PR #152, opened 2016-09-03
Project struct update

Pirksy:
- Changed (android) library test structure
- Added some missing dependencies
- Implemented Travis script


android/src/test/resources/local/testAppSpec.json
```diff
@@ -0,0 +1,74 @@
+{
```
mattheworiordan: Why are you committing this file when it already exists in this repo https://github.com/ably/ably-java/blob/master/lib/src/test/resources/local/testAppSpec.json?  From what I can tell, your build can simply copy when running the build instead?  I don't see any good reason to have two copies of the same file in the repo.

mattheworiordan:
Why do you have two commits with the same name and the second cd7c7f0 only has a few lines?

android/src/test/java/io/ably/lib/test/common/android/AssetResourceLoader.java
```diff
@@ -0,0 +1,31 @@
+package io.ably.lib.test.common.android;
```
paddybyers: So this is now identical to the JRE resource loader apart from the path for reading the files - should we instead refactor it so there is just a single common class?

mattheworiordan:
@Pirksy you have not replied to our feedback, please reply.

android/src/test/resources/local/testAppSpec.json
Pirksy: I agree with you, I'll remove this file.

dsvishchov:
As per my recent e-mail:

>> Please disregard that PR already issued. I spent the whole today on this one and made it much clearer from the commit(s) structure, also adjusted according to your comments on GitHub. Also made it run on Android, fixed issues with git submodules (tests have been failing after first Paddy’s commit on that branch due to resources not being loaded pulled from ably-common, now everything is ok), also now it almost works in Travis (I’m using my own account to test it, I can provide more details on the current state if required). I’m continuing to work on this and once I have tests run properly both locally and on Travis for both platforms I’ll issue a first PR for you further review and we’ll continue with the rest of the items required for this. We’re on a right way :)

mattheworiordan:
Ok, please close the PR then it you don't want it to be merged

PR #153, opened 2016-09-04
TO3j11, RSA10g

VOstopolets:


lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -0,0 +1,109 @@
...
+             * storedAuthOptions isn't null,
+             * and storedAuthOptions equals custom authOptions */
+            assertNotNull(storedAuthOptions);
+            assertEquals(storedAuthOptions.authCallback, authOptions.authCallback);
+            assertEquals(storedAuthOptions.authUrl, authOptions.authUrl);
+            assertEquals(storedAuthOptions.token, authOptions.token);
+            assertEquals(storedAuthOptions.key, authOptions.key);
+            assertEquals(storedAuthOptions.tokenDetails, authOptions.tokenDetails);
+            assertArrayEquals(storedAuthOptions.authHeaders, authOptions.authHeaders);
+            assertArrayEquals(storedAuthOptions.authParams, authOptions.authParams);
```
mattheworiordan: Checking store auth options and stored token params is nice, but the spec states "Stores the AuthOptions and TokenParams arguments as defaults for subsequent authorisations".  You are not testing this as you have not made subsequent authorisations.  Please make a subsequent auth without any details to see that the details are actually used.

mattheworiordan: Note that call authorise will not actually do anything if a token exists, so you need the first token to expire so that a new one has to be issued and you need to assert that the new one is not the same as the old in your test.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -324,6 +324,9 @@ public JsonObject asJSON() {
 	 * @param callback (err, tokenDetails)
 	 */
 	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
+		/* Spec: RSA10g */
+		storeAuthOptions(options);
```
paddybyers: I'd prefer if we have something like:

.... in `AuthOptions` :

```
AuthOptions storedValues() {
  AuthOptions result = new AuthOptions();
  result.key = this.key;
  result.authUrl = authUrl;
  ...
}
```

and then here:

```
  this.authOptions = options.storedValues();
  this.tokenParams = params.storedValues();
```

Then at least we don't need to leak too much dependency between those types and this class.

lib/src/main/java/io/ably/lib/types/ClientOptions.java
```diff
@@ -7,6 +7,8 @@
 
 import java.util.Map;
 
+import static io.ably.lib.rest.Auth.*;
```
paddybyers: I do prefer individual explicit imports rather than wildcards.

.travis.yml
```diff
@@ -3,12 +3,18 @@ jdk:
   - oraclejdk8
   - openjdk7
```
paddybyers: I'm not sure why this is part of this PR. If we're going to do this change, then preferably it should be a different PR so we don't create unnecessary dependencies between PRs.

paddybyers:
I added a few comments but apart from that LGTM.

.travis.yml
dsvishchov: Are you talking about lines above or below? Lines above were not changed. Lines below required to make tests run properly, but agree this could be done as a separate PR. Also it's also already properly handled in the update project structure branch along with the rest of required changes to Travis script.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -0,0 +1,95 @@
...
+                    @Override
+                    public Object getTokenRequest(TokenParams params) throws AblyException {
+                        return ably.auth.requestToken(params, null);
+                    }
+                };
+                key = testVars.keys[1].keyStr;
+            }};
+
+            /* authorise with custom options */
+			TokenDetails tokenDetails1 = ably.auth.authorise(authOptions, tokenParams);
```
mattheworiordan: Weird spacing here?

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -0,0 +1,95 @@
...
+			TokenDetails tokenDetails2 = ably.auth.authorise(null, null);
+
+			/* Verify that,
+			 * tokenDetails1 and tokenDetails2 aren't null,
+			 * new token has to be issued,
+			 * capability and clientId for different TokenDetails are the same */
+			assertNotNull(tokenDetails1);
+			assertNotNull(tokenDetails2);
+			assertNotEquals(tokenDetails1.token, tokenDetails2.token);
+			assertEquals(tokenDetails1.capability, tokenDetails2.capability);
```
mattheworiordan: I appreciate this probably works, but if tokenDetails1.capability was invalid then the second test here would be pointless as at no point are you checking that the capability matches `capability`.  Same applies to clientID, you never check that it is ever `firstClientId` so it could feasibly be `defaultClientId` and this test would pass.  Please update your tests to assert the expectations after the first token is issued.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
VOstopolets: Sorry, it's old test, and I haven't changed these spaces. I changed my IDE settings, and in the future I hope it will not happen.

VOstopolets: I am going to fix this in the next commit

VOstopolets:
Thanks for comments. I have improved

VOstopolets:

.travis.yml
mattheworiordan: He was referring to the lines below.  It should be a separate PR really as it's unrelated to this PR

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -0,0 +1,103 @@
...
+		Setup.TestVars testVars = Setup.getTestVars();
+		ClientOptions opts = testVars.createOptions(testVars.keys[0].keyStr);
+		opts.clientId = "defaultClientId";
+		ably = new AblyRest(opts);
+	}
+
+	/**
+	 * Stores the AuthOptions and TokenParams arguments as defaults
+	 * for subsequent authorisations with the exception of the attributes:
+	 * {@link TokenParams#timestamp}, {@link AuthOptions#queryTime}, {@link AuthOptions#force}
```
mattheworiordan: How are you asserting that `timestamp`,`queryTime` and `force` are not used as defaults for the next request?

VOstopolets: Oh, You are right. I'll improve and check these values.

VOstopolets:

trenouf:
Just done a push where I squashed all the commits, removed the irrelevant travis build one, and did a minor mod or two.

trenouf:
I haven't actually tested it yet, so that push was a bit previous.

trenouf:
Pushed a rebase and a fix to the test class so each test works standalone and does not interact with other tests.

This is ready to go now.

trenouf:
Perhaps it's best not to merge this yet. PR https://github.com/ably/ably-java/pull/156 seems to have changes that implement this in a slightly different way, as well as the changes that its title advertises it as having. I need to understand what the interaction is.

trenouf:
Superseded by PR https://github.com/ably/ably-java/pull/182

PR #156, opened 2016-09-05
RSA8e, RSA9h, RSA10j

VOstopolets:


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -128,16 +128,53 @@ public AuthOptions(String key) throws AblyException {
...
-			if(authCallback == null) authCallback = defaults.authCallback;
-			if(authUrl == null) authUrl = defaults.authUrl;
-			if(key == null) key = defaults.key;
-			if(authHeaders == null) authHeaders = defaults.authHeaders;
-			if(authParams == null) authParams = defaults.authParams;
-			queryTime = queryTime & defaults.queryTime;
-			return this;
+		public AuthOptions replace(AuthOptions options) {
+			AuthOptions copy;
+			if (this instanceof ClientOptions) {
```
mattheworiordan: Sorry, not sure I understand this logic, can you explain why this would be a `ClientOptions`?

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -234,6 +271,48 @@ public static TokenDetails fromJSON(JsonObject json) {
...
 			return params;
 		}
+
+		/**
+		 * Internal, when TokenParams are provided, the values of each attribute are not merged,
+		 * but instead replace all corresponding values, even when null.
+		 *
+		 * should be used for:
+		 * - {@link Auth#requestToken(TokenParams, AuthOptions)} (Spec:RSA8e),
+		 * - {@link Auth#createTokenRequest(AuthOptions, TokenParams)} (Spec:RSA9h),
```
mattheworiordan: The method signature is wrong, it's `TokenParams, AuthOptions` for all methods.  It's in the spec you're referencing.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -324,9 +403,15 @@ public JsonObject asJSON() {
 	 * @param callback (err, tokenDetails)
 	 */
 	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
+		/* Spec: RSA10j */
```
mattheworiordan: See RSA10j, you have an incorrect method signature

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -460,8 +541,13 @@ public TokenDetails handleResponse(int statusCode, String contentType, Collectio
 	 * @throws AblyException
 	 */
 	public TokenRequest createTokenRequest(AuthOptions options, TokenParams params) throws AblyException {
-		if(options == null) options = this.authOptions;
-		else options.merge(this.authOptions);
+		/* Spec: RSA9h */
```
mattheworiordan: See RSA9h, method signature is wrong

lib/src/main/java/io/ably/lib/types/ClientOptions.java
```diff
@@ -31,6 +31,47 @@ public ClientOptions(String key) throws AblyException {
...
+		return ClientOptions.copyAttributes(copy, options);
+	}
+
+	/**
+	 * Copy the values of each attribute
+	 * @param src source object for copy attributes
+	 * @param dest destination object
+	 * @return source object with new attributes
+	 */
+	public static ClientOptions copyAttributes(ClientOptions src, ClientOptions dest) {
```
mattheworiordan: What is this needed for and how does this relate to `RSA8e, RSA9h, RSA10j`?   I am not following how `ClientOptions` would ever be passed in again to the client library as you can only instance it once.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -24,86 +25,268 @@
...
+			 * storedAuthOptions isn't null,
+			 * and storedAuthOptions equals custom authOptions */
+			assertNotNull(storedAuthOptions);
+			assertEquals(storedAuthOptions.authCallback, authOptions.authCallback);
+			assertEquals(storedAuthOptions.authUrl, authOptions.authUrl);
+			assertEquals(storedAuthOptions.token, authOptions.token);
+			assertEquals(storedAuthOptions.key, authOptions.key);
+			assertEquals(storedAuthOptions.tokenDetails, authOptions.tokenDetails);
+			assertArrayEquals(storedAuthOptions.authHeaders, authOptions.authHeaders);
+			assertArrayEquals(storedAuthOptions.authParams, authOptions.authParams);
```
mattheworiordan: You are not testing behaviour you are testing an undocumented method `getAuthOptions` and `getTokenParams`.  This is not how we would like our tests structured.  Where possible, we always test the public API and it's behaviour.  If the implementation of the library changes, these tests will be broken, yet the underlying behaviour may not have changed.  Test the public API unless there is no other way.  In this case, there was.  Please bear that in mind for all tests.

lib/src/main/java/io/ably/lib/rest/Auth.java
VOstopolets: ClientOptions class extends from AuthOptions, and we can pass ClientOptions to `requestToken`, `createTokenRequest`, `authorise` method as argument. Is this argument AuthOptions instance for these methods?

lib/src/main/java/io/ably/lib/rest/Auth.java
VOstopolets: Do I create a new method with correct signature or just change exists method?

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: No you cannot pass `ClientOptions` to these methods.  ClientOptions is a superset of AuthOptions so should not be allowed to be passed in at all. Why did you think that ClientOptions is valid here?

mattheworiordan: No you cannot pass `ClientOptions` to these methods.  ClientOptions is a superset of AuthOptions so should not be allowed to be passed in at all. Why did you think that ClientOptions is valid here?

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: We need to fix the existing method please, this was a mistake from some time back and should be fixed so we have a consistent API please

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
VOstopolets: I understood that test `auth_stores_options_params` for `RSA10g` is incorrect. For this PR, I have created other tests `auth_custom_options_request_token` (`RSA8e`), `auth_custom_options_create_token_request`(`RSA9h`), auth_custom_options_authorise (`RSA10j`), what do you think about these tests?

lib/src/main/java/io/ably/lib/rest/Auth.java
dsvishchov: Well, this is a bit more complicated... Question is: why do we extend ClientOptions from AuthOptions? And not including AuthOptions within ClientOptions? I.e. not using composition instead of inheritance? 

I've tried to find some examples of authorise in the docs but no luck so far. Trying to get a vision from the end user perspective. Technically user can pass ClientOptions to those methods as well. According to the docs let's say there is a reference like "token or tokenDetails attributes of ClientOptions" (http://docs.ably.io/rest/authentication/) which make me feels like I can definitely use it as AuthOptions. Also I might think that once I authorise using AuthOptions I might found my AuthOptions within AblyRest.options.

Just confusing a bit. If user can't pass ClientOptions to those methods then I think we should at least raise an exception telling that it's not a proper way of using it.

So all in all we just tried to cover different cases possible from the technical standpoint. Not doing checks for ClientOptions could be easier for sure.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
mattheworiordan: I have only commented on the things that I noticed appear to be wrong.  I believe the others are OK but we do still need @paddybyers to review.

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: The implementation choice to extend ClientOptions with AuthOptions is fine, perhaps composition would have been better (we use that primarily with Ruby), but I don't think that in itself is a problem.

The problem is that you're doing checks for a ClientOptions argument which we do not support in our API.  The API docs and IDL is all that we care about, and you have added code to accept ClientOptions and work with that specifically when that is not supported.

So if ClientOptions adheres to the AuthOptions spec, and that is passed in to the `authorise` method, then that is up to the developer even though it would be weird.  But we should then only be treating it as a AuthOptions object and by specifying the type as AuthOptions for this argument, we don't care and should not care if it's FooOptions but inherits from AuthOptions as we will only use the interface defined by AuthOptions.

Following the current logic in this PR, we'd need to support potentially any other objects that may inherit from AuthOptions, but that is never necessary and entire point of having interfaces.

So please stick to the spec, the types defined, and only handle those types based on the interfaces specified in the types.  That way the compiler will ensure the types you work with match the requirements of your code.

dsvishchov: Got it. Thanks for the detailed answer.

VOstopolets:

trenouf:
Superseded by PR https://github.com/ably/ably-java/pull/182

PR #157, opened 2016-09-05
RSA10d

VOstopolets:


lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -289,4 +290,80 @@ public Object getTokenRequest(TokenParams params) throws AblyException {
...
+
+			/* authorise with default options */
+			TokenDetails tokenDetails2 = ably.auth.authorise(authOptions, null);
+
+			/* Verify that,
+			 * tokenDetails1 and tokenDetails2 aren't null,
+			 * tokens are different */
+			assertNotNull(tokenDetails1);
+			assertNotNull(tokenDetails2);
+			assertNotEquals(tokenDetails1.token, tokenDetails2.token);
```
mattheworiordan: I don't think you've asserted this correctly. You should set an attribute on the token request that is unique such as a `clientId` and then check again that the second token issued has the same attributes.  Otherwise "the previously configured authentication options will remain intact" is not being asserted which is the purpose of this test.

VOstopolets:

trenouf:
RSA10d is no longer in the 0.9 spec.

mattheworiordan:
> RSA10d is no longer in the 0.9 spec.

Not sure why that matters?  We are still on 0.8 AFAICT?

trenouf:
Sorry, the 0.9 note was just a note to self from when I was talking to Paddy the other day.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -176,6 +186,24 @@ public static AuthOptions copyAttributes(AuthOptions src, AuthOptions dest) {
...
+
+		/**
+		 * A special case convenience for AuthOption
+		 * that if all its attributes are null apart from force,
+		 * then when passed to authorise as an argument
+		 * the previously configured authentication options will remain intact
+		 * This behaviour takes precedence over RSA10j
+		 * <p>
+		 * Spec: RSA10d
+		 * </p>
```
trenouf: @paddy Why are we doing this? I can't see anything in the spec that says to do this, especially not RSA10d.

trenouf:
I think we want just the test from this PR, and I have put the test into PR https://github.com/ably/ably-java/pull/182, so that PR supersedes this one.

PR #158, opened 2016-09-06
0.9 RSA10k persist server-local time offset

VOstopolets:


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -730,6 +735,18 @@ public void onAuthError(ErrorInfo err) {
 
 	public static long timestamp() { return System.currentTimeMillis(); }
 
+	/**
+	 * A client may discard the cached local clock offset in situations
+	 * in which it may have been invalidated, such as if there is a local change
+	 * to the date, time, or timezone, of the client device
+	 * <p>
+	 * Spec: RSA10k
```
mattheworiordan: This means the realtime client, not the customer, hence there is no documented API for this.  This is wrong.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -617,10 +617,15 @@ else if(!request.keyName.equals(keyName))
...
-				request.timestamp = ably.time();
-			else
+			if (options.queryTime) {
+				/* Spec: RSA10k */
+				if (timeOffset == null) {
+					timeOffset = ably.time() - timestamp();
+				}
+				request.timestamp = timeOffset + timestamp();
+			} else {
 				request.timestamp = timestamp();
```
mattheworiordan: This appears to be broken. You are not using the offset for subsequent requests so there is no point querying the time as it will not be used for any future requests when the local clock is out of sync

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -366,4 +366,67 @@ public void auth_authorise_only_force() {
...
+			/* Verify that,
+			 * new token isn't null, tokens aren't equals and
+			 * method time() not invoked again */
+			assertNotNull("Expected token value", tokenDetails2.token);
+			assertNotEquals(tokenDetails1.token, tokenDetails2.token);
+			assertEquals("Excepted one invoke", counter, 1);
+
+			/* discard the cached local clock offset and
+			 * re-authorise with same custom auth options */
+			ablyRest.auth.discardTimeOffset();
```
mattheworiordan: As above, undocumented API.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
```diff
@@ -366,4 +366,67 @@ public void auth_authorise_only_force() {
...
+			TokenDetails tokenDetails = ablyRest.auth.authorise(authOptions, null);
+
+			/* Verify that, issued token isn't null and method time() invoked twice */
+			assertNotNull("Expected token value", tokenDetails.token);
+			assertEquals("Excepted one invoke", counter, 2);
```
mattheworiordan: You are not testing that the offset is being used at ny point, you are simply testing that the server time is being loaded.  As you can see from the comment at the top, you are not using the offset so this test is broken.  You will need to consider a way to test this effectively, perhaps inject an intentional offset of 10s and check that the token request has a timestamp roughly 10s out.  This test is not doing anything unfortunately because you are not testing the intended behaviour but instead testing the innards of the client library.

lib/src/main/java/io/ably/lib/rest/Auth.java
dsvishchov: So for now we just need this method implemented but no such "situations" handled? Or do we need to handle things like local change of the date, time or timezone?

lib/src/main/java/io/ably/lib/rest/Auth.java
dsvishchov: So the flow should look like this:
1) customer initiates a single request with queryTime = 1 to get synced with server time;
2) all subsequent requests should happen with queryTime = 0 but use timeOffset calculated in the initial request?
3) if new request with queryTime = 1 initiated we need to get synced with server time once again?

Right behavior?

Current implementation was done with assumption that all requests which should rely on server time and offset should have queryTime = 1 and use discardTimeOffset when required to query time once again. Look like it was not a correct assumption. Please confirm.

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
dsvishchov: This is clear (once answered on the first comment above)

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
dsvishchov: This is clear (once answered on the second comment above)

VOstopolets: @mattheworiordan So, for testing by intentional offset, I should make persisted `timeOffset` field public for changing it in a test. But there is no documented API for this. I can get to access private field via reflection. It'll be correct?

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: > So for now we just need this method implemented but no such "situations" handled? Or do we need to handle things like local change of the date, time or timezone?

Well the spec item was just leaving that option open but nots stating that this was actually needed, it said the "Client may...".

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: > 3) if new request with queryTime = 1 initiated we need to get synced with server time once again?

Yes, although to be clear that's not obvious in the spec whether it should query again or not if queryTime is explicitly set.  However, the point is that if queryTime is set true, then it gets the server time and offset, and then uses that offset for all time calculations moving forwards as in most conditions that will suffice.

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: No, we don't need to handle thinks like local change fortunately, we just left it open so that it could be implemented if we wanted without breaking the spec

dsvishchov: But if there is no option to do so then client "can't" not "may" :)

lib/src/test/java/io/ably/lib/test/rest/RestAuthAttributeTest.java
mattheworiordan: Yes, via reflection is right, or perhaps some way to hook into it, but it's definitely not a public exposed option if we can help it.

lib/src/main/java/io/ably/lib/rest/Auth.java
mattheworiordan: The "Client" is the "client library" not the client who is using the client library.  We should update that in the spec if that is unclear.

dsvishchov: Then it makes sense, right.

mattheworiordan: See https://github.com/ably/docs/commit/3dc3f5cf644c871ba3bd969edfa5cf669633c3d7

paddybyers:
> Yes, via reflection is right, or perhaps some way to hook into it, but it's definitely not a public exposed option if we can help it.

Preferably use Mockito in line with existing tests or suggest some other framework or systematic way instead of writing your own raw reflection code because this could easily become very messy.

trenouf:
RSA10k itself is not in the 0.8 spec, but other bits of spec kind of imply it, so I guess we want this in 0.8.

trenouf:
Changed my mind: RSA10k is the bit about only querying the server once and persisting the offset. Nothing implies that in 0.8, so this is not wanted in 0.8.

paddybyers:
@tpr did you look at this at all?

trenouf:
Will have another look later today.

paddybyers:
Superseded

PR #162, opened 2016-09-14
RSC11 environment prefix

VOstopolets:

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/rest/RestInitTest.java
+++ b/lib/src/test/java/io/ably/lib/test/rest/RestInitTest.java
@@ -235,4 +235,62 @@ public class RestInitTest {
 			System.setOut(oldTarget);
 		}
 	}
+
+	/**
+	 * Init library with 'production' environment
+	 * Spec: RSC11
+	 */
+	@Test
+	public void init_production_environment() {
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.environment = "production";
+			AblyRest ably = new AblyRest(opts);
+			assertEquals("Unexpected host mismatch", Defaults.HOST_REST, ably.options.restHost);
+		} catch (AblyException e) {
+			e.printStackTrace();
+			fail("init4: Unexpected exception instantiating library");
+		}
+	}
+
+	/**
+	 * Init library with given environment
+	 * Spec: RSC11
+	 */
+	@Test
+	public void init_given_environment() {
+		final String givenEnvironment = "staging";
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.environment = givenEnvironment;
+			AblyRest ably = new AblyRest(opts);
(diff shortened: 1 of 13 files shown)


mattheworiordan:
👍  looks very good to me!

paddybyers:
I think as part of this PR there should also be a change in `Setup.java` so that the `environment` option is passed in, instead of passing in `restHost` and `realtimeHost`.

VOstopolets:

VOstopolets:

lib/src/test/java/io/ably/lib/test/realtime/RealtimeInitTest.java
```diff
@@ -219,4 +219,62 @@ public void println(int severity, String tag, String msg, Throwable tr) {
...
+	 */
+	@Test
+	public void init_given_host_environment() {
+		final String givenEnvironment = "staging";
+		final String specifiedHost = "fake.ably.io";
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.realtimeHost = specifiedHost;
+			opts.environment = givenEnvironment;
```
mattheworiordan: This is not allowed and should simply raise an exception or fail in some way.  See http://docs.ably.io/client-lib-development-guide/versions/features-0-8/#TO3k3

VOstopolets: So, I can't use `realtimeHost` and `environment` together for init `AblyRealtime`. 
If I want to use "env-test.realtime.host" as `realtimeHost`, I should set this host like 
`options.realtimeHost="env-test.realtime.host"`
instead 
`options.environment = "env"` 
`options.realtimeHost = "test.realtime.host"`.
`environment` name can be used only for default host endpoint.
And the same logic for `AblyRest`. Right?

mattheworiordan: Yes, that is why it states it is never valid to specify both because it is unnecessary i.e. either a client will have a specific host they want to connect to, or they will use our built in environment naming convention, but never both

VOstopolets: I'll change it, and raise an exception, when user will create ably with both values. Thanks.

VOstopolets:

trenouf:
This is ready to go now with that push.

paddybyers:
Thanks

lib/src/test/java/io/ably/lib/test/rest/RestInitTest.java
```diff
@@ -235,4 +235,62 @@ public void init_default_log_output_stream() {
...
+	public void init_given_host_environment() {
+		final String givenEnvironment = "staging";
+		final String specifiedHost = "fake.ably.io";
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.restHost = specifiedHost;
+			opts.environment = givenEnvironment;
+			AblyRest ably = new AblyRest(opts);
+			assertEquals("Unexpected host mismatch", specifiedHost, ably.options.restHost);
```
mattheworiordan: This should never be allowed and should raise an exception, not allow this, http://docs.ably.io/client-lib-development-guide/versions/features-0-8/#TO3k2

PR #163, opened 2016-09-16
RSC15b, TO3k6

VOstopolets:

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -10,7 +10,9 @@ import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -325,7 +327,17 @@ public class Http {
 	 * @throws AblyException
 	 */
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
-		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
+		List<String> customFallbackHosts = null;
+		int retryCountRemaining = 0;
+		if (options.fallbackHosts != null) {
+			if (options.fallbackHosts.length > 0) {
+				retryCountRemaining = options.httpMaxRetryCount;
+				customFallbackHosts = Arrays.asList(options.fallbackHosts);
+				Collections.shuffle(customFallbackHosts);
+			}
+		} else if (Hosts.isRestFallbackSupported(this.host)) {
+			retryCountRemaining = options.httpMaxRetryCount;
+		}
 		String candidateHost = this.host;
 		URL url;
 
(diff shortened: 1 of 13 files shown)


lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -325,7 +327,17 @@ public void finalize() {
 	 * @throws AblyException
 	 */
 	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
-		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
+		List<String> customFallbackHosts = null;
+		int retryCountRemaining = 0;
+		if (options.fallbackHosts != null) {
+			if (options.fallbackHosts.length > 0) {
```
mattheworiordan: Why if > 0, shouldn't we set the fallback hosts to an empty array if provided i.e. so that the user can disable fallback if they want?

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -70,6 +72,152 @@ public static void tearDown() {
...
+		ClientOptions options = new ClientOptions();
+		options.fallbackHosts = expectedFallbackHosts;
+		AblyRest ably = new AblyRest(options);
+
+		Http http = Mockito.spy(new Http(ably.options, ably.auth));
+
+		String responseExpected = "Lorem Ipsum";
+		ArgumentCaptor<URL> url = ArgumentCaptor.forClass(URL.class);
+
+        /* Partially mock http */
```
mattheworiordan: Indentation

lib/src/test/java/io/ably/lib/http/HttpTest.java
```diff
@@ -70,6 +72,152 @@ public static void tearDown() {
...
+		ClientOptions options = new ClientOptions();
+		options.fallbackHosts = null;
+		AblyRest ably = new AblyRest(options);
+
+		Http http = Mockito.spy(new Http(ably.options, ably.auth));
+
+		String responseExpected = "Lorem Ipsum";
+		ArgumentCaptor<URL> url = ArgumentCaptor.forClass(URL.class);
+
+        /* Partially mock http */
```
mattheworiordan: Indentation

lib/src/main/java/io/ably/lib/http/Http.java
VOstopolets: If ClientOptions#fallbackHosts array isn't empty, we should use only these fallback hosts. So, if user want to disable fallback functionality, user should provide specific restHost (it isn't default primary host) or empty array of ClientOptions#fallbackHosts.

VOstopolets: It's wrong?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -325,7 +327,17 @@ public void finalize() {
...
+				retryCountRemaining = options.httpMaxRetryCount;
+				customFallbackHosts = Arrays.asList(options.fallbackHosts);
```
paddybyers: We can't do this work (create an array and shuffle it) on each HTTP request, even when fallback hosts are not needed

VOstopolets:

lib/src/main/java/io/ably/lib/http/Http.java
mattheworiordan: Unless I am reading this incorrectly, having `options.fallbackHosts` with an empty array is now going to use the defaults instead of disabling fallback host functionality?  Have I misread it?

VOstopolets:
@mattheworiordan If `options.fallbackHosts` array is empty, `retryCountRemaining` won't be changed, and `retryCountRemaining` value still is 0. So, when we catch HostFailedException, `(--retryCountRemaining < 0)` is `true`, and fallback host functionality don't used.

mattheworiordan:
> If options.fallbackHosts array is empty, retryCountRemaining won't be changed, and retryCountRemaining value still is 0. So, when we catch HostFailedException, (--retryCountRemaining < 0) is true, and fallback host functionality don't used.

Thanks for the explanation

trenouf:
I think this is all now covered by other PRs for environment and for fallback shuffling.

PR #165, opened 2016-09-16
RSC15a

VOstopolets:


trenouf:
The only changes in here not already covered by other PRs are some new tests for rest host fallback, which I have now put in a new PR https://github.com/ably/ably-java/pull/184. So this one is superseded.

PR #173, opened 2016-10-06
heartbeat support

trenouf:


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -109,6 +109,8 @@ public StateIndication(ConnectionState state, ErrorInfo reason, boolean fallback
 		put(ConnectionState.failed, new StateInfo(ConnectionState.failed, false, false, true, false, 0, REASON_FAILED));
 	}};
 
+	public long maxIdleInterval;
```
paddybyers: public?

lib/src/main/java/io/ably/lib/transport/ITransport.java
```diff
@@ -39,6 +39,11 @@
 		String connectionKey;
 		String connectionSerial;
 		Mode mode;
+		boolean heartbeats;
+
+		public TransportParams() {
+			heartbeats = true; // default to requiring Ably heartbeats
```
paddybyers: We're using `/* .... */ ` for all comments please.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
trenouf: Done

lib/src/main/java/io/ably/lib/transport/ITransport.java
trenouf: Done

mattheworiordan:
Thanks, looks good, would have been nice to have seen a test to ensure we don't introduce a regression down the line which could be pretty nasty given none of our tests would probably detect it now, and in production connections would just be terminated.

mattheworiordan:
@trenouf @paddybyers I think we should add a test to cover this. Happy to add an issue that we look at later though.  Please advise

trenouf:
So @mattheworiordan are you thinking a test that does nothing for 2 mins and checks that the connection stays?

mattheworiordan:
> So @mattheworiordan are you thinking a test that does nothing for 2 mins and checks that the connection stays?

Nope, I would expect that we could change the heartbeat period by perhaps creating a simple WS server that responds with a Connected message that indicates heartbeats are expected say every 1s. You could then a) not send them to see if it fails when expected, b) send them at that rate for say 15 seconds and ensure we've not disconnected the connection.  It will require a bit of jiggery, but where necessary we do this in all libraries as a regression in the future of this kind would be quite devastating.

mattheworiordan:
@paddybyers @trenouf I've just realised we've merged this and tagged a release, yet as far as I am aware, we don't yet support native WebSocket heartbeats on realtime i.e. this is not yet released as it's part of 0.9.  Does that not mean this change would result in Java clients thinking that the connection has gone away, or does this change ensure that both native heartbeats and Ably protocol heartbeats will work as a means to detect liveness?

paddybyers:
> as far as I am aware, we don't yet support native WebSocket heartbeats on realtime

Realtime has been sending ws ping frames for a long time.

The 0.9 feature is to request that these are the only pings, ie protocol heartbeats are suppressed.

mattheworiordan:
> Realtime has been sending ws ping frames for a long time.

Ok, good, thanks

PR #175, opened 2016-10-06
issues/170: Fixed message serial out of sync after recover

trenouf:
Also protocol message logging.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -15,6 +15,7 @@ import io.ably.lib.types.ClientOptions;
 import io.ably.lib.types.ErrorInfo;
 import io.ably.lib.types.ProtocolMessage;
 import io.ably.lib.types.ProtocolMessage.Action;
+import io.ably.lib.types.ProtocolSerializer;
 import io.ably.lib.util.Log;
 
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


trenouf:
This change might still be wrong. Paddy and I are discussing.

trenouf:
What do you reckon to that Paddy?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -271,6 +272,7 @@ public void run() {
 	 ***************************************/
 
 	public void onMessage(ProtocolMessage message) throws AblyException {
+		Log.v(TAG, "onMessage(): " + new String(ProtocolSerializer.writeJSON(message)));
```
paddybyers: I don't like the way that this serialises every message even though the log level doesn't require it to be emitted. I think it has to be done conditionally on the log level, or the log API has to take a format string and arguments so expansion is under control of the logger.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -860,6 +860,20 @@ public synchronized void nack(long serial, int count, ErrorInfo reason) {
...
 		}
+
+		/**
+		 * reset the pending message queue, failing any currently pending messages.
+		 * Used when a resume fails and we get a different connection id.
+		 * @param oldMsgSerial the next message serial number for the old
+		 *		connection, and thus one more than the highest message serial
+		 *		in the queue.
+		 */
+		public synchronized void reset(long oldMsgSerial) {
```
paddybyers: I think `reset()` should take an `ErrorInfo` argument and pass that to the nack, and then the specific error is determined by whoever calls `reset()`. There will be different situations that will require different error messages/codes.

trenouf:
@paddy fixed your comments.

paddybyers:
Thanks

PR #176, opened 2016-10-10
custom fallback support

trenouf:
This is a tidy-up of what is in
https://github.com/ably/ably-java/pull/141
based on the commits from Piotr.

Don't merge it yet, as I am currently unable to run tests due to sandbox being down. But, @mattheworiordan, could you comment please on whether this addresses your comments in the earlier PR?

Also moved the http tests into rest suite (as done by Paddy over on project_struct_update) and moved the transport tests into realtime suite (as not done elsewhere as far as I know).

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
@@ -155,11 +155,15 @@ public class AsyncHttp extends ThreadPoolExecutor {
 					break;
 				} catch (AblyException.HostFailedException e) {
 					if(--retryCountRemaining < 0) {
-						setError(new ErrorInfo("Connection failed; no host available", 404, 80000));
+						setError(e.errorInfo);
 						break;
 					}
 					Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
 					candidateHost = Hosts.getFallback(candidateHost);
+					if (candidateHost == null) {
+						setError(e.errorInfo);
+						break;
+					}
 					Log.d(TAG, "Switched to `" + candidateHost + "`.");
 				} catch(AblyException e) {
 					setError(e.errorInfo);
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
@@ -145,8 +145,8 @@ public class AsyncHttp extends ThreadPoolExecutor {
 		}
 		@Override
 		public void run() {
-			int retryCountRemaining = Hosts.isRestFallbackSupported(http.host) ? http.options.httpMaxRetryCount : 0;
-			String candidateHost = http.host;
+			String candidateHost = http.getHost();
+			int retryCountRemaining = http.hosts.getFallback(candidateHost) != null ? http.options.httpMaxRetryCount : 0;
 
 			while(!isCancelled) {
 				try {
@@ -159,7 +159,7 @@ public class AsyncHttp extends ThreadPoolExecutor {
 						break;
 					}
 					Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
-					candidateHost = Hosts.getFallback(candidateHost);
+					candidateHost = http.hosts.getFallback(candidateHost);
 					if (candidateHost == null) {
 						setError(e.errorInfo);
 						break;
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -151,9 +151,9 @@ public class Http {
 	public Http(ClientOptions options, Auth auth) throws AblyException {
 		this.options = options;
 		this.auth = auth;
-		this.host = options.restHost;
 		this.scheme = options.tls ? "https://" : "http://";
 		this.port = Defaults.getPort(options);
+		this.hosts = new Hosts(options.restHost, Defaults.HOST_REST);
 
 		this.proxyOptions = options.proxy;
 		if(proxyOptions != null) {
@@ -177,7 +177,7 @@ public class Http {
 	 * @param host URL string
 	 */
 	public void setHost(String host) {
-		this.host = host;
+		hosts.setHost(host);
 	}
 
 	/**
@@ -186,7 +186,7 @@ public class Http {
 	 * @return
      */
 	public String getHost() {
-		return host;
+		return hosts.getHost();
 	}
 
 	/**
@@ -325,8 +325,8 @@ public class Http {
 	 * @throws AblyException
 	 */
 	public <T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
-		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
-		String candidateHost = this.host;
+		String candidateHost = getHost();
+		int retryCountRemaining = hosts.getFallback(candidateHost) != null ? options.httpMaxRetryCount : 0;
 		URL url;
 
 		while(true) {
@@ -337,7 +337,7 @@ public class Http {
 				if(--retryCountRemaining < 0)
 					throw e; /* reached httpMaxRetryCount */
 				Log.d(TAG, "Connection failed to host `" + candidateHost + "`. Searching for new host...");
-				candidateHost = Hosts.getFallback(candidateHost);
+				candidateHost = hosts.getFallback(candidateHost);
 				if (candidateHost == null)
 					throw e; /* run out of fallback hosts */
 				Log.d(TAG, "Switched to `" + candidateHost + "`.");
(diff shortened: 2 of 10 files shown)


lib/src/main/java/io/ably/lib/http/AsyncHttp.java
```diff
@@ -147,6 +148,8 @@ private AblyRequestWithFallback(
 		public void run() {
 			int retryCountRemaining = Hosts.isRestFallbackSupported(http.host) ? http.options.httpMaxRetryCount : 0;
 			String candidateHost = http.host;
+			if(http.options.fallbackHosts != null)
+				Collections.shuffle(http.options.fallbackHosts);
```
mattheworiordan: Perhaps we should raise an issue and defer fixing this, but @paddybyers was concerned with shuffling unnecessarily for every HTTP request instead of just shuffling once at the outset the the custom fallbackhosts are provided.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -324,8 +325,10 @@ public void finalize() {
 	 * @return
 	 * @throws AblyException
 	 */
-	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
+	public <T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
+		if(options.fallbackHosts != null)
+			Collections.shuffle(options.fallbackHosts);
```
mattheworiordan: As above, we should really only need to shuffle this once and use that ordering for the rest of the lifetime of the instanced library

lib/src/test/java/io/ably/lib/test/realtime/HostsTest.java
```diff
@@ -0,0 +1,145 @@
...
+     * Expect multiple calls will provide different (relatively) valid fallback hosts
+     */
+    @Test
+    public void hosts_fallback_multiple() {
+        String host = Hosts.getFallback(null, Defaults.HOST_FALLBACKS);
+        assertThat(Hosts.getFallback(host, Defaults.HOST_FALLBACKS), is(not(allOf(isEmptyOrNullString(), equalTo(host)))));
+    }
+
+	/**
+	 * Expect a null, when we requested more than available fallback hosts
```
mattheworiordan: trivial but spacing

lib/src/test/java/io/ably/lib/test/rest/HttpTest.java
```diff
@@ -147,6 +163,118 @@ public Http setUrlArgumentStack(List<String> urlArgumentStack) {
 	}
 
 	/**
+	 * Validates that fallbacks are disabled when ClientOpitons#fallbackHosts are set to null and the only host used is Defaults#HOST_REST
```
mattheworiordan: This description is wrong. According to http://docs.ably.io/client-lib-development-guide/features/#TO3k6 fallback hosts are only disabled if an empty array is provided. Setting fallbackHosts to null, is effectively the same as not providing them, which AFAICT would break every client in that they will no longer us fallback hosts by default (unless ClientOptions sets that default of course, which I assume it does).  If `ClientOptions#fallbackHosts` default is populated, then perhaps this is OK to be honest. @paddybyers WDYT?

lib/src/test/java/io/ably/lib/test/rest/HttpTest.java
```diff
@@ -147,6 +163,118 @@ public Http setUrlArgumentStack(List<String> urlArgumentStack) {
...
+			ErrorInfo expectedErrorInfo = new ErrorInfo("Connection failed; no host available", 404, 80000);
+			assertThat(e, new ErrorInfoMatcher(expectedErrorInfo));
+		}
+
+		int expectedCallCount = options.httpMaxRetryCount + 1;
+		Assert.assertTrue(urlHostArgumentStack.size() == expectedCallCount);
+		assertThat(urlHostArgumentStack.get(0), is(equalTo(Defaults.HOST_REST)));
+
+		for (int i = 1; i < expectedCallCount; i++) {
+			Assert.assertTrue(urlHostArgumentStack.get(i).matches(PATTERN_HOST_FALLBACK));
```
mattheworiordan: Given this regex is `"(?i)[a-z]\\.ably-realtime.com"`, then if the default fallback hosts are used the this test would still pass?  Shouldn't this be testing that the fallback hosts are actually used?

lib/src/main/java/io/ably/lib/http/AsyncHttp.java
paddybyers: No, sorry, we have to change it. I thought I had seen an earlier update that did that.

trenouf: Piotr said:
I reverted my changes with ClientOptions#fallbackHosts back to the public field and shuffling it before every usage, that way we can be sure that every time someone use this field, it'll be random. If we randomize it only on the init, it can be overwritten later in the code since it's a public field and host order won't be random anymore. It also ensures that default hosts will also be randomized.

trenouf: So was Piotr wrong? Is there any situation where someone overwrites the public field and we need to ensure the new value gets randomised?

paddybyers:
> Is there any situation where someone overwrites the public field and we need to ensure the new value gets randomised?

The use case we handle is the fallback hosts being passed in in the options when initialising the library. If someone, using knowledge of the undocumented structure of the library, overwrites a field that happens to be public, then that's up to them and they get the resulting behaviour. We're not going to add extra work into the path of every request just to deal with that case.

mattheworiordan:
> We're not going to add extra work into the path of every request just to deal with that case.

👍 

Also, any reason we can't make that immutable if those options are not meant to modified by the user?  I don't think it's a big enough issue to worry about though if more than a trivial change.

paddybyers:
> Also, any reason we can't make that immutable if those options are not meant to modified by the user?

`defaultHosts` is public in `ClientOptions`, and the given `ClientOptions` instance is a public member of the library instance. So that means that anyone can access, modify or delete the `defaultHosts` that were originally passed in at any time. There's no way in java to make that member read-only.

It would be possible to make the shuffled copy of the `defaultHosts` and keep that in a separate private member but tbh I don't really see why it's that important. There are lots of other places in the library where the user can break it if they start modifying the contents of public members. For example, `auth` and `connection` are public members of `AblyRealtime` but we don't worry about anyone deleting the `auth` or `connection` member.

mattheworiordan:
>  For example, auth and connection are public members of AblyRealtime but we don't worry about anyone deleting the auth or connection member.

Urr, what about a public getter and no public setter for `auth` and `connection`.  Yes they coudl mutate the Auth and Connection objects, but that would only be possible based on the public interfaces available to those objects surely?

I don't think it's worth fixing, it's not important, but given Java provides a rich language feature set that allows you to control access to properties and methods, I find it odd that we didn't prevent mutation for interfaces that did not need to be public.  But that's a discussion for another day.

paddybyers:
> Urr, what about a public getter and no public setter for auth and connection.

Not possible in java.

mattheworiordan:
>> Urr, what about a public getter and no public setter for auth and connection.

> Not possible in java.

Oh, another reason why I don't  use Java then :)

paddybyers:
> > Urr, what about a public getter and no public setter for auth and connection.

> Not possible in java.

That's not quite true; if a member is initialised in the constructor it can be declared as `final` and is then read-only. This is in fact done for `auth` and `connection`.

trenouf:
Just done a push where I have refactored this fallback stuff. Note that only the last 4 commits belong in this PR; the other ones belong in https://github.com/ably/ably-java/pull/179 "issues/178: fix fallback". Once that PR merged, I'll rebase this PR.

trenouf:
I have also been able to run tests, so this is ready to go if approved.

lib/src/main/java/io/ably/lib/http/Http.java
trenouf: See above for shuffle changes.

lib/src/test/java/io/ably/lib/test/realtime/HostsTest.java
trenouf: Indents now tabs.

lib/src/test/java/io/ably/lib/test/rest/HttpTest.java
trenouf: Fixed by setting fallbackHosts to empty rather than null. The library now treats the two differently: empty means override with an empty fallback list, null means don't override.

lib/src/test/java/io/ably/lib/test/rest/HttpTest.java
trenouf: Fixed.

trenouf:
Rebased to get rid of common commits with PR 179.

Issue #178, opened 2016-10-11
fallback problems


trenouf:
In trying to move ConnectionManagerTest into realtime suite so it gets run, I find that connectionmanager_reconnect_default_endpoint hangs.

I suspect that fallback has rotted and doesn't work at all.

trenouf:
Fixed by PR https://github.com/ably/ably-java/pull/179, now merged.

PR #179, opened 2016-10-11
issues/178: fix fallback

trenouf:
Fixes
https://github.com/ably/ably-java/issues/178

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -324,7 +324,7 @@ public class Http {
 	 * @return
 	 * @throws AblyException
 	 */
-	<T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
+	public <T> T ablyHttpExecute(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
 		int retryCountRemaining = Hosts.isRestFallbackSupported(this.host) ? options.httpMaxRetryCount : 0;
 		String candidateHost = this.host;
 		URL url;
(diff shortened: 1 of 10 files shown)


trenouf:
I've got an intermittent failure in RealtimeConnectFailTest.connect_fail_suspended. Not sure if it was happening before my changes.

trenouf:
I think the failure is new.

trenouf:
The travis builds said:
io.ably.lib.test.realtime.RealtimeSuite > io.ably.lib.test.realtime.ConnectionManagerTest.connectionmanager_reconnect_default_endpoint FAILED
    java.lang.AssertionError at ConnectionManagerTest.java:209

paddybyers:
Reviewing now, but if the failure is only on Travis we will merge this anyway pending further investigation. Travis seems to have trouble with tests where there is a dependency on things happening at specific times.

trenouf:
I still need to track down the first test failure I mentioned: connect_fail_suspended.

paddybyers:
But any transition is first vetted by `checkSuspend()`, which decides, based on how long since the last connection existed, whether to go into `disconnected` or `suspended`. Doesn't that still apply?

trenouf:
Good point, I need to work out why that isn't happening on the second one. So I guess you don't think there is any need to suppress the extra notification from the transport.

trenouf:
It only calls checkSuspend if it was in connecting state. But, on my second event, it's on suspended. I'll add code in handleStateChange to stop it transitioning from suspended to disconnected.

paddybyers:
> I'll add code in handleStateChange to stop it transitioning from suspended to disconnected.

Instead of having explicit code to handle different cases, is it not possible to have it handled uniformly? If handling the disconnection (or failure to connect) of a transport, then check whether to go into disconnected or suspended?

trenouf:
You mean call checkSuspend even when already in disconnected? I fixed the fallback problems by working out what to fall back to in there, so I would need to separate that out again if checkSuspend can be called multiple times for a disconnection or failure to connect.

trenouf:
Anyway, I've done another push.

PR #182, opened 2016-10-17
0.8 authorisation changes

trenouf:
Supersedes PR https://github.com/ably/ably-java/pull/156 and PR https://github.com/ably/ably-java/pull/153

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -136,8 +136,47 @@ public class Auth {
 			if(key == null) key = defaults.key;
 			if(authHeaders == null) authHeaders = defaults.authHeaders;
 			if(authParams == null) authParams = defaults.authParams;
-			queryTime = queryTime & defaults.queryTime;
+			queryTime = queryTime | defaults.queryTime;
 			return this;
+		}
+
+		/**
+		 * Stores the AuthOptions arguments as defaults for subsequent authorisations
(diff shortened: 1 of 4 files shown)


trenouf:
I'm getting a new failure
io.ably.lib.test.rest.RestTokenTest.authtime2

trenouf:
After talking to Simon, it looks like we do not want either of the changes here while we're still on 0.8. So let's put this on hold.


tpr [5:28]  
While you’re here :slightly_smiling_face: I think I might be getting confused between 0.8 and 0.9 in the way the auth calls work. I have just sorted out and merged a change from vostopolets to implement the bits of the spec that say that supplying a non-null AuthOptions or TokenParams completely overrides the stored default, rather than merging. But is that meant to happen in 0.8? The java lib is still meant to be on 0.8. The 0.8 spec is not as explicit as the 0.9 spec on that point. Was the spec change a clarification, or a change in behaviour?

simon [5:30 PM]  
....to be honest, a bit of both :). Before the 0.9 spec, different libraries did different things for this. But in general, 0.8 versions of libraries should keep doing whatever they were doing before -- there shouldn't be any potentially breaking API changes happening within patch versions of a minor version. So if the java lib wasn't doing that before, it should only change on the 0.9 branch

trenouf:
I have done a new push that does not contain the compatibility-breaking changes that should be delayed to 0.9.

It now has:
RSA10g Auth.authorise stores TokenParams and AuthOptions
Test for RSA10d AuthOptions.force forces issue of new token

paddybyers:
Thanks

PR #184, opened 2016-10-18
Tests for host fallback behaviour on rest

trenouf:
Tests for host fallback behaviour on rest. Supersedes PR https://github.com/ably/ably-java/pull/165


paddybyers:
Thanks

PR #195, opened 2016-10-27
RSC7a: X-Ably-Version header

trenouf:


paddybyers:
LGTM

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -15,6 +15,9 @@
  */
 public class HttpUtils {
 	/* Headers */
+	public static final String X_ABLY_VERSION_HEADER = "X-Ably-Version";
+	public static final String X_ABLY_VERSION_VALUE = BuildConfig.VERSION.substring(0,
+          BuildConfig.VERSION.indexOf('.', BuildConfig.VERSION.indexOf('.') + 1));
```
mattheworiordan: I dislike this.  If at any point you wanted to have a client library version diverge from the Ably protocol version this will no longer work.  Adding a constant is straightforward and gives us flexibility. See https://github.com/ably/ably-ruby/blob/master/lib/ably/version.rb for a similar example.

trenouf:
OK, will fix in a new PR as this one is already merged.

As a matter of protocol, should I wait for both of you @paddy and @mattheworiordan to review before I merge?

mattheworiordan:
> As a matter of protocol, should I wait for both of you @paddy and @mattheworiordan to review before I merge?

Well TBH I am supposed to be offline this week, so I wouldn't wait too long for me, but a day leeway would be appreciated.

PR #196, opened 2016-10-27
useTokenAuth forces token authorization

trenouf:
TO3j4, RSA4: Setting useTokenAuth true forces token authorization, even
if no clientId is provided.

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/rest/RestAuthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/rest/RestAuthTest.java
@@ -79,6 +79,23 @@ public class RestAuthTest {
 	}
 
 	/**
+	 * Init library with useTokenAuth set
+	 */
+	@Test
+	public void authinit0_useTokenAuth() {
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.useTokenAuth = true;
+			AblyRest ably = new AblyRest(opts);
+			assertEquals("Unexpected Auth method mismatch", ably.auth.getAuthMethod(), AuthMethod.token);
+		} catch (AblyException e) {
+			e.printStackTrace();
+			fail("authinit0point5: Unexpected exception instantiating library");
+		}
+	}
+
+	/**
 	 * Init library with a token only
 	 */
 	@Test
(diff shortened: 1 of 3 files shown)


paddybyers:
Can we have a test that verifies that token auth does in fact get used if it is set?

trenouf:
OK. How would I tell in a test?

paddybyers:
> How would I tell in a test?

There are tests already I believe that add an `authCallback` and verify that the callback is called.

trenouf:
Right, but the point of useTokenAuth is that it forces token auth even if you don't have an authCallback or any other means of renewing the token.

paddybyers:
> Right, but the point of useTokenAuth is that it forces token auth even if you don't have an authCallback or any other means of renewing the token.

Yes, of course. Can you connect and then inspect any of the `Auth` attributes and confirm that the `method` is token, or `tokenDetails` is populated?

Failing that you could use the Http Mock stuff to verify that an `accessToken` is passed to rest requests but I would have thought that that would be avoidable.

lib/src/test/java/io/ably/lib/test/rest/RestAuthTest.java
```diff
@@ -79,6 +79,23 @@ public void authinit0() {
...
+	 * Init library with useTokenAuth set
+	 */
+	@Test
+	public void authinit0point5() {
+		try {
+			TestVars testVars = Setup.getTestVars();
+			ClientOptions opts = new ClientOptions(testVars.keys[0].keyStr);
+			opts.useTokenAuth = true;
+			AblyRest ably = new AblyRest(opts);
+			assertEquals("Unexpected Auth method mismatch", ably.auth.getAuthMethod(), AuthMethod.token);
```
mattheworiordan: Why not simply check if a token exists after a REST operation i.e. inspect the Auth object to ensure it has not token, perform a REST operation, inspect the Auth object to see if it now has a token.  Should be pretty straightforward.

trenouf: Good plan. I've just done a realtime test for it, so I think I'll do what you suggest and have both.

trenouf:
New push has improved tests. Please review.

paddybyers:
LGTM

PR #199, opened 2016-10-27
Add note on proguard to readme

SimonWoolf:


mattheworiordan:
LGTM.  Would be good to add to our FAQs in the support site too in case people search there.

SimonWoolf:
https://support.ably.io/solution/articles/3000059661-classnotfoundexception-when-using-proguard-with-ably-java

paddybyers:
Shouldn't we be doing something like this?

http://stackoverflow.com/questions/30201420/how-to-include-a-proguard-configuration-in-my-android-library-aar

paddybyers:
> Shouldn't we be doing something like this?
> 
> http://stackoverflow.com/questions/30201420/how-to-include-a-proguard-configuration-in-my-android-library-aar

@psolstice can you comment on this please?

psolstice:
Yes it's possible to do that (consumerProguardFiles option)
What you need to keep in mind that there is a new toolchain that does its own proguard minify/obfuscate processing
http://tools.android.com/tech-docs/jackandjill
and you need to test your lib against apps built both with classic and Jack toolchains. I don't recommend using Jack to build the lib right now (it's too new for that) but compatibility with Jack-built app must be ensured.

paddybyers:
@psolstice ok. Can you please do a PR (or update this one) pls to do the right thing for the the classic toolchain; and for the new Jack toolchain it can either work out of the box or we say in the README what to do?

Issue #202, opened 2016-10-28
intermittent failure: RealtimeCryptoTest.set_cipher_params


trenouf:
Paddy said:

That’s been an intermitted failure since the beginning of time. I looked at it once, and have no real idea why. If it happens 16th of the time it might be that the failed decryption is only detected that often, when the PKCS un-pad detects the error (which it will only do 1/16th of the time or something like that)

PR #203, opened 2016-10-29
Worked around RealtimeCryptoTest.set_cipher_params intermittent failure

trenouf:
See issue https://github.com/ably/ably-java/issues/202
This commit works around the intermittent failure in the last section of
the test by doing it up to 4 times until it succeeds.


trenouf:
Ready to go -- please review.

trenouf:
@paddy @mattheworiordan Please review. Thanks.

mattheworiordan:
LGTM, although I'd like @paddybyers view on whether there may be an underlying issue we need to address, or if this is more likely to be a test failure.

paddybyers:
TBH I really can't see how the original test can fail, and I'm uneasy just addressing it in this way. If there  is a race condition that means that you need to take account of multiple possible outcomes, then that would be ok - but in this case I just don't see one.

trenouf:
Yes, but this is just a workaround so we can get to 100% test pass before the actual issue is addressed. The issue remains open. An alternative workaround is to disable the test altogether.

paddybyers:
Thanks

PR #207, opened 2016-10-31
0.9 authorise->authorize, and fix param order

trenouf:
I think we haven't decided yet about whether to fix param order in 0.8, but in the meantime one of the two commits here fixes it in 0.9.


trenouf:
Oops, I created this PR against the wrong branch. Closing.

PR #208, opened 2016-10-31
0.9 authorise->authorize, fix param order

trenouf:
I think we haven't decided yet about whether to fix param order in 0.8, but in the meantime one of the two commits here fixes it in 0.9.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/TokenAuth.java
+++ b/lib/src/main/java/io/ably/lib/http/TokenAuth.java
@@ -36,7 +36,7 @@ public class TokenAuth {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
+	public TokenDetails authorise(TokenParams params, AuthOptions options) throws AblyException {
 		Log.i("TokenAuth.authorise()", "");
 		if(tokenDetails != null) {
 			if(tokenDetails.expires == 0 || tokenValid(tokenDetails)) {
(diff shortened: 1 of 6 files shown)


mattheworiordan:
I would like to see `(RSA10l) Has an alias method RestClient#authorise and RealtimeClient#authorise that will log a deprecation warning stating that this alias method will be removed in v1.0 and the user should instead use authorize` implemented with this if possible.  Should be straightforward.  The question then is do we use the old authorise signature or the new correct one, clearly I prefer using the new one to avoid all of this unnecessary confusion.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -438,10 +438,18 @@ public TokenDetails authorise(TokenParams params, AuthOptions options) throws Ab
 	}
 
 	/**
+	 * Alias of authorize() (0.9 RSA10l)
+	 */
+	public TokenDetails authorise(TokenParams params, AuthOptions options) throws AblyException {
+		Log.w(TAG, "authorise() alias will be removed in 1.0");
```
mattheworiordan: Oh, sorry, not sure how I missed that :/

trenouf:
So is this approved? Can I merge it?

lib/src/main/java/io/ably/lib/http/TokenAuth.java
```diff
@@ -36,21 +36,21 @@ public void setTokenDetails(TokenDetails tokenDetails) {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
-		Log.i("TokenAuth.authorise()", "");
+	public TokenDetails authorize(AuthOptions options, TokenParams params) throws AblyException {
```
paddybyers: I realise the spec doesn't require us to change this, but why not do it and keep things consistent?

paddybyers:
> So is this approved? Can I merge it?

Sorry, pls see my comment.

lib/src/main/java/io/ably/lib/http/TokenAuth.java
trenouf: @paddy You mean the param order?

paddybyers:
Thanks

PR #212, opened 2016-11-01
RSA8e,RSA9h,RSA10j: do not merge TokenParams and AuthOptions

trenouf:
In Auth.authorize(), createTokenRequest() and requestToken(), any
supplied TokenParams and/or AuthOptions replaces the previously
configured one, rather than merging with the previous one.

This is in the 0.8 spec, but is not being done in the 0.8 version of the
library in order to maintain compatibility.

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -128,19 +128,6 @@ public class Auth {
 		}
 
 		/**
-		 * Internal
-		 */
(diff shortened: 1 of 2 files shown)


trenouf:
@mattheworiordan @paddy Please review. Thanks.
I got a failure on io.ably.lib.test.realtime.RealtimeChannelHistoryTest.channelhistory_simple_text_withoutlistener on the PR build, but not on the equivalent push build. So I guess this is another one to add to the intermittent list.

paddybyers:
LGTM, thanks

PR #213, opened 2016-11-01
Fixed travis crash when using OpenJDK 7

psolstice:


paddybyers:
LGTM, thanks

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: java
 jdk:
   - oraclejdk8
+  - openjdk7
 # DISABLED openjdk7 as it still gives me a buffer overflow fault.
```
mattheworiordan: Can we remove these comments in they no longer apply?

psolstice: Sorry, forgot to do that
The link below should also be changed to https://github.com/travis-ci/travis-ci/issues/5227#issuecomment-165131913

PR #214, opened 2016-11-01
0.9 RTC8 in-place authorization

trenouf:
I'm doing a PR for this now so @paddy can start code review, but:

1. It includes the commit "RSA8e,RSA9h,RSA10j: do not merge TokenParams and AuthOptions", which also has its own PR. Once that other PR is merged, I can rebase to get rid of that commit.

2. I think this causes a rest test failure, but I haven't worked out what yet as travis is refusing to show me the build log.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -288,7 +288,7 @@ public class Http {
 			authHeader = "Basic " + Base64Coder.encodeString(auth.getBasicCredentials());
 		} else {
 			if (renew)
-				auth.authorize(null, null);
+				auth.renew();
 			else
 				auth.ensureValidAuth();
 			authHeader = "Bearer " + auth.getTokenAuth().getEncodedToken();
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -426,9 +426,20 @@ public class Http {
 			conn.setRequestProperty(HttpUtils.X_ABLY_VERSION_HEADER, HttpUtils.X_ABLY_VERSION_VALUE);
 			conn.setRequestProperty(HttpUtils.X_ABLY_LIB_HEADER, HttpUtils.X_ABLY_LIB_VALUE);
 
+			if (Log.level <= Log.VERBOSE) {
+				Log.v(TAG, "HTTP request: " + conn.getURL() + " " + method);
+				if (credentialsIncluded)
+					Log.v(TAG, "  " + AUTHORIZATION + ": " + authHeader);
+				Map<String, List<String>> props = conn.getRequestProperties();
+				for (Map.Entry<String, List<String>> entry : props.entrySet())
+					for (String val : entry.getValue())
+						Log.v(TAG, "  " + entry.getKey() + ": " + val);
+			}
 			/* send request body */
 			if(requestBody != null) {
 				writeRequestBody(requestBody, conn);
+				if (Log.level <= Log.VERBOSE)
+					Log.v(TAG, System.lineSeparator() + new String(requestBody.getEncoded()));
 			}
 
 			response = readResponse(conn);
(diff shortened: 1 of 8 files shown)


trenouf:
New push fixes the test failure, so this is now ready to go. Although it would be good to get the separate PR for "RSA8e,RSA9h,RSA10j: do not merge TokenParams and AuthOptions" in first.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -618,6 +635,7 @@ private Response readResponse(HttpURLConnection connection) throws IOException {
 
 		try {
 			response.body = readInputStream(is, response.contentLength);
+			Log.v(TAG, "\n" + new String(response.body));
```
paddybyers: Maybe `System.lineSeparator()` if we're being Windows-friendly?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -94,6 +94,53 @@ public StateIndication(ConnectionState state, ErrorInfo reason, String fallback,
 		}
 	}
 
+	/*************************************
+	 * a class that listens for state change
+	 * events for in-place authorization
+	 *************************************/
+	public static class ConnectionWaiter implements ConnectionStateListener {
```
paddybyers: You could avoid the `private Connection` if this wasn't static.

trenouf:
New push fixes both those comments. Ready to go once the travis builds are done.

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -582,6 +593,8 @@ private void writeRequestBody(RequestBody requestBody, HttpURLConnection conn) t
 		conn.setRequestProperty(CONTENT_TYPE, requestBody.getContentType());
 		conn.setRequestProperty(CONTENT_LENGTH, Integer.toString(length));
 		OutputStream os = conn.getOutputStream();
+		if (Log.level <= Log.VERBOSE)
+			Log.v(TAG, "\n" + new String(body));
```
paddybyers: You missed this one ....

trenouf: Oh bother.

trenouf:
Really fixed this time.

paddybyers:
Thanks

lib/src/main/java/io/ably/lib/types/ProtocolMessage.java
```diff
@@ -205,4 +214,42 @@ public JsonElement serialize(Action action, Type t, JsonSerializationContext ctx
...
+			int fieldCount = unpacker.unpackMapHeader();
+			for(int i = 0; i < fieldCount; i++) {
+				String fieldName = unpacker.unpackString().intern();
+				MessageFormat fieldFormat = unpacker.getNextFormat();
+				if(fieldFormat.equals(MessageFormat.NIL)) { unpacker.unpackNil(); continue; }
+
+				if(fieldName == "accessToken") {
+					accessToken = unpacker.unpackString();
+				} else {
+					System.out.println("Unexpected field: " + fieldName);
```
mattheworiordan: Why does this message not go to the logger?

paddybyers: That's cut+paste from other existing code that does the same thing. I'm not sure we should even log it tbh - it's legitimate for a message to contain unknown fields when there is a protocol change, for example.

mattheworiordan: Yes, but my point is that it should not be going to system out ever anywhere in the library, it should be going to the logger with a suitable log severity.  I am not bothered by which severity you think is appropriate, but I don't like that our library code (not test code) is incorrectly writing to STDOUT. Can we fix anywhere else this happens?

paddybyers:
> Can we fix anywhere else this happens?

We will review all instances and convert to a Logger message or remove.

paddybyers:
> I don't like that our library code (not test code) is incorrectly writing to STDOUT. Can we fix anywhere else this happens?

https://github.com/ably/ably-java/issues/217

PR #215, opened 2016-11-03
RSL6b test, log errors

psolstice:
Fixed decoding message bugs, added logging for decoding/decryption problems, added test for inconsistent message encoding

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/BaseMessage.java
+++ b/lib/src/main/java/io/ably/lib/types/BaseMessage.java
@@ -81,7 +81,7 @@ public class BaseMessage implements Cloneable {
 						try {
 							data = Base64Coder.decode((String) data);
 						}
-						catch (IllegalArgumentException e) { throw AblyException.fromThrowable(e); }
+						catch (IllegalArgumentException e) { break; }
 						continue;
 					}
 					if(xform == "utf-8") {
@@ -93,7 +93,7 @@ public class BaseMessage implements Cloneable {
 							String jsonText = ((String)data).trim();
 							data = Serialisation.gsonParser.parse(jsonText);
 						}
-						catch(JsonParseException e) { throw AblyException.fromThrowable(e); }
+						catch(JsonParseException e) { break; }
 						continue;
 					}
 					if(xform == "cipher") {
--- a/lib/src/main/java/io/ably/lib/types/BaseMessage.java
+++ b/lib/src/main/java/io/ably/lib/types/BaseMessage.java
@@ -81,9 +81,7 @@ public class BaseMessage implements Cloneable {
 						try {
 							data = Base64Coder.decode((String) data);
 						}
-						catch (IllegalArgumentException e) {
-							throw AblyException.fromThrowable(e);
-						}
+						catch (IllegalArgumentException e) { throw AblyException.fromThrowable(e); }
 						continue;
 					}
 					if(xform == "utf-8") {
@@ -99,12 +97,13 @@ public class BaseMessage implements Cloneable {
 						continue;
 					}
 					if(xform == "cipher") {
-						if (opts != null && opts.encrypted) {
+						if(opts != null && opts.encrypted) {
 							data = opts.getCipher().decrypt((byte[]) data);
 							continue;
 						}
-						else
-							Log.e(TAG, "Encrypted message received but encryption is not set up");
+						else {
+							Log.i(TAG, "Encrypted message received but encryption is not set up");
+						}
 					}
 					break;
 				}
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -78,7 +78,12 @@ public void decode(ChannelOptions opts) throws AblyException {
 					if(!match.matches()) break;
 					String xform = match.group(1).intern();
 					if(xform == "base64") {
-						data = Base64Coder.decode((String)data);
+						try {
+							data = Base64Coder.decode((String) data);
+						}
+						catch (IllegalArgumentException e) {
+							throw AblyException.fromThrowable(e);
```
paddybyers: What's going to happen to this exception? I don't think it will propagate to anywhere useful. In this case I think we should do the same as in the case of being unable to decrypt - ie leave the `data` and `encoding` unchanged, and log a message saying that the base64 decoding step failed.

lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -93,9 +98,13 @@ public void decode(ChannelOptions opts) throws AblyException {
...
 					}
-					if(xform == "cipher" && opts != null && opts.encrypted) {
-						data = opts.getCipher().decrypt((byte[])data);
-						continue;
+					if(xform == "cipher") {
+						if (opts != null && opts.encrypted) {
+							data = opts.getCipher().decrypt((byte[]) data);
+							continue;
+						}
+						else
```
paddybyers: If we have a compound if/then/else then it either needs braces everywhere, or nowhere, but not mixed like this. In in doublt, use braces.

lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -93,9 +98,13 @@ public void decode(ChannelOptions opts) throws AblyException {
...
+							Log.e(TAG, "Encrypted message received but encryption is not set up");
```
paddybyers: I'm not sure this is an error - I would make this `Log.i` I think.

Also lets say ".. encryption is not enabled on channel", and also include the channel name in the message pls.

lib/src/main/java/io/ably/lib/types/BaseMessage.java
psolstice: Without rethrowing the lib would crash on IllegalArgumentException. Exceptions is caught in Channel.onMessage or Channel.onPresense and logged. Rethrowing is used in the same method for JSON case.

lib/src/main/java/io/ably/lib/types/BaseMessage.java
psolstice: Sure, do you have a doc on code formatting style?

lib/src/main/java/io/ably/lib/types/BaseMessage.java
psolstice: I would if it was easy to get channel name at this point. Most I can do is throw custom AblyException, catch it in Channel.onMessage and log it there. But it seems like a lot of effort for a minor thing

paddybyers:
> Without rethrowing the lib would crash on IllegalArgumentException

No, without catching the lib would crash on IllegalArgumentException.

> Rethrowing is used in the same method for JSON case.

Ok, well we definitely need to do the same thing in each case.

If we throw here, then the client doesn't see the event; only a log message. If we catch and do nothing except log it, the client will get an event, with an `encoding` field that says there are un-unwrapped JSON/Base64 encoding steps. I can't say that's especially useful either TBH but we should decide.

paddybyers:
> do you have a doc on code formatting style?

No we don't. Use the same coding style as is there already pls to the extent that it's obvious what that is :)

psolstice:
By rethrowing I mean of course throwing it as a different exception. I just saw a bug (crash in case of wrong base64 data) and fixed it in a way similar to what was used in a nearby statement in the same method.

paddybyers:
@psolstice [RSL6b](http://docs.ably.io/client-lib-development-guide/features/#RSL6b) says that invalid Base64 does not prevent the event being delivered to the listener. So I think throwing here is wrong; the decode exception should caught, and a log message written out, and the `Message` with unchanged `data` and `encoding` is emitted to subscribers in the normal way.

I think we should make the JSON decode failure do the same.

psolstice:
I think you're right, I'll change the behavior

lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -81,7 +81,7 @@ public void decode(ChannelOptions opts) throws AblyException {
 						try {
 							data = Base64Coder.decode((String) data);
 						}
-						catch (IllegalArgumentException e) { throw AblyException.fromThrowable(e); }
+						catch (IllegalArgumentException e) { break; }
```
paddybyers: Sorry, I should have been clearer. I think we need a log message (`Log.e()` I guess) in this case as well.

paddybyers:
Thanks

PR #216, opened 2016-11-04
Notify connection state change listener of auth/reauth failures

psolstice:


lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -71,6 +71,13 @@ public void onAuthUpdated() {
 	}
 
 	/**
+	 * Authentication error occurred
+	 */
+	public void onAuthError(ErrorInfo errorInfo) {
```
paddybyers: Can't we make this package scope so it's not visible to users of the library (and `onAuthUpdated()`) ?

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -419,10 +419,23 @@ public TokenDetails authorize(TokenParams params, AuthOptions options) throws Ab
...
+		try {
+			TokenDetails tokenDetails = tokenAuth.authorize(params, options, /*force=*/true);
+			ably.onAuthUpdated();
 
-		ably.onAuthUpdated();
-		return tokenDetails;
+			return tokenDetails;
+		}
+		catch (AblyException e) {
+			if(e.errorInfo != null) {
```
paddybyers: What happens if `e.errorInfo == null` ? Can that happen? We need to call `onAuthError()` in any case.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -298,6 +297,31 @@ public void onAuthUpdated() {
...
+	 * @param errorInfo Error associated with unsuccessful authentication
+	 */
+	public void onAuthError(ErrorInfo errorInfo) {
+		Log.i(TAG, String.format("onAuthError: (%d) %s", errorInfo.code, errorInfo.message));
+		switch (state.state) {
+			case connecting:
+				ITransport transport = this.transport;
+				if (transport != null)
+					/* onTransportUnavailable will send state change event and set transport to null */
+					onTransportUnavailable(transport, null, errorInfo);
```
paddybyers: This looks right to me but @tpr can you have a quick look pls?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -298,6 +297,31 @@ public void onAuthUpdated() {
...
+				break;
+
+			case connected:
+				/* stay connected but notify of authentication error */
+				setState(new StateIndication(ConnectionState.connected, errorInfo));
```
paddybyers: I'm not sure what this achieves. The auth error might have been triggered by a an error response to a REST request; I don't see that that has anything to do with the connection, and I don't think we should be setting any error here.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
```diff
@@ -94,6 +97,7 @@ public void reauth_tokenDetails() {
 			Auth.AuthOptions authOptions = new Auth.AuthOptions();
 			authOptions.key = optsTestVars.keys[0].keyStr;
 			authOptions.tokenDetails = secondToken;
+			//authOptions.authUrl = "https://nonexistent-domain-abcdef.com";
```
paddybyers: Did you intend for this to stay here?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
psolstice: This is what is done in onDisconnect() so I did the same here

paddybyers: Now I think of it, why do this at all? The server will disconnect the transport with the relevant error if it thinks that its auth params are not valid. I don't think there is any need for this here. It is possible that the transport is already using a new, valid, token, for example, and some operation elsewhere just failed and triggered the `onAuthError()`

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
psolstice: There's no other way to notify user of auth error if reauth is initiated by the server (not implemented yet but nonetheless). I can remote it if you want.

lib/src/main/java/io/ably/lib/rest/Auth.java
psolstice: It can be specifically constructed with errorInfo set to null and thrown from authCallback. I took a look and I think it's not the expected case and the check needs to be removed altogether

lib/src/test/java/io/ably/lib/test/realtime/RealtimeReauthTest.java
psolstice: Sorry, missed that one

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
psolstice: The error can happen earlier that token is transmitted to the server and server has a chance to reject it. If authUrl lookup fails or authCallback throws an exception for example

paddybyers: OK, I see that if the transport is closed because of an expired token, and the subsequent attempt to get a new token fails for that reason, this is how we will communicate that failure to the client.

In the case that the transport is still intact, and the server signals that the token is about to expire, and the preemptive token renewal fails: I think in this case we will just log the error instead of setting any error state on the existing connection - and then token will shortly expire and there will be a new attempt. If that second attempt succeeds, we're ok; if it fails, we're in the `connecting` case above, and the client will see that in a connection state change event.

@trenouf does that sound right?

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
psolstice: No we can't because these methods are accessed across our subpackages (realtime, rest, transport etc)

paddybyers:
> No we can't because these methods are accessed across our subpackages (realtime, rest, transport etc)

But isn't this only called from `rest.Auth` ? And that is calling the method declared on `rest.AblyRest` ?

psolstice:
Unfortunately there are three parts here: AblyRest, AblyRealtime and ConnectionManager that communicate. These three are technically different java packages

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
mattheworiordan: Is there not a convention at least to indicate that this is not a method people should be calling in the documentation at least then?  Seems horrible that we are exposing all these methods to users, especially in langs like Java where most people use IDEs with intellisense.

mattheworiordan: In fact, whatever calls `onAuthError` could instead just call `connection.connectionManager.onAuthError` surely which will at least reduce the exposure of methods / objects that users should not be using.

lib/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -140,4 +140,11 @@ public void statsAsync(Param[] params, Callback<AsyncPaginatedResult<Stats>> cal
 	public void onAuthUpdated(String token) throws AblyException {
 		/* Default is to do nothing. Overridden by subclass. */
 	}
+
+	/**
+	 * Authentication error occurred
+	 */
+	public void onAuthError(ErrorInfo errorInfo) {
```
mattheworiordan: As above, can we not remove this completely from this object and hide in a sub-object so that we are only polluting the REST object with one attribute / object?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
mattheworiordan: Well a connection can also emit an error event, see http://docs.ably.io/client-lib-development-guide/features/#RTN4g.  If the state has not changed, it would be more appropriate to emit that as state changes should only be emitted when there is indeed a state change no?

psolstice:
In java library connection cannot emit anything but ConnectionStateChange. If you want to incorporate error reporting to that you'll have to emit state->state change. Extending this class can lead to changes that will break client apps.

psolstice:
No, you can't use connection.connectionManager, you don't have access to it
It is defined in AblyRealtime but in the Auth class you have access to its ancestor (AblyRest) so the call goes exactly like onAuthUpdate() through virtual method of the base class. I don't know why it was done like that in the first place

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
trenouf: Matt:
> In fact, whatever calls onAuthError could instead just call 
> connection.connectionManager.onAuthError surely

No, because onAuthError is called from rest.Auth, which doesn't know anything about realtime and connections.

Paddy:
> But isn't this only called from rest.Auth ? And that is calling the method declared on 
> rest.AblyRest ?

Yes, this might work. Make AblyRest.onAuthError protected so (a) it cannot be called from outside the rest package, but (b) it can be overridden by AblyRealtime, which is a subclass of AblyRest.

The same thing applies to onAuthUpdated that I added in the in-place auth changes. It too should be protected.

paddybyers:
> Is there not a convention at least to indicate that this is not a method people should be calling in the documentation at least then?

There is a way to ensure it's not visible to the client. We just haven't agreed what yet, but we will.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
trenouf: @paddy Are you questioning whether the spec is right here? Ignoring that question, I think Peter's change implements the spec as written. However there is the issue mentioned by @matt (and I think discussed elsewhere) that the connected case does a connected->connected state change as a way of staying in connected state but emitting the error. RTN4g says there should be a way to emit an error event, but that does not exist in the java lib.

paddybyers:
> Are you questioning whether the spec is right here?

No ....

> Peter's change implements the spec as written.

ok

> RTN4g says there should be a way to emit an error event, but that does not exist in the java lib.

Yes, agreed, and we will address that as a separate PR.

mattheworiordan:
> In java library connection cannot emit anything but ConnectionStateChange. If you want to incorporate error reporting to that you'll have to emit state->state change. Extending this class can lead to changes that will break client apps.

@psolstice OK, but this 0.9 change is breaking anyway, so I don't see this as a problem.  I agree it should not go into the 0.8 version, but it's part of the spec (even in 0.8) and as such as we have to support emitting Error events for Connections and Channels.  If you deem this outside of the scope of this PR please raise a separate issue to address this and come back and fix the implementation here.

> No, you can't use connection.connectionManager, you don't have access to it
It is defined in AblyRealtime but in the Auth class you have access to its ancestor (AblyRest) so the call goes exactly like onAuthUpdate() through virtual method of the base class. I don't know why it was done like that in the first place

Well I really dislike adding more and more public methods to this object that are not part of this API. @paddybyers what do you suggest to avoid adding more & more "noise" to the API?

psolstice:
> @psolstice OK, but this 0.9 change is breaking anyway, so I don't see this as a problem. I agree it should not go into the 0.8 version, but it's part of the spec (even in 0.8) and as such as we have to support emitting Error events for Connections and Channels. If you deem this outside of the scope of this PR please raise a separate issue to address this and come back and fix the implementation here.

I wouldn't be happy if some lib broke my app during 0.8->0.9 version bump. Anyway we'll agree on alternative method to emit ErrorInfo soon.

> Well I really dislike adding more and more public methods to this object that are not part of this API. @paddybyers what do you suggest to avoid adding more & more "noise" to the API?

I've already fixed this visibility problem in latest commit. In future I'll have it in mind

paddybyers:
> If you deem this outside of the scope of this PR please raise a separate issue to address this and come back and fix the implementation here.

@mattheworiordan we have this in hand (see earlier comments).

> @paddybyers what do you suggest to avoid adding more & more "noise" to the API?

@mattheworiordan we have this in hand (see earlier comments).

paddybyers:
> I wouldn't be happy if some lib broke my app during 0.8->0.9 version bump

In updating the spec from 0.8 to 0.9 there are changes in the behaviour of the library and these mean that the library is not compatible. In making those changes we have decided that, although we would have preferred to avoid it, there will be some API incompatibilities. Although we would probably have a different policy post-1.0, that's the reality with 0.9.

paddybyers:
@psolstice can you rebase this and say whether or not you think you have responded to all the comments?

psolstice:
Rebased, fixed, addressed

paddybyers:
Thanks

PR #218, opened 2016-11-04
0.9 RTL2,3 channel suspended state

trenouf:
This PR is the changes for the new channel suspended state, plus the new resumed flag in the attached message, and the new ChannelStateChange object.

I believe the code is complete, but there is no test coverage for it yet. @paddy I don't know whether you want to
1. approve and merge it anyway on the basis that existing tests pass, and ask @peter.shugalev to write tests in a different PR; or
2. not approve it, and ask @peter.shugalev to write tests in this PR.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -16,7 +16,7 @@ public class Defaults {
 	public static final int TIMEOUT_CONNECT         = 15000;
 	public static final int TIMEOUT_DISCONNECT      = 30000;
 	public static final int TIMEOUT_SUSPEND         = 120000;
-	public static int TIMEOUT_CHANNEL_RETRY			= 15000;
+	public static int TIMEOUT_CHANNEL_RETRY         = 15000;
 	/* TO313 */
 	public static final int TIMEOUT_HTTP_OPEN = 4000;
 	/* TO314 */
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -16,6 +16,7 @@ public class Defaults {
 	public static final int TIMEOUT_CONNECT         = 15000;
 	public static final int TIMEOUT_DISCONNECT      = 30000;
 	public static final int TIMEOUT_SUSPEND         = 120000;
+	public static int TIMEOUT_CHANNEL_RETRY			= 15000;
 	/* TO313 */
 	public static final int TIMEOUT_HTTP_OPEN = 4000;
 	/* TO314 */
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -21,13 +21,13 @@ public class Defaults {
 	/* TO314 */
 	public static final int TIMEOUT_HTTP_REQUEST = 15000;
 	/* DF1b */
-	public static final long realtimeRequestTimeout = 10000L;
+	public static long realtimeRequestTimeout = 10000L;
 	/* CD2h (but no default in the spec) */
 	public static final long maxIdleInterval = 20000L;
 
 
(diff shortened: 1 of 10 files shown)


paddybyers:
> @paddy I don't know whether you want to ....

I want at least some tests covering the principal requirements before we merge.

mattheworiordan:
@psolstice I suspect you're going to need to pick this up at some point... just a head's up!

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -21,6 +21,7 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import javax.management.DescriptorAccess;
```
paddybyers: I assume this is supported on Android?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1113,9 +1114,9 @@ public void detach_on_clean_connection_preserves_channel() {
 			assertNotEquals("Verify new connection established", recoverConnectionId, ably.connection.id);
 			assertNotNull("Verify error was returned with connected state", ably.connection.reason);
 
-			/* verify existing channel is detached but not removed */
-			(new ChannelWaiter(channel)).waitFor(ChannelState.detached);
-			assertEquals("Verify attached state reached", channel.state, ChannelState.detached);
+			/* verify existing channel is failed but not removed */
```
paddybyers: Can we have a spec reference for this change pls in the comment?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1126,4 +1127,142 @@ public void detach_on_clean_connection_preserves_channel() {
...
+			final String channelName = "test_state_channel";
+			final Channel channel = ably.channels.get(channelName);
+			channel.attach();
+
+			ChannelWaiter channelWaiter = new ChannelWaiter(channel);
+			channelWaiter.waitFor(ChannelState.attached);
+			assertEquals("Verify attached state reached", channel.state, ChannelState.attached);
+
+			/* switch to suspended state */
+			ably.connection.connectionManager.requestState(ConnectionState.suspended);
```
paddybyers: I'm inclined to think that this should be `notifyState()` - I thought `requestState()` is for the situation where the `ConnectionManager` needs to take some action to put it into that state - eg it calls `connectImpl()` in response to a call to `requestState("connecting")`

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
psolstice: Hm, I have no idea what is it doing here and why is it needed here, will remove it

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
psolstice: It's Tim's line, I can't say right now but I can look for it

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -16,6 +16,7 @@
 	public static final int TIMEOUT_CONNECT         = 15000;
 	public static final int TIMEOUT_DISCONNECT      = 30000;
 	public static final int TIMEOUT_SUSPEND         = 120000;
+	public static int TIMEOUT_CHANNEL_RETRY			= 15000;
```
paddybyers: It looks like you have tabs in here. Tabs only ever at the beginning of a line pls.

paddybyers:
> I think you haven't yet addressed https://github.com/ably/ably-java/pull/218/files#r90880457

@psolstice pls can you do this and then we can land this. Does 0.9 need to be rebased first?

paddybyers:
Thanks

PR #219, opened 2016-11-07
Added logging, clarified code

psolstice:


lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -5,15 +5,7 @@
...
-import io.ably.lib.types.AblyException;
-import io.ably.lib.types.ChannelOptions;
-import io.ably.lib.types.ErrorInfo;
-import io.ably.lib.types.Message;
-import io.ably.lib.types.MessageSerializer;
-import io.ably.lib.types.PaginatedResult;
-import io.ably.lib.types.Param;
-import io.ably.lib.types.PresenceMessage;
-import io.ably.lib.types.ProtocolMessage;
+import io.ably.lib.types.*;
```
mattheworiordan: I believe @paddybyers is not a big fan of these broad imports, but I'll let him chime in :)

psolstice: It was an automatic change by IDEA.

paddybyers:
> It was an automatic change by IDEA.

Well I prefer that the imports are explicit but if by default IDEA is going to give warnings about it then I don't see the point in fighting it ...

psolstice:
It gave no warning whatsoever. It just changed it. It will change it again in future if we revert it now. So I don't see the point to fight it too.

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -383,8 +375,10 @@ private void onMessage(ProtocolMessage message) {
 			Message msg = messages[i];
 			try {
 				msg.decode(options);
+			} catch (MessageDecodeException e) {
+				Log.e(TAG, String.format("%s on channel %s", e.errorInfo.message, name));
 			} catch(AblyException e) {
-				Log.e(TAG, "Unexpected exception decrypting message", e);
+				Log.e(TAG, String.format("Unexpected exception decrypting message on channel %s", name), e);
```
paddybyers: Will this only be a decrypt exception? Ie should it say "Unexpected exception decoding message" ?

Actually, what exception could occur other than a `MessageDecodeException` ?

mattheworiordan: @paddybyers I saw the same thing, but then realised the previous code was like this before... I would agree we should instead catch `MessageDecodeException` if that is the exception we expect for this message, and perhaps another generic AblyException catch all?

psolstice: Exception other than `MessageDecodeException` can be thrown only from here
[https://github.com/psolstice/ably-java/blob/RSL6_test/lib/src/main/java/io/ably/lib/types/BaseMessage.java#L105]
And as you can see it's a decryption exception

paddybyers:
> And as you can see it's a decryption exception

That's also a `MessageDecodeException` then, because an failure to decrypt is logged, and the `data`/`encoding` are left in the unencrypted state, and message processing continues (ie the message is broadcast to listeners in that state. (See http://docs.ably.io/client-lib-development-guide/features/#RSL6b)

paddybyers:
Thanks, LGTM

paddybyers:
LGTM, thanks

Issue #222, opened 2016-11-09
ConnectionManager.failQueuedMessages() does not remove messages once the callback is called


paddybyers:
https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L815-L826

mattheworiordan:
Is this issue now resolved with #225?

paddybyers:
Yes

PR #225, opened 2016-11-10
Messages are now removed from the queue after onError() call

psolstice:
Fixed issue #222 along with a test to ensure messages are indeed removed from the queue

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -814,20 +814,16 @@ public class ConnectionManager implements Runnable, ConnectListener {
 
 	private void failQueuedMessages(ErrorInfo reason) {
 		synchronized(this) {
-			for (int n=0; n<queuedMessages.size(); ) {
-				QueuedMessage queued = queuedMessages.get(n);
-				try {
-					if(queued.listener != null) {
-						queuedMessages.remove(n);
+			for (QueuedMessage queued: queuedMessages) {
+				if (queued.listener != null) {
+					try {
 						queued.listener.onError(reason);
+					} catch (Throwable t) {
+						Log.e(TAG, "failQueuedMessages(): Unexpected error calling listener", t);
 					}
-					else {
-						n++;
-					}
-				} catch (Throwable t) {
-					Log.e(TAG, "failQueuedMessages(): Unexpected error calling listener", t);
 				}
 			}
+			queuedMessages.clear();
 		}
 	}
 
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -814,10 +814,16 @@ public class ConnectionManager implements Runnable, ConnectListener {
 
 	private void failQueuedMessages(ErrorInfo reason) {
 		synchronized(this) {
-			for(QueuedMessage queued : queuedMessages) {
+			for (int n=0; n<queuedMessages.size(); ) {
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -814,20 +814,16 @@ private void sendQueuedMessages() {
...
+						Log.e(TAG, "failQueuedMessages(): Unexpected error calling listener", t);
 					}
-					else {
-						n++;
-					}
-				} catch (Throwable t) {
-					Log.e(TAG, "failQueuedMessages(): Unexpected error calling listener", t);
 				}
 			}
+			queuedMessages.clear();
```
psolstice: Yes, it can mess up messages and even crash the app. I can change it but then we can get into infinite loop if onError() adds message and then we will call onError() on the new message and it adds a message too and so on

paddybyers: What about something like https://github.com/ably/ably-java/blob/413edbc288d54657aac57e24426df0f8375fb17d/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L804 ?

psolstice:
Same, infinite loop in case messages keep coming in onError call. There is no ideal solution here

paddybyers:
> Same, infinite loop in case messages keep coming in onError call. There is no ideal solution here

ok, fine, I agree that provided each callback that was there to begin with is called, and it is clear at the end, we've done what we need to do.

I suppose the only other possibility is to call the error callback immediately, instead of adding, if there is an attempt to add a new entry to the queue whilst the enumeration is in progress.

psolstice:
> I suppose the only other possibility is to call the error callback immediately, instead of adding, if there is an attempt to add a new entry to the queue whilst the enumeration is in progress.
Ok, stack overflow instead of infinite loop

The most foolproof variant I came up with is something like this:
`oldMessageQueue = queuedMessages;
queuedMessages = new ListArray<>();
for (... oldMessageQueue...) { ... }
queuedMessages.clear()`

psolstice:
Actually it doesn't matter because I just figured out that adding a new message is technically impossible in `onError()`. `failMessageQueue` is called only if `state.queueEvents` is `false`. `queuedMessages.add()` can be called only if `state.queueEvents` is `true` which is not the case while we're in `failMessageQueue`.

paddybyers:
> Actually it doesn't matter because I just figured out that adding a new message is technically impossible in onError(). failMessageQueue is called only if state.queueEvents is false. queuedMessages.add() can be called only if state.queueEvents is true which is not the case while we're in failMessageQueue.

ok :)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -277,4 +279,50 @@ public void connect_unknown_recover_fail() {
...
+			// transition once more to ensure onError() won't be called twice
+			ably.connection.connectionManager.requestState(ConnectionState.closed);
+			try { Thread.sleep(500); } catch (InterruptedException e) {}
+
+			// onError() should be called only once
+			assertEquals("Verifying number of onError() calls", numberOfErrors[0], 1);
+		}
+		catch (AblyException e) {
+			e.printStackTrace();
+			fail("Unexepcted exception");
```
paddybyers: oops, typo, probably cut+paste so there's at least one other :)

paddybyers:
I'll merge this if you can fix that typo pls.

paddybyers:
LGTM, thanks

PR #227, opened 2016-11-14
Fixed BuildConfig problems

psolstice:


mattheworiordan:
Thanks @psolstice, not clear on what problem specifically this was aiming to solve though?

paddybyers:
> not clear on what problem specifically this was aiming to solve though?

We had a class name conflict with `BuildConfig` being included twice, once by us and once by the android build framework.

psolstice:
> We had a class name conflict with BuildConfig being included twice, once by us and once by the android build framework.

Not quite like that, we had two modules: java and lib each producing their own copy on BuildConfig, I removed BuildConfig from java module. Then I renamed it to AblyBuildConfig to prevent potential conflict with BuildConfig from Android framework. Just in case.

paddybyers:
LGTM

PR #236, opened 2016-11-16
Fixed issue 233, made changes to allow ITransport mocking

psolstice:


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -513,6 +518,7 @@ private void handleStateChange(StateIndication stateChange) {
 		if(stateChange.state == ConnectionState.disconnected) {
 			switch(state.state) {
 			case connecting:
+			case disconnected:
```
paddybyers: I think we no longer need this line now. Did you re-run the tests without it?

paddybyers:
Thanks

PR #237, opened 2016-11-18
RSA10k; token expiration now takes client-server time delta into account

psolstice:
It's different from https://github.com/ably/ably-java/pull/158 in two aspects
1. Time offset is shared between all the connections.
2. Offset is used to calculate token expiration time if client time is off. This solves a lot of problems when client time is out of sync with server.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/TokenAuth.java
+++ b/lib/src/main/java/io/ably/lib/http/TokenAuth.java
@@ -61,7 +61,7 @@ public class TokenAuth {
 	}
 
 	private static boolean tokenValid(TokenDetails tokenDetails) {
-		return tokenDetails.expires > Auth.timestamp();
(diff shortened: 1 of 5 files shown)


mattheworiordan:
Looks good, but do we have suitable test coverage on `queryTime` to check that it's doing what it should be doing?

paddybyers:
Is there any way we can detect changes in the client's date/time settings, and invalidate the cached time delta?

If not, should we maybe invalidate the cached delta after a period of time? Otherwise if the delta is wrong for some reason, then things are permanently broken :(
Otherwise

paddybyers:
> Looks good, but do we have suitable test coverage on queryTime to check that it's doing what it should be doing?

@psolstice did you look at this?

paddybyers:
> Added test for queryTime caching behaviour

LGTM

paddybyers:
Thanks

PR #238, opened 2016-11-21
RTN22: server-initiated reauthorization

psolstice:


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -711,6 +725,25 @@ public void run() {
 		}
 	}
 
+	private void handleReauth() {
+		pendingReauth = false;
```
paddybyers: Should this only be cleared once reauth is successful? If we don't clear it here, and the `authCallback` fails (with an exception), when would there be a new attempt? The next time the thread exits `tryWait()` ?

lib/src/main/java/io/ably/lib/transport/ITransport.java
```diff
@@ -51,6 +51,7 @@ public TransportParams() {
 
 		public Param[] getConnectParams(Param[] baseParams) {
 			List<Param> paramList = new ArrayList<Param>(Arrays.asList(baseParams));
+			paramList.add(new Param("v", HttpUtils.X_ABLY_VERSION_VALUE));
```
mattheworiordan: Can we have a test for this please to prevent regressions?  I could have sworn this has been implemented previously though as I recall seeing tests, perhaps they only exist for REST?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -166,23 +167,32 @@ public void connect_while_disconnected() {
...
 		try {
 			final Setup.TestVars optsTestVars = Setup.getTestVars();
 			ClientOptions optsForToken = optsTestVars.createOptions(optsTestVars.keys[0].keyStr);
 			optsForToken.logLevel = Log.VERBOSE;
 			final AblyRest ablyForToken = new AblyRest(optsForToken);
-			TokenDetails tokenDetails = ablyForToken.auth.requestToken(new TokenParams() {{ ttl = 8000L; }}, null);
+			TokenDetails tokenDetails = ablyForToken.auth.requestToken(new TokenParams() {{ ttl = 38000L; }}, null);
 			assertNotNull("Expected token value", tokenDetails.token);
 
+            final boolean[] flags = new boolean[] {
```
mattheworiordan: Is this indentation not wrong?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -200,26 +210,27 @@ public Object getTokenRequest(TokenParams params) throws AblyException {
 			opts.logLevel = Log.VERBOSE;
 			AblyRealtime ably = new AblyRealtime(opts);
 
-			/* wait for connected state */
-			ConnectionWaiter connectionWaiter = new ConnectionWaiter(ably.connection);
-			connectionWaiter.waitFor(ConnectionState.connected);
-			assertEquals("Verify connected state is reached", ConnectionState.connected, ably.connection.state);
+            ably.connection.on(new ConnectionStateListener() {
```
mattheworiordan: Identation looks odd agian

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -166,23 +167,32 @@ public void connect_while_disconnected() {
...
 	/**
-	 * Verify that the connection enters the disconnected state, after a token
-	 * used for successful connection expires
+	 * Verify that the server issues reauth message 30 seconds before token expiration time, authCallback is
+	 * called to obtain new token and in-place re-authorization takes place with connection staying in connected
+	 * state
 	 */
 	@Test
-	public void connect_token_expire_disconnected() {
+	public void connect_token_expire_inplace_reauth() {
```
mattheworiordan: Whilst I see the point of changing this test, we do need to know still that a client does not authenticate in response to the server does move to the disconnected state and then attempt a reconnect.  I would prefer to have two tests in fact to make sure we're not breaking the behaviour whereby a token expires without the change to renew (or perhaps with a failed renew etc).

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
psolstice: I'm not sure how many attempts should be given for reauth. If we don't clear it the thread will exit `tryWait()` immediately and will consume all the CPU time if `authCallback` keeps failing.

lib/src/main/java/io/ably/lib/transport/ITransport.java
psolstice: What do you want to test here? There won't be reauth messages from server if I remove this line.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
psolstice: Behaviour has changes and so did the test. I can add a new test but it will have to wait 30+ seconds for the disconnect message and I can't fix it by changing default timeouts on the client side.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
psolstice: Probably tabs/spaces problem, i'll fix it

lib/src/main/java/io/ably/lib/transport/ITransport.java
mattheworiordan: See https://github.com/ably/ably-java/blob/master/lib/src/test/java/io/ably/lib/test/rest/HttpHeaderTest.java.  That test prevents a regression where the header is no longer sent mistakenly, or incorrectly sent.  Same thing no?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -210,25 +202,88 @@ public Object getTokenRequest(TokenParams params) throws AblyException {
...
+	 * state
+	 */
+	@Test
+	public void connect_token_expire_inplace_reauth() {
+		try {
+			final Setup.TestVars optsTestVars = Setup.getTestVars();
+			ClientOptions optsForToken = optsTestVars.createOptions(optsTestVars.keys[0].keyStr);
+			optsForToken.logLevel = Log.VERBOSE;
+			final AblyRest ablyForToken = new AblyRest(optsForToken);
+			TokenDetails tokenDetails = ablyForToken.auth.requestToken(new TokenParams() {{ ttl = 34000L; }}, null);
```
mattheworiordan: Why 34s, if we want to test what happens when the token expires, why not go for a short TTL?

mattheworiordan: Sorry, I didn't realise this was the inplace auth test, ignore me 😕

paddybyers:
LGTM

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: I guess if the single attempt fails, then a few seconds later the token will actually expire, and there will be a new transport after getting a new token. So the existing retry/fallback/backoff logic associated with the reconnection attempt will do what we need better than retrying the preemptive in-transport reauth

paddybyers:
@psolstice are you ok for me to merge this? Should we rebase 0.9 first?

psolstice:
It's better to merge before rebase. Less work overall

paddybyers:
Thanks

PR #243, opened 2016-12-02
Set minimum Android SDK version to 14 (4.0+)

psolstice:


paddybyers:
Thanks

mattheworiordan:
Out of interest why did we need to bump, and should we put that in the README?

mattheworiordan:
Oops, I see it is mentioned in README.

paddybyers:
> Out of interest why did we need to bump

Since we're now an AAR, the manifest metadata is relevant (and enforced by the toolchain) when including the library; so whereas before it was irrelevant we needed to set it to something, and 4.0 seemed a reasonable balance between being recent enough to be supportable and covering a large proportion of the device population

psolstice:
According to Google we're covering ~98.6% active devices

PR #250, opened 2016-12-07
New presence features in 0.9 specs

psolstice:


lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1941,4 +1941,188 @@ public void realtime_presence_get_throws_when_channel_failed() throws AblyExcept
...
+			try {
+				synchronized (presenceWaiter) {
+					while (!presenceWaiter[0])
+						presenceWaiter.wait();
+				}
+			}
+			catch (InterruptedException e) {
+			}
+
+			ably.connection.connectionManager.requestState(ConnectionState.suspended);
```
paddybyers: I think this sequence only really affects the client state, and doesn't do anything to the server state. So when you later check that the server has the expected presence data, that doesn't confirm that the client re-entered; the server might just have the same presence members from when the client first entered.

Is it possible to modify the data cached in the internal presence map, so that updated member is reentered with the channel is re-attached, and you can confirm that it was updated on the server?

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1979,38 +1988,64 @@ public void onSuccess() {
...
 				Thread.sleep(500);
-				PresenceMessage[] presenceMessages = channel.presence.get(testClientId1, true);
-				assertNotNull("Verify there are presence messages after reattach");
-				assertEquals("Verify there is exactly one presence message", presenceMessages.length, 1);
-				assertEquals("Verify presence message data is the same as before", presenceMessages[0].data, presenceData);
-			}
-			catch (InterruptedException e) {
-			}
+				assertEquals("Verify correct presence message data has been received",
+						channel.presence.get(testClientId2, true)[0].data, presenceData);
```
paddybyers: I still don't think that this does what you think. Calling (realtime)Channel.presence.get()` gets the value from the local `PresenceMap`; so by this test you still don't know that the update has reached the server as a result of reattaching after `suspended`.

If this was a REST `Channel.presence.get()`, you'd be certain that you're getting the value from the server.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1941,4 +1937,321 @@ public void realtime_presence_get_throws_when_channel_failed() throws AblyExcept
...
+				ably.connection.connectionManager.onMessage(null, protocolMessage);
+			}
+
+			int n = 0;
+			for (PresenceMessage testMsg: testData) {
+				if (testMsg.data != wontPass) {
+					PresenceMessage factualMsg = n < presenceMessages.size() ? presenceMessages.get(n++) : null;
+					assertTrue("Verify message passed newness test",
+							factualMsg != null && factualMsg.id.equals(testMsg.id));
+				}
```
mattheworiordan: Might be nice to also test that the total presenceMessages count matches that the testData...

paddybyers: @psolstice can you do this quickly pls and then we can merge?

PR #252, opened 2016-12-09
Support for emitting UPDATE event: RTN4h, RTL2g, RTC8a, RTN24, tests

psolstice:


lib/src/main/java/io/ably/lib/realtime/ChannelStateListener.java
```diff
@@ -37,6 +37,13 @@
 			this.reason = reason;
 			this.resumed = resumed;
 		}
+
+		/* constructor for UPDATE event */
```
paddybyers: I think we could use a factory constructor here, so it can have a meaningful name instead of needing a comment like this

lib/src/main/java/io/ably/lib/realtime/ConnectionStateListener.java
```diff
@@ -18,6 +18,13 @@ public ConnectionStateChange(ConnectionState previous, ConnectionState current,
 			this.retryIn = retryIn;
 			this.reason = reason;
 		}
+
+		/* constructor for UPDATE event */
```
paddybyers: See aove comment re factory constructor

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -240,7 +241,9 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
 	/**
 	 * Verify that the server issues reauth message 30 seconds before token expiration time, authCallback is
 	 * called to obtain new token and in-place re-authorization takes place with connection staying in connected
-	 * state
+	 * state. Also tests if UPDATE event is delivered on the connection
+	 *
+	 * Test for RTN4h, RTC8a, RTN24 features
```
paddybyers: I think if I'm being pedantic this should say `RTC8a1`

paddybyers:
Thanks

PR #253, opened 2016-12-14
RTL4, RTL5 0.9 tweaks

psolstice:


psolstice:
Added RTL12, RTL13 changes & tests on top of this PR

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -681,7 +764,7 @@ private void sendQueuedMessages() {
 		queuedMessages.clear();
 	}
 
-	private void failQueuedMessages(ErrorInfo reason) {
+	private synchronized void failQueuedMessages(ErrorInfo reason) {
```
paddybyers: I don't think you can call back into user code with the `Channel` monitor held. I think you need to take a copy of the listener array whilst holding the lock, and then release the lock to make the callbacks.

paddybyers:
Thanks

PR #254, opened 2016-12-21
New approach for update event, resolved conflicts

psolstice:


lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -949,7 +949,19 @@ void onChannelMessage(ProtocolMessage msg) {
...
+	public void emit(ChannelState state, ChannelStateListener.ChannelStateChange channelStateChange) {
+		super.emit(state.getChannelEvent(), channelStateChange);
+	}
+
+	public void on(ChannelState state, ChannelStateListener listener) {
+		super.on(state.getChannelEvent(), listener);
+	}
+
+	public void once(ChannelState state, ChannelStateListener listener) {
+		super.on(state.getChannelEvent(), listener);
```
paddybyers: `super.once()` ?

psolstice: Of course. Thank you for noticing!

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1221,14 +1222,14 @@ public void onChannelStateChanged(ChannelStateChange stateChange) {
 				channel = channelName;
 				flags |= 1 << Flag.resumed.ordinal();
 			}};
-			ably.connection.connectionManager.onMessage(attachedMessage);
+			ably.connection.connectionManager.onMessage(null, attachedMessage);
```
paddybyers: When did this API change? This PR doesn't contain that change

psolstice: It was in presence PR. There was a bug of not setting transport to null on transport disconnect and then under some race conditions allowing receiving messages from dead transport. So we need to pass transport as an argument to onMessage() to detect dead transports and null is special case to bypass check. Just for tests.

psolstice: This PR resolves conflicts between two previous PRs, that's why it's here

PR #255, opened 2016-12-22
RSN2, RSN3

psolstice:


paddybyers:
Since the RSN2 requirement is "Methods should exist to check if a channel exists or iterate through the existing channels" I think we should add tests that do that (eg with `contains()` or `iterator()`)

paddybyers:
LGTM, thanks

PR #264, opened 2017-01-02
0.8 auth update

paddybyers:
This PR aims to retrofit in-place auth, and channel attach timers and retries, to the 0.8 branch, specifically to deal with race conditions in the existing `ConnectionManager` using the 0.8 style of reauthorisation.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/TokenAuth.java
+++ b/lib/src/main/java/io/ably/lib/http/TokenAuth.java
@@ -36,13 +36,13 @@ public class TokenAuth {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
+	public boolean authorise(AuthOptions options, TokenParams params, boolean force) throws AblyException {
 		Log.i("TokenAuth.authorise()", "");
 		if(tokenDetails != null) {
 			if(tokenDetails.expires == 0 || tokenValid(tokenDetails)) {
-				if(options == null || !options.force) {
+				if(options == null || !force) {
 					Log.i("TokenAuth.authorise()", "using cached token; expires = " + tokenDetails.expires);
-					return tokenDetails;
+					return false;
 				}
 			} else {
 				/* expired, so remove */
@@ -52,7 +52,7 @@ public class TokenAuth {
 		}
 		Log.i("TokenAuth.authorise()", "requesting new token");
 		setTokenDetails(auth.requestToken(params, options));
-		return tokenDetails;
+		return true;
 	}
 
 	public void clear() {
--- a/lib/src/main/java/io/ably/lib/realtime/Channel.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Channel.java
@@ -5,6 +5,7 @@ import io.ably.lib.http.PaginatedQuery;
 import io.ably.lib.http.Http.BodyHandler;
 import io.ably.lib.transport.ConnectionManager;
 import io.ably.lib.transport.ConnectionManager.QueuedMessage;
+import io.ably.lib.transport.Defaults;
 import io.ably.lib.types.*;
 import io.ably.lib.types.ProtocolMessage.Action;
 import io.ably.lib.types.ProtocolMessage.Flag;
@@ -16,6 +17,8 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Timer;
+import java.util.TimerTask;
 
 
 /**
@@ -105,6 +108,11 @@ public class Channel extends EventEmitter<ChannelState, ChannelStateListener> {
 	 * @throws AblyException
 	 */
 	public void attach(final CompletionListener listener) throws AblyException {
+		clearAttachTimers();
+		attachWithTimeout(listener);
+	}
+
+	private void attachImpl(final CompletionListener listener) throws AblyException {
 		Log.v(TAG, "attach(); channel = " + name);
 		/* check preconditions */
 		switch(state) {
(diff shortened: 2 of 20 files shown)


lib/src/main/java/io/ably/lib/http/TokenAuth.java
```diff
@@ -36,13 +36,13 @@ public void setTokenDetails(TokenDetails tokenDetails) {
 		this.encodedToken = Base64Coder.encodeString(tokenDetails.token).replace("=", "");
 	}
 
-	public TokenDetails authorise(AuthOptions options, TokenParams params) throws AblyException {
+	public boolean authorise(AuthOptions options, TokenParams params, boolean force) throws AblyException {
```
mattheworiordan: I am not sure I follow why `force` is added as an argument here when `force` is an attribute of the `options` being passed in and can be determined using the logic in `authorise`, https://github.com/ably/ably-java/blob/0a0f3ecb1d0810d26fceb4e2f9b80614649ed642/lib/src/main/java/io/ably/lib/rest/Auth.java#L420

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -215,6 +228,155 @@ public void sync() throws AblyException {
...
+					public void run() {
+						String errorMessage = String.format("Attach timed out for channel %s", name);
+						Log.v(TAG, errorMessage);
+						synchronized (Channel.this) {
+							if(attachTimer != currentAttachTimer)
+								return;
+							attachTimer = null;
+							if(state == ChannelState.attaching) {
+								setSuspended(new ErrorInfo(errorMessage, 91200));
+								reattachAfterTimeout();
```
mattheworiordan: This is not part of the 0.8 spec, once a channel moves to `detached` (which `setSuspended` confusingly does) then it remains in that state.

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -215,6 +228,155 @@ public void sync() throws AblyException {
...
+						}
+					}
+				}, Defaults.realtimeRequestTimeout);
+	}
+
+	/**
+	 * Must be called in suspended state. Wait for timeout specified in clientOptions, and then
+	 * try to attach the channel
+	 */
+	synchronized private void reattachAfterTimeout() {
```
mattheworiordan: As above, I don't think this code as anything to do with 0.8 or is necessary for inline auth

lib/src/main/java/io/ably/lib/realtime/Channel.java
```diff
@@ -215,6 +228,155 @@ public void sync() throws AblyException {
...
+		final Timer currentDetachTimer = new Timer();
+		attachTimer = currentDetachTimer;
+
+		try {
+			detachImpl(new CompletionListener() {
+				@Override
+				public void onSuccess() {
+					clearAttachTimers();
+					if (listener != null)
+						listener.onSuccess();
```
mattheworiordan: Completely aside from this, the listeners don't appear to conform to http://docs.ably.io/client-lib-development-guide/versions/v0.8/features/#RTE6 even though RTE6 only talks about events being emitted.  Should we add a new spec item that states all callbacks / listeners that raise an exception must be caught, logged, and execution should continue?

lib/src/main/java/io/ably/lib/realtime/Channel.java
paddybyers: But the whole point of this is that if an attach request never gets a response (because of failure of the transport either to send the request, or to process the response) then the attach is retried. If that doesn't happen it's a problem in 0.8 just as much as it is in 0.9. I don't think the idea of the 0.8 spec is to forbid it, and ably-js has been doing these retries for long before it was put into the 0.9 spec.

lib/src/main/java/io/ably/lib/realtime/Channel.java
paddybyers: Yes we should

lib/src/main/java/io/ably/lib/http/TokenAuth.java
paddybyers: In an earlier version of these changes I was calling it from elsewhere with `force` explicitly `true`, irrespective of `options.force`, but that later disappeared.

This is also called from `ensureValidAuth()` (https://github.com/ably/ably-java/blob/0a0f3ecb1d0810d26fceb4e2f9b80614649ed642/lib/src/main/java/io/ably/lib/rest/Auth.java#L383) with an explicit `false`, and it is possible for `Auth.authOptions` to have `force=true` because that isn't prohibited in the `authOptions` passed in the constructor. I guess [this](https://github.com/ably/ably-java/blob/0a0f3ecb1d0810d26fceb4e2f9b80614649ed642/lib/src/main/java/io/ably/lib/rest/Auth.java#L732) should have a `.storedValues()` as well.

paddybyers: Actually it is still called with `force=true`, with the `authOptions` otherwise being the stored defaults, from `Auth.renew()`, which is called in response to a server-initiated renewal.

> I guess this should have a .storedValues() as well.

This is wrong also, because that would remove any specified `queryTime` value.

So I don't propose any change for this.

lib/src/main/java/io/ably/lib/realtime/Channel.java
mattheworiordan: > I don't think the idea of the 0.8 spec is to forbid it, and ably-js has been doing these retries for long before it was put into the 0.9 spec.

There's a difference though as far as I can tell in that this moves the channel to `detached` with the error, and then retries.  If the channel remained in `attaching` and was retried that's a different thing.

Also, saying "I don't think the idea of the 0.8 spec is to forbid it" doesn't hold water for me.  The idea behind the spec is that there is predictable behaviour for customers.  The spec does not explicitly state all the possible things the library could do that it shouldn't do, it merely states what should happen so that developers understand expected behaviour.  If every dev working on the client libs added in behaviours they thought were useful but not part of the spec we'd have inconsistent libraries.

mattheworiordan:
Should this be closed now that 1.0 is released?

paddybyers:
Overtaken by events; this was a stopgap pending availability of 1.0, but that's now released.

Issue #262, opened 2016-12-29
Remove deprecated ProtocolMessage#connectionKey


mattheworiordan:
See https://github.com/ably/docs/commit/713122267fcb0af45fa68384109fe8d27f21e726

PR #271, opened 2017-01-08
Remove ProtocolMessage.connectionKey

paddybyers:
Fixes https://github.com/ably/ably-java/issues/262


mattheworiordan:
Weren't we also going to downgrade that warning?

paddybyers:
> Weren't we also going to downgrade that warning?

You're confusing it with https://github.com/ably/ably-java/pull/270 (also now merged)

PR #275, opened 2017-01-09
Request API support (RSC19)

paddybyers:
This is an initial implementation and tests for the generic request API.

There are issues with the API in Java that are not evident in dynamic languages, and also implementation issues in Java.

Firstly, our `PaginatedQuery<T>` primitive is parameterised by the type `T` of the members of a collection that are the subject of the request. That is, the request returns instances of `PaginatedResult<T>` that wraps arrays of type `T`. So for the `stats()` API, for example, we use a `PaginatedQuery<Stats>`, and we get back a page of results containing a `Stats[] items`.

This means that the current API is hard-wired to return arrays of results rather than single objects. To change that we would either have to have an `items` that is essentially untyped (ie it might be an array or an object, which is very unfriendly for the majority use-case) or we allow a result page to have both `T item` and `T[] items` members, with one or other being populated, which is also a bit ugly.

In this implementation I've opted instead for different methods on the `Rest` object: a `request()` method, which returns single objects, and a `paginatedRequest()`, which returns paginated responses, with arrays of members.

There are `Async` variants of each method.

The other implementation issue we have is that you can't get the msgpack library to return a `JsonElement`; and in fact the subset of the msgpack library that we include doesn't do reflection-based construction of types for deserialisation - so in fact there is no generic way to return an untyped and navigable structure for msgpack responses as there is for JSON. This implementation does not include support for msgpack; and the API is hard-coded to return `JsonElement`s; and it uses JSON even if `useBinaryProtocol=true`.

There are no tests for the `DELETE` method because there's nothing currently in the REST API that we can call.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttp.java
@@ -11,7 +11,6 @@ import java.util.concurrent.TimeoutException;
 
 import io.ably.lib.http.Http.RequestBody;
 import io.ably.lib.http.Http.ResponseHandler;
-import io.ably.lib.transport.Hosts;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.Callback;
 import io.ably.lib.types.ErrorInfo;
@@ -30,6 +29,19 @@ public class AsyncHttp extends ThreadPoolExecutor {
 	 */
 	public <T> Future<T> get(String path, Param[] headers, Param[] params, ResponseHandler<T> responseHandler, Callback<T> callback) {
 		return ablyHttpExecuteWithFallback(path, Http.GET, headers, params, null, responseHandler, callback);
+	}
+
+	/**
+	 * Async HTTP PUT for Ably host, with fallbacks
+	 * @param path
+	 * @param headers
+	 * @param params
+	 * @param requestBody
+	 * @param responseHandler
+	 * @param callback
+	 */
+	public <T> Future<T> put(String path, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler, Callback<T> callback) {
+		return ablyHttpExecuteWithFallback(path, Http.PUT, headers, params, requestBody, responseHandler, callback);
 	}
 
 	/**
@@ -55,6 +67,20 @@ public class AsyncHttp extends ThreadPoolExecutor {
 	 */
 	public <T> Future<T> del(String path, Param[] headers, Param[] params, ResponseHandler<T> responseHandler, Callback<T> callback) {
 		return ablyHttpExecuteWithFallback(path, Http.DELETE, headers, params, null, responseHandler, callback);
+	}
+
+	/**
+	 * Async HTTP request for Ably host, with fallbacks
+	 * @param path
+	 * @param method
+	 * @param headers
+	 * @param params
+	 * @param requestBody
+	 * @param responseHandler
+	 * @param callback
+	 */
+	public <T> Future<T> exec(String path, String method, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler, Callback<T> callback) {
+		return ablyHttpExecuteWithFallback(path, method, headers, params, requestBody, responseHandler, callback);
 	}
 
 	/**************************
@@ -352,7 +378,7 @@ public class AsyncHttp extends ThreadPoolExecutor {
 			final ResponseHandler<T> responseHandler,
 			final Callback<T> callback) {
 
-		AblyRequestWithFallback<T> request = new AblyRequestWithFallback<>(path, method, headers, null, requestBody, responseHandler, callback);
+		AblyRequestWithFallback<T> request = new AblyRequestWithFallback<>(path, method, headers, params, requestBody, responseHandler, callback);
 		execute(request);
 		return request;
 	}
@@ -379,7 +405,7 @@ public class AsyncHttp extends ThreadPoolExecutor {
 			final ResponseHandler<T> responseHandler,
 			final Callback<T> callback) {
 
-		AblyRequestWithRetry<T> request = new AblyRequestWithRetry<>(host, path, method, headers, null, requestBody, responseHandler, callback);
+		AblyRequestWithRetry<T> request = new AblyRequestWithRetry<>(host, path, method, headers, params, requestBody, responseHandler, callback);
 		execute(request);
 		return request;
 	}
--- a/lib/src/main/java/io/ably/lib/http/Http.java
+++ b/lib/src/main/java/io/ably/lib/http/Http.java
@@ -42,6 +42,7 @@ import io.ably.lib.util.Serialisation;
  */
 public class Http {
 	public static final String GET    = "GET";
+	public static final String PUT    = "PUT";
 	public static final String POST   = "POST";
 	public static final String DELETE = "DELETE";
 
@@ -245,6 +246,20 @@ public class Http {
 	}
 
 	/**
+	 * HTTP PUT for Ably host, with fallbacks
+	 * @param path
+	 * @param headers
+	 * @param params
+	 * @param requestBody
+	 * @param responseHandler
+	 * @return
+	 * @throws AblyException
+	 */
+	public <T> T put(String path, Param[] headers, Param[] params, RequestBody requestBody, ResponseHandler<T> responseHandler) throws AblyException {
+		return ablyHttpExecute(path, PUT, headers, params, requestBody, responseHandler);
+	}
+
+	/**
 	 * HTTP POST for Ably host, with fallbacks
 	 * @param path
 	 * @param headers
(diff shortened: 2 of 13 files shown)


lib/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -149,6 +152,51 @@ public void statsAsync(Param[] params, Callback<AsyncPaginatedResult<Stats>> cal
 	}
 
 	/**
+	 * Make a generic HTTP request for a collection of the given type
+	 * @param params query options: see Ably REST API documentation
+	 * for available options
+	 * @return a PaginatedResult of Stats records for the requested params
+	 * @throws AblyException
+	 */
+	public PaginatedResult<JsonElement> paginatedRequest(String method, String path, Param[] params, RequestBody body, Param[] headers) throws AblyException {
```
mattheworiordan: If we cannot support both a paginated array of results and a single JsonElement in a single method, I propose we instead name them according to the original method prefix i.e. `requestObject` and `requestPaginated` so that at least anyone expecting the `request` method would find them in intellisense.  I am not sure if `requestObject` is right, but I don't like the idea that `request` implements the spec somewhat and relies on the user discovering `paginatedRequest`. Instead I'd prefer that the user finds neither and discovers why there are two variations instead.

lib/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -149,6 +152,51 @@ public void statsAsync(Param[] params, Callback<AsyncPaginatedResult<Stats>> cal
 	}
 
 	/**
+	 * Make a generic HTTP request for a collection of the given type
+	 * @param params query options: see Ably REST API documentation
+	 * for available options
+	 * @return a PaginatedResult of Stats records for the requested params
```
mattheworiordan: Return type documentation is wrong here and in many other places below

lib/src/main/java/io/ably/lib/http/AsyncHttp.java
```diff
@@ -33,6 +32,19 @@
 	}
 
 	/**
+	 * Async HTTP PUT for Ably host, with fallbacks
+	 * @param path
```
mattheworiordan: I am not sure how Javadocs work but is there any value in these arguments being defined with no comments or types?  Is this not just additional noise in the code if it adds nothing for the end-user?

lib/src/main/java/io/ably/lib/http/Http.java
```diff
@@ -244,6 +245,20 @@ public String getUrlString(String url) throws AblyException {
 	}
 
 	/**
+	 * HTTP PUT for Ably host, with fallbacks
+	 * @param path
```
mattheworiordan: As above... same applies to exec below

lib/src/main/java/io/ably/lib/rest/AblyRest.java
mattheworiordan: Do you think it's obvious to a developer that `RequestBody` accepts `JsonRequestBody`?  Should we at least add it to the comments for this argument given the spec explicitly states the body should be a JSON object?

lib/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -149,6 +152,51 @@ public void statsAsync(Param[] params, Callback<AsyncPaginatedResult<Stats>> cal
...
+	 * @throws AblyException
+	 */
+	public PaginatedResult<JsonElement> paginatedRequest(String method, String path, Param[] params, RequestBody body, Param[] headers) throws AblyException {
+		headers = HttpUtils.mergeHeaders(HttpUtils.defaultAcceptHeaders(false), headers);
+		return new PaginatedQuery<JsonElement>(http, path, headers, params, HttpUtils.jsonArrayResponseHandler).exec(method);
```
mattheworiordan: As far as I can tell you're not actually sending the `body` with this request?  Same applies to `paginatedRequestAsync`

lib/src/test/java/io/ably/lib/test/rest/RestRequestTest.java
```diff
@@ -0,0 +1,287 @@
...
+				assertTrue("Verify id member is a matching channelName", items[i].getAsJsonObject().get("id").getAsString().startsWith(channelNamePrefix));
+			}
+			/* get next page */
+			channels = channels.next();
+			assertNotNull("Verify a result is returned", channels);
+			items = channels.items();
+			assertTrue("Verify one channels is returned", items.length == 1);
+			for(int i = 0; i < items.length; i++) {
+				assertTrue("Verify id member is a matching channelName", items[i].getAsJsonObject().get("id").getAsString().startsWith(channelNamePrefix));
+			}
```
mattheworiordan: Is it worth checking there are no more pages to at least know that the complete end-to-end paging works?

lib/src/main/java/io/ably/lib/rest/AblyRest.java
paddybyers: `requestItem()` and `requestItems()` ?

mattheworiordan: Would it not simply be better for us to have a PaginatedResponse where items is a `JsonStructure`, and therefore just have one `request` method?  I really dislike the idea that a user has to determine what type of response they are getting beforehand.

paddybyers: We support `JSONRequestBody` and `ByteArrayRequestBody` in the library; I've added comments to that effect

paddybyers: As discussed, we now have just `request()` and `requestAsync()`

lib/src/main/java/io/ably/lib/http/AsyncHttp.java
paddybyers: Comments added

paddybyers:
Last call for any comments on this.

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -57,6 +75,26 @@ public static String encodeParams(String path, Param[] params) {
...
 	}
 
+	public static Map<String, String> decodeParams(String query) {
+	    Map<String, String> params = new HashMap<String, String>();
+	    String[] pairs = query.split("&");
+        try {
+		    for (String pair : pairs) {
+		        int idx = pair.indexOf('=');
+				params.put(URLDecoder.decode(pair.substring(0, idx), "UTF-8"), URLDecoder.decode(pair.substring(idx + 1), "UTF-8"));
+			}
```
mattheworiordan: Some funky stuff going on here with indentation

lib/src/main/java/io/ably/lib/http/HttpUtils.java
```diff
@@ -71,4 +109,24 @@ public static String encodeURIComponent(String input) {
 		} catch (UnsupportedEncodingException e) {}
 		return null;
 	}
+
+	public static BodyHandler<JsonElement> jsonArrayResponseHandler = new BodyHandler<JsonElement>() {
+		@Override
+		public JsonElement[] handleResponseBody(String contentType, byte[] body) throws AblyException {
+			if(!"application/json".equals(contentType)) {
+				return null;
```
mattheworiordan: Silent failure?  I am not suggesting we raise an exception necessarily, but would it be at least logging a warning?

lib/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -149,6 +152,37 @@ public void statsAsync(Param[] params, Callback<AsyncPaginatedResult<Stats>> cal
...
+	 * of some type; this is to provide a forward compatibility path for new APIs.
+	 * @param method: the HTTP method to use (see constants in io.ably.lib.http.Http)
+	 * @param path: the path component of the resource URI
+	 * @param params (optional; may be null): any parameters to send with the request; see API-specific documentation
+	 * @param body (optional; may be null): an instance of RequestBody; either a JSONRequestBody or ByteArrayRequestBody
+	 * @param headers (optional; may be null): any additional headers to send; see API-specific documentation
+	 * @return a page of results, each represented as a JsonElement
+	 * @throws AblyException if it was not possible to complete the request, or an error response was received
+	 */
+	public PaginatedResult<JsonElement> request(String method, String path, Param[] params, RequestBody body, Param[] headers) throws AblyException {
```
mattheworiordan: The return type is a `HttpPaginatedResponse` not a `PaginatedResult`?

lib/src/test/java/io/ably/lib/test/rest/RestRequestTest.java
```diff
@@ -0,0 +1,554 @@
...
+			return;
+		}
+	}
+
+	/**
+	 * Verify error responses are indicated with an exception
+	 * Spec: RSC19e
+	 */
+	@Test
+	public void request_404() {
```
mattheworiordan: This is not correct according to the spec for two reasons:

* Exceptions are only raised when an HTTP response is not received, see http://docs.ably.io/client-lib-development-guide/features/#RSC19e
* Because you are not returning an `HttpPaginatedResponse`, you cannot check for a status of 404, or `success=false`.  I think unfortunately this PR does not address a lot of the requirements detailed in `HP1 - HP8` as a result, see http://docs.ably.io/client-lib-development-guide/features/#HP1

paddybyers:
This is all up to spec; closing

PR #277, opened 2017-01-10
RTP2 (except RTP2f) tests, RTP18c test

psolstice:


lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -302,6 +303,10 @@ public void enter_leave_simple() {
 			leaveComplete.waitFor();
 			assertTrue("Verify leave callback called on completion", leaveComplete.success);
 
+			try {
+				assertEquals("Verify item is removed from the presence map", client1Channel.presence.get(testClientId1, false).length, 0);
```
paddybyers: Why not test that the member has gone (ie based on `memberKey` instead of `clientId` ?

Edit: I guess you can't because it's private

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2062,7 +2067,8 @@ public void onPresenceMessage(PresenceMessage message) {
 				}
 			});
 
-			PresenceMessage[] testData = new PresenceMessage[] {
+			/* Test message newness criteria as described in RTP2b */
+			final PresenceMessage[] testData = new PresenceMessage[] {
```
paddybyers: We could add a message that has an earlier timestamp, but a later msgSerial, expecting it to be added, so we confirm that that serial takes precedence over the timestamp.

We could also add message serials that explicitly test that the comparison is numeric instead of lexicographic, eg "9" < "10"

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Actually I found `TP3h` spec and I'll move the `memberKey` function to `PresenceMessage` and make it public

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Agree for the first part, for the second part the test already does that ("500" < "1000")

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1949,81 +1954,130 @@ public void realtime_presence_get_throws_when_channel_failed() throws AblyExcept
...
+				assertEquals("Verify presence data is received by the server",
+						restChannel.presence.get(null).items().length, i==0 ? 1 : 2);
 
-			ably.connection.connectionManager.requestState(ConnectionState.connected);
-			channelWaiter.waitFor(ChannelState.attached);
+				/* In both cases we should have one leave message in the leaveMessages */
+				assertEquals("Verify exactly one LEAVE message was generated", leaveMessages.size(), 1);
+				PresenceMessage leaveMessage = leaveMessages.get(0);
+				assertTrue("Verify LEAVE message follows specs",
+						leaveMessage.clientId.equals(testClientId2) && leaveMessage.id == null &&
```
mattheworiordan: The spec states `The PresenceMessage published should contain the original attributes of the presence member with the action set to LEAVE, PresenceMessage#id set to null, and the timestamp set to the current time.`

You have not checked the action or the `data` attribute of the message.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2262,4 +2351,117 @@ public void onError(ErrorInfo reason) {
...
+						id = "4:2:2";
+					}}
+			};
+
+			channel.presence.subscribe(Action.leave, new Presence.PresenceListener() {
+				@Override
+				public void onPresenceMessage(PresenceMessage message) {
+					try {
+						if (message.clientId.equals("4") && message.action == Action.leave)
+							assertEquals("Verify LEAVE message is recorded as ABSENT during sync",
```
mattheworiordan: How do you even know this assertion ran?  This test could complete without this having run no?

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2262,4 +2351,117 @@ public void onError(ErrorInfo reason) {
...
+				channelSerial = "2:";
+				presence = testPresence3;
+			}});
+
+			PresenceMessage[] resultingPresence = null;
+			try {
+				assertEquals("Verify incomplete sync was discarded", channel.presence.get("1", false).length, 0);
+				assertEquals("Verify client with id==2 is in presence map", channel.presence.get("2", false).length, 1);
+				assertEquals("Verify client with id==3 is in presence map", channel.presence.get("3", false).length, 1);
+				assertEquals("Verify nothing else is in presence map", channel.presence.get(false).length, 2);
```
mattheworiordan: Would be good to confirm the expected order and number of presence events emitted please.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Action type is checked implicitly on subscribe:
`channel.presence.subscribe(Action.leave, ...)`
I'll add checks for the data though

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -179,80 +171,113 @@ public void unsubscribe() {
...
+
+							@Override
+							public void onError(ErrorInfo reason) {
+									/*
+									 * (RTP5c3) If any of the automatic ENTER presence messages published in RTP5c2
+									 * fail, then an UPDATE event should be emitted on the channel, with a reason
+									 * set to the ErrorInfo received from realtime
+									 */
+								Log.e(TAG, String.format("Cannot automatically re-enter channel %s", channel.name));
+								channel.emitUpdate(reason, false);
```
psolstice: This is an outdated line of code, not sure how you managed to comment on it. Now only client_id is missing

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -511,6 +546,14 @@ public void updatePresence(PresenceMessage msg, CompletionListener listener) thr
 		}
 	}
 
+	/**
+	 * (RTP13) Presence#syncComplete returns true if the initial SYNC operation has completed for
+	 * the members present on the channel
+	 */
+	public boolean syncComplete() {
```
paddybyers: Is this meant now to be a field or a method?

psolstice: In spec it is described as a method (returns). There is no problem changing it to field

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1949,81 +1943,131 @@ public void realtime_presence_get_throws_when_channel_failed() throws AblyExcept
 	 * Test if after reattach when returning from suspended mode client re-enters the channel with the same data
 	 * @throws AblyException
 	 *
-	 * Tests RTP17
+	 * Tests RTP17, RTP19, RTP19a, RTP5f
```
paddybyers: How does this test RTP17?

psolstice: It tests automatic re-enter feature described in RTP17. Test is done by injecting a member into internal presence map by tricking the client library that enter was confirmed by the server (though it really didn't) and then suspending and resuming the connection. Then server is queried using REST if there was a re-enter

lib/src/main/java/io/ably/lib/realtime/Presence.java
mattheworiordan: I don't think it's a method, RTP13 states `Presence#syncComplete returns true..` which does not mean it's a method, the spec is language agnostic.

More importantly though, in the IDL, it's defined as:
`syncComplete: Bool` which is an attribute

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
mattheworiordan: > Action type is checked implicitly on subscribe

That's not true.  That's the event that is emitted, and has nothing to do with the `PresenceMessage`.  It may do in your implementation, but that's not relevant to the tests as the library implementation can change in the future and decouple those things.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2262,4 +2341,539 @@ public void onError(ErrorInfo reason) {
...
+			opts.tokenDetails = token;
+			ably = new AblyRealtime(opts);
+
+			Channel channel = ably.channels.get(channelName);
+			channel.attach();
+
+			ChannelWaiter channelWaiter = new ChannelWaiter(channel);
+			channelWaiter.waitFor(ChannelState.attached);
+
+			final boolean[] flags = new boolean[] {
```
mattheworiordan: I don't propose you change this, but I don't follow why you would use cryptic flags over simply two varaibles `boolean callbackCalled` and `boolean success`?

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2262,4 +2341,539 @@ public void onError(ErrorInfo reason) {
...
+			assertFalse("Verify enter client failed", flags[1]);
+		} finally {
+			if (ably != null)
+				ably.close();
+		}
+	}
+
+	/**
+	 * Enter wrong client (mismatching one set in the token), check exception
+	 * Tests RTP8i
```
mattheworiordan: This is the wrong test, it stats `If the Ably service determines that the client is unidentified, a NACK is sent to the client resulting in an error`??  You are testing client ID mismatch instead of an unidentified client

mattheworiordan: BTW. You can probably keep this test as useful, but the actual requirement in the spec is not being tested as far as I can tell.

paddybyers:
> I don't think it's a method, RTP13 states Presence#syncComplete returns true

attributes/fields have values; functions return things. I definitely think it's open to misinterpretation :)

mattheworiordan:
> attributes/fields have values; functions return things. I definitely think it's open to misinterpretation :)

Agreed, wording could be better, but we have an IDL to avoid this no? That's why it exists.

paddybyers:
> we have an IDL to avoid this no?

yes of course, but when I read the description and the IDL I assumed that one or other hadn't been updated because they appeared to say different things.

mattheworiordan:
I will get the docs fixed soon https://github.com/ably/ably-java/pull/277#issuecomment-272860644

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2876,4 +2875,139 @@ public void onError(ErrorInfo reason) {
...
+							sentPresence.get(1).clientId.equals(testClientId2)
+			);
+
+			channel.detach();
+			new ChannelWaiter(channel).waitFor(ChannelState.detached);
+
+			try {
+				channel.presence.enterClient("testClient3");
+				fail("Presence.enterClient() shouldn't succeed in detached state");
+			} catch (AblyException e) {}
```
mattheworiordan: I really dislike these blanket exception catches because if another exception is thrown in future (i.e. due to a regression), then this test still passes.  Can you please check the error code and possibly the error message please to make sure it's the right exception being thrown?

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Implemented

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Added timestamps to the messages containing correct conenctionId to check newness comparison priority

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Done

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
psolstice: Done

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
mattheworiordan: Fair enough! 👍

PR #286, opened 2017-01-18
Implement Auth.clientId and all associated tests (except for presence-related)

paddybyers:


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -869,27 +935,74 @@ private static final String hmac(String text, String key) {
...
+				return;
+			}
+
+			if(!clientId.equals(msgClientId) && !clientId.equals(WILDCARD_CLIENTID)) {
+				throw AblyException.fromErrorInfo(new ErrorInfo("Incompatible clientId specified in message", 400, 40012));
+			}
+		}
+	}
+
+//	/**
```
mattheworiordan: You mean to commit these comments?

paddybyers: oops, no; forgot to implement something ...

PR #287, opened 2017-01-18
RTP11c, RTP11c, RTP11d implementation and tests

psolstice:


lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -26,37 +26,49 @@
...
+
+	/**
+	 * Get the presence state for this channel. Take Param[] array as an argument
+	 * @param params
+	 * @return
 	 * @throws AblyException
+	 * @throws InterruptedException
 	 */
-	public synchronized PresenceMessage[] get() throws AblyException {
+	public synchronized PresenceMessage[] get(Param... params) throws InterruptedException, AblyException {
```
paddybyers: Hmm, I'm not sure if we should expose this or we should wrap an `InterruptedException` in an `AblyException`; otherwise existing code will fail to compile.

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -777,7 +835,7 @@ synchronized boolean hasNewerItem(String key, PresenceMessage item) {
 		 * @return
 		 */
 		synchronized Collection<PresenceMessage> values() {
-			try { return values(false); } catch (InterruptedException e) { return null; }
+			try { return values(false); } catch (InterruptedException|AblyException e) { return null; }
```
paddybyers: Does this syntax work with the java version on all Android targets?

psolstice:
I meant RTP11b, RTP11c, RTP11d

lib/src/main/java/io/ably/lib/realtime/Presence.java
psolstice: According to spec `Presence.get()` without argument is a blocking call as well and can throw `InterruptedException` as well. So yes, compatibility will be broken in a lot of places. On the other hand `InterruptedException` is a well established standard in Java for blocking calls

lib/src/main/java/io/ably/lib/realtime/Presence.java
psolstice: From Java 7 only "try with resource" is not supported on all the targets, everything else is fine. Also I don't think it generates "special" byte code

lib/src/main/java/io/ably/lib/realtime/Presence.java
mattheworiordan: I concur with @paddybyers here, I would prefer to avoid this exceptional exception being thrown here when everywhere else (and certainly in either client libs) we try and stick with Ably exceptions.

psolstice: It might cause another problem. It is an error in Java to have `catch` block for the exception never thrown in the `try` block. So if we move to `InterruptedException` integrated into `AblyException` we need to modify methods `Presence.get(boolean wait)` and `Presence.get(String clientId, boolean wait)` too so they'll stop throwing `InterruptedException`. And this in turn will break code that uses old `Presence.get()` stuff.

psolstice: We can of course revert to old default in `Presence.get()` without arguments of wait=false. Then everything can still be done in compatible way

paddybyers: OK, so I missed the fact that the existing methods throw `InterruptedException`. But we're not consistent here, because we don't have those on other blocking methods such as the REST HTTP API (and the mainstream java blocking IO model doesn't routinely have `InterruptedException` on APIs that block for IO). You can argue that in the case of `presence.get()` we're not just blocking for IO, but we're performing a wait for a sequence of operations to complete, which includes IO; but i you look at higher level libraries (eg the Apache HttpClient) then you still don't have an `InterruptedException` on the blocking `HttpExecutor.execute()`; you just have `IOException`, when something to do with the IO failed, and `HttpException`, when the IO happened but there was a protocol error with the resulting data. We're wrapping both of those kinds of problems in an `AblyException`, for better or worse, and I don't think this `presence.get(waitForSync=true)` is in a different category.

As for the point that we're already declaring that as thrown, so existing callers that catch it will have to modify their code if we decide to remove them; I think it's ok. It is easier to remove a catch block from existing code than it is to add one and think about what to do with the exception.

Also, I'm not sure that `InterruptedException` is even semantically the correct thing if a thread blocked on the `presence.get()` is interrupted - I would say that `InterruptedIOException` is probably more correct (and this, being an `IOException`, is already wrapped as `AblyException`)

psolstice: `InterruptedException` is IMHO semantically correct, it is a result of client calling 'Thread.interrupt()`. You don't have an `InterruptedException` in HTTP code because you have to work with `HttpUrlConnection` and it has a lot of shortcomings.
Nonetheless, I changed it to `AblyException.fromThrowable(e)`. It breaks code too. First, the default for `Presence.get()` is now to wait for SYNC to complete. And secondly, several versions of `Presence.get()` are no longer throwing `InterruptedException`

PR #289, opened 2017-01-20
Added more pre-deploy tests

psolstice:


paddybyers:
LGTM but do you think our eventual aim should be to have a gradle task for this instead of a shell script?

psolstice:
Yes, of course. Publishing on JCenter can be done with gradle but this task is much easier to do in shell script

PR #291, opened 2017-01-20
Platform nw events

paddybyers:
@psolstice this is some work that I did to implement awareness of platform connectivity events; I've not tested any of it yet but pls review quickly and give your comments.


paddybyers:
Wrong branch, again

PR #292, opened 2017-01-20
Platform network status event handling

paddybyers:
@psolstice this is some work that I did to implement awareness of platform connectivity events; I've not tested any of it yet but pls review quickly and give your comments.

### DIFF
--- a/android/src/main/java/io/ably/lib/util/AndroidPlatform.java
+++ b/android/src/main/java/io/ably/lib/util/AndroidPlatform.java
@@ -1,0 +1,71 @@
+package io.ably.lib.util;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+import android.os.Build;
+
+import java.util.Properties;
+
+import io.ably.lib.util.Log;
+import io.ably.lib.util.Platform.PlatformEvent;
+
+public class AndroidPlatform extends Platform {
+	public static final String ANDROID_SDK_VERSION = "android-sdk-version";
+
+	public AndroidPlatform() {
+		name = "android";
+		properties.setProperty(ANDROID_SDK_VERSION, String.valueOf(Build.VERSION.SDK_INT));
+
+		try {
+			setContext((Context) Class.forName("android.app.ActivityThread").getMethod("currentApplication").invoke(null, (Object[]) null));
+		} catch(Throwable t) {
+			Log.e(TAG, "Unable to access application context", t);
+		}
+	}
+
+	public void setContext(Context ctx) {
+		this.context = ctx;
+		connectivityReceiver = new ConnectivityReceiver();
+	}
+
+	private class ConnectivityReceiver extends BroadcastReceiver {
+		ConnectivityReceiver() {
+			try {
+				/* get current network state */
+				ConnectivityManager cm = (ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE);
+				NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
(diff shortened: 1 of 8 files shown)


android/src/main/java/io/ably/lib/util/AndroidPlatform.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class AndroidPlatform extends Platform {
+	public static final String ANDROID_SDK_VERSION = "android-sdk-version";
+
+	public AndroidPlatform() {
+		name = "android";
+		properties.setProperty(ANDROID_SDK_VERSION, String.valueOf(Build.VERSION.SDK_INT));
+
+		try {
+			setContext((Context) Class.forName("android.app.ActivityThread").getMethod("currentApplication").invoke(null, (Object[]) null));
```
paddybyers: Yes, I know. I think probably there is no way to do this purely as a library - you need to get the app to add something to its manifest (eg Service belonging to the library) or the application needs to call something in the library to pass in a `Context`.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -925,6 +935,17 @@ protected boolean checkConnectivity() {
 		}
 	}
 
+	@Override
+	public void onPlatformEvent(PlatformEvent event, Object... args) {
+		switch(event) {
+		case NETWORK_UP:
+			connect();
+			break;
+		default:
```
mattheworiordan: Do we not move the connection to `DISCONNECTED` for `NETWORK_DOWN`?

paddybyers:
> is there no way we can test this by at least fabricating an event with onPlatformEvent?

Yes we can, that's the intention.

mattheworiordan:
>> is there no way we can test this by at least fabricating an event with onPlatformEvent?

> Yes we can, that's the intention.

Ok, look forward to seeing the tests then :)

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: > Do we not move the connection to DISCONNECTED for NETWORK_DOWN?

Android does that anyway, because the transport will be closed. The question is whether or not we accelerate the transition to `suspended` - but we will probably experience quite short interruptions in connectivity, so I'm not convinced there is a good use-case for that.

mattheworiordan: > Android does that anyway, because the transport will be closed

If you are sure, but I don't see the downside in catching the event as well.  I know in browsers we found that was not always the case.

> The question is whether or not we accelerate the transition to suspended - but we will probably experience quite short interruptions in connectivity, so I'm not convinced there is a good use-case for that.

I would agree...

mattheworiordan:
@paddybyers catching these events potentially significantly improves responsiveness for users on Android to changing network conditions.  Any chance we can add at least one simple test and merge in?

Leslieannewallace:
Help me i cant understand none of thia

Leslieannewallace:
Cant understad what this is an im worrie about my safety

android/src/main/java/io/ably/lib/util/AndroidPlatform.java
```diff
@@ -0,0 +1,71 @@
...
+			Log.e(TAG, "Unable to access application context", t);
+		}
+	}
+
+	public void setContext(Context ctx) {
+		this.context = ctx;
+		connectivityReceiver = new ConnectivityReceiver();
+	}
+
+	private class ConnectivityReceiver extends BroadcastReceiver {
```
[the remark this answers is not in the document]
paddybyers: Relevant: https://medium.com/@andretietz/auto-initialize-your-android-library-2349daf06920

paddybyers:
Closing as superseded

PR #294, opened 2017-01-20
Fixes and tests for RTP8i, RTP8f

psolstice:


lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -3013,4 +3013,57 @@ public void onPresenceMessage(PresenceMessage messages) {
...
+			channel.presence.enter(null, completionWaiter);
+
+			ably.connection.connect();
+
+			completionWaiter.waitFor(1);
+			assertFalse("Verify enter() failed", completionWaiter.success);
+
+			/* Now clientId is known to be "*" and subsequent enter() should fail immediately */
+			try {
+				channel.presence.enter(null, null);
```
mattheworiordan: Given how easy it is to add `update` and `leave` as well, why not add them as well to this test.

psolstice: This line is supposed to be the test for RTP8f. `clientId` is '*' at the moment and enter() fails as expected

mattheworiordan: Yes, I know that, but see https://docs.ably.io/client-lib-development-guide/features/#RTP9e , https://docs.ably.io/client-lib-development-guide/features/#RTP10e and https://docs.ably.io/client-lib-development-guide/features/#RTP15a.  For efficiency, this test could cover this behaviour for all of those methods so why not do that now?

psolstice: sure, later today

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -425,7 +425,12 @@ public void enterClient(String clientId, Object data) throws AblyException {
 	 * @throws AblyException
 	 */
 	public void enterClient(String clientId, Object data, CompletionListener listener) throws AblyException {
-		Log.v(TAG, "enterClient(); channel = " + channel.name + "; clientId = " + clientId);
+		Log.v(TAG, "enterClient(); channel = " + channel.name + "; clientId = " + (clientId != null ? clientId : "(null)"));
+		if (clientId == null && "*".equals(channel.ably.auth.clientId)) {
```
paddybyers: There is a static final for "*"

psolstice: Yes, in Auth but it's private

psolstice: nonetheless it looks better with static final, I'll make it public

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -3013,4 +3013,73 @@ public void onPresenceMessage(PresenceMessage messages) {
...
+
+			/* enter without attaching first */
+			Channel channel = ably.channels.get("enter_before_clientid_is_known"+testParams.name);
+			CompletionWaiter completionWaiter = new CompletionWaiter();
+			channel.presence.enter(null, completionWaiter);
+
+			ably.connection.connect();
+
+			completionWaiter.waitFor(1);
+			assertFalse("Verify enter() failed", completionWaiter.success);
```
paddybyers: I think the test should get the error from the CompletionWaiter and verify that it has the expected error code

paddybyers:
@psolstice I added https://github.com/ably/ably-java/pull/294/commits/2076a2913febe23994695f06fc3dfbdf72acdef8; pls review

PR #296, opened 2017-01-20
RSA4c1: Wrap auth callback err

paddybyers:
Implement and test RSA4c1

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -510,7 +510,14 @@ public class Auth {
 			tokenDetails = authOptions.tokenDetails;
 			setTokenDetails(tokenDetails);
 		} else {
-			tokenDetails = assertValidToken(params, options, true);
+			try {
+				tokenDetails = assertValidToken(params, options, true);
+			} catch (AblyException e) {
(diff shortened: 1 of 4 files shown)


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -510,7 +510,14 @@ public TokenDetails authorize(TokenParams params, AuthOptions options) throws Ab
 			tokenDetails = authOptions.tokenDetails;
 			setTokenDetails(tokenDetails);
 		} else {
-			tokenDetails = assertValidToken(params, options, true);
+			try {
+				tokenDetails = assertValidToken(params, options, true);
+			} catch (AblyException e) {
+				/* Give AblyRealtime a chance to update its state and emit an event according to RSA4c */
+				if (e.errorInfo.code == 80019)
```
paddybyers: I don't see why `onAuthError()` is only called if there was an error from `authCallback` or `authUrl`; there are other reasons for an `authorise()` failing; eg a token request failed because of a rate limit. The `ConnectionManager` will need to be notified of that problem as well.

psolstice: I thought 80019 code covered this case as well. Ok, I'll remove this if

PR #298, opened 2017-01-23
RTP6 fixes and tests

psolstice:


lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -173,6 +197,21 @@ public void unsubscribe() {
...
 	 */
 
+	/**
+	 * Implicitly attach channel on subscribe. Throw exception if channel is in failed state
+	 * @param completionListener
+	 * @throws AblyException
+	 */
+	private void implicitAttachOnSubscribe(CompletionListener completionListener) throws AblyException {
+		if (channel.state == ChannelState.failed) {
+			String erroString = String.format("Channel %s: subscribe in FAILED channel state", channel.name);
```
paddybyers: typo: erroString

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1683,20 +1685,35 @@ public void realtime_presence_attach_implicit_subscribe_fail() throws AblyExcept
 		AblyRealtime ably = null;
```
paddybyers: Is there a way in this test to confirm "if the channel is in or enters the FAILED state before the operation succeeds, **it will result in the listener not being registered**"

psolstice: I'm a bit confused because it's clearly different here:
https://github.com/ably/docs/pull/260/commits/91c700a4f32d0a1139676cb0a7648cb899f35143

paddybyers: Right, I'm sorry, and I was thinking of that requirement :)

What I want is something that checks whether or not the listener was added, in the failure case. Then pass or fail according to what the spec says :)

paddybyers:
LGTM

PR #300, opened 2017-01-26
RTL15: moved Channel.attachSerial to Channel.properties.attachSerial

psolstice:


lib/src/main/java/io/ably/lib/realtime/ChannelProperties.java
```diff
@@ -0,0 +1,15 @@
+package io.ably.lib.realtime;
```
paddybyers: I think I would have put this into `types`, in case in the future we use it also for Rest channels

paddybyers:
Thanks

PR #301, opened 2017-01-26
RTN16b, RTN16c: added recoveryKey to Connection

psolstice:


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -496,8 +496,11 @@ public void onMessage(ITransport transport, ProtocolMessage message) throws Ably
 	}
 
 	private void onChannelMessage(ProtocolMessage message) {
-		if(message.connectionSerial != null)
+		if(message.connectionSerial != null) {
 			connection.serial = message.connectionSerial.longValue();
+			if (connection.key != null)
+				connection.recoveryKey = String.format("%s:%d", connection.key, message.connectionSerial);
```
paddybyers: Since this happens on every message, I'm not happy with the overhead of using `String.format()` when a simple expression `a + ':' + b` would do

paddybyers:
Thanks

PR #302, opened 2017-01-26
RSA8c1b: added authMethod to AuthOptions, implemented POST for authUrl

psolstice:


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -613,7 +624,20 @@ public Object handleResponse(Response response, ErrorInfo error) throws AblyExce
...
+				};
+
+				if (Http.POST.equals(tokenOptions.authMethod)) {
+					URL postUrl = null;
+					try {
+						postUrl = new URL(tokenOptions.authUrl);
+					} catch (MalformedURLException e) {
+						throw AblyException.fromErrorInfo(new ErrorInfo("Malformed token URL", 40140));
+					}
+					authUrlResponse = ably.http.httpExecute(postUrl, Http.POST, tokenOptions.authHeaders, new Http.FormRequestBody(requestParams), responseHandler);
```
paddybyers: These two cases now seem needlessly different. EIther both use `httpExecute` or we add an http helper for `POST` like `getUri()`. Probably the first is better.

psolstice: Second method is clearer because we don't need to build GET URL in Auth code

paddybyers:
Thanks

PR #303, opened 2017-01-27
RSE2: Crypto.generateRandomKey() implementation and test

psolstice:


mattheworiordan:
Docs need to be updated when this is merged in?  @paddybyers any reason this is not merged in then?

PR #308, opened 2017-02-14
Android push implementation.

tcard:
# Installation

1. Clone the `push` branch: `git clone -b push git@github.com:ably/ably-java.git`
2. Follow [the build instructions](https://github.com/ably/ably-java#building) to get either an Android AAR or a Java JAR.
3. Link the generated archive file into your project.
4. If you're using Android, the generated AAR won't include Ably's dependencies. You should add this to your own build.gradle files:

```
apply from: '/path/to/your/clone/of/ably-java/dependencies.gradle'
apply from: '/path/to/your/clone/of/ably-java/android/dependencies.gradle'
```


android/build.gradle
```diff
@@ -61,11 +61,16 @@ tasks.withType(com.android.build.gradle.internal.tasks.AndroidTestTask) { task -
...
+    compile 'io.azam.ulidj:ulidj:1.0.0'
+
     androidTestCompile 'com.android.support.test:runner:0.5'
     androidTestCompile 'com.android.support.test:rules:0.5'
     androidTestCompile "com.crittercism.dexmaker:dexmaker:1.4"
     androidTestCompile "com.crittercism.dexmaker:dexmaker-dx:1.4"
     androidTestCompile "com.crittercism.dexmaker:dexmaker-mockito:1.4"
+
+    compile 'com.google.android.gms:play-services-gcm:10.0.1'
+    compile 'com.google.firebase:firebase-messaging:10.0.1'
```
mattheworiordan: Is Firebase needed or can we avoid that dependency by using native APIs?

android/src/main/java/io/ably/lib/rest/Push.java
```diff
@@ -0,0 +1,247 @@
...
+        bodyJson.add("recipient", recipientJson);
+        for (Map.Entry<String, JsonElement> entry : payload.entrySet()) {
+            bodyJson.add(entry.getKey(), entry.getValue());
+        }
+        return rest.http.requestBodyFromGson(bodyJson);
+    }
+
+    public void activate(Context context) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        if (prefs.getBoolean(SharedPrefKeys.ACTIVATED, false)) {
```
mattheworiordan: Out of interest, what if a user disables push notifications for their device. Does the token get invalidated, or are push notifications simply disabled?  The reason I ask is that if I allow push, then disable push for my app, and then the dev tries to request push access again, will `prefs.getBoolean(SharedPrefKeys.ACTIVATED, false)` be false?  Is there any way for a dev to know if push becomes disabled or not?

android/src/main/java/io/ably/lib/rest/Push.java
```diff
@@ -0,0 +1,247 @@
...
+                }
+                @Override
+                public void onError(ErrorInfo reason) {
+                    new ActivateCallback(context, "PUSH_ACTIVATED").onError(reason);
+                }
+            });
+        } else {
+            rest.device = device;
+            Http.RequestBody body = rest.http.requestBodyFromGson(device.pushMetadataJsonObject());
+            rest.asyncHttp.put("/push/deviceRegistrations/" + device.id, HttpUtils.defaultAcceptHeaders(rest.options.useBinaryProtocol), null, body, null, new ActivateCallback(context, "PUSH_RENEWED_TOKEN"));
```
mattheworiordan: Do we ever retry a failed request i.e. if the HTTP request fails, is it then up to the dev to pick up the pieces, or do we retry updating Ably perhaps next time the lib is connected?

android/src/main/java/io/ably/lib/rest/Push.java
```diff
@@ -0,0 +1,247 @@
...
+    public static class RegistrationToken {
+        public RegistrationTokenType type;
+        public String token;
+
+        private RegistrationToken(RegistrationTokenType type, String token) {
+            this.type = type;
+            this.token = token;
+        }
+
+        public static RegistrationToken persisted(Context context) {
```
mattheworiordan: I am not quite sure I understand why this method is named `persisted`.  Mind explaining?

android/build.gradle
tcard: It is needed. There's nothing available to the public behind it.

android/src/main/java/io/ably/lib/rest/Push.java
tcard: Push.activate is about Ably, not about F/GCM. Push.activate/deactivate registers/unregisters _with Ably_. So the token isn't invalidated in F/GCM's servers or anything like that, it's just unregistered on Ably's so that we don't send notifications to it anymore.

> The reason I ask is that if I allow push, then disable push for my app, and then the dev tries to request push access again, will prefs.getBoolean(SharedPrefKeys.ACTIVATED, false) be false?

It will be false once deactivation completes, then true again once re-activation completes.

> Is there any way for a dev to know if push becomes disabled or not?

Not directly, but why would he want to do that? It can only become disabled if she disables it.

android/src/main/java/io/ably/lib/rest/Push.java
tcard: It returns the persisted registration token. `getPersisted` maybe?

android/src/main/java/io/ably/lib/rest/Push.java
tcard: Yeah, that's something I've considered here but I think it's preferable to just call the callback (the BroadcastReceivers in this case) with an error, just like we do with all other requests. What can we really do if all fallback hosts, etc. fail? It's probably the user who needs to fix something (auth, etc.).

tcard:
So, while implementing deactivate I realized the current implementation of the activation process it 1. an unclear mess and 2. broken. I'm refactoring it to follow a well-defined, exhaustive, event-based state machine that will be included in the spec. So don't look at that part just yet.

android/build.gradle
mattheworiordan: OK

android/src/main/java/io/ably/lib/rest/Push.java
mattheworiordan: Nope. It is common to detect that push notifications are not enabled and prompt the user to enable them. Using Ably, how would a developer know if push notifications are still enabled for this app?

android/src/main/java/io/ably/lib/rest/Push.java
mattheworiordan: > What can we really do if all fallback hosts, etc. fail? It's probably the user who needs to fix something (auth, etc.).

Wait until an internet connection becomes available and try again?

android/src/main/java/io/ably/lib/rest/Push.java
mattheworiordan: Yes, perhaps, I was a little confused what it did to be honest.

android/src/main/java/io/ably/lib/rest/Push.java
tcard: If you mean if push notifications are enabled on Android's part, Android apps don't need to ask for permissions to publish notifications as iOS apps do. So it's always enabled. The user can't disable them in that sense.

If you mean if the device is registered for push notifications with Ably, then... yeah, there's no way to check that. So you want a way to actively check with Ably's servers if the device is registered?

tcard:
OK, the state machine thing is done now. I'll write down a spec for it so that @ricardopereira can make sure iOS does the exact same thing.

android/src/main/java/io/ably/lib/rest/Push.java
tcard: Done at 0bd203b36c4d718fff01dcd5cb423287ece94fcd.

tcard:
So there it goes the admin API and other missing pieces. I think that's all. Some of it I haven't even tested once, so expect bug fixes down the line.

android/maven.gradle
```diff
@@ -39,10 +39,17 @@ uploadArchives {
...
         pom.whenConfigured {
             p -> p.dependencies = p.dependencies.findAll {
+                // Exclude dependency on lib subproject.
                 dep -> dep.artifactId != "lib"
+            }.findAll {
+                // Exclude Google services since we don't want to impose a particular
+                // version on users. Ideally we would specify a version range,
+                // but the Google services Gradle plugin doesn't seem to
+                // support that.
+                // TODO: Make sure this works when installing from Maven!
```
paddybyers: Can this TODO be resolved before we merge?

Chris59160:
I tried to follow your instructions...but the Android AAR can't get generated.

I can generate the jars, but the AAR can't be. I keep getting 
"* What went wrong:
Project 'android' not found in root project 'ably-java'."

Even though the android project is indeed in my ably-java folder...I am not sure why it keeps doing that, but it is stopping me from using the Push Notifications :/

Any idea ?

paddybyers:
> Even though the android project is indeed in my ably-java folder...I am not sure why it keeps doing that, but it is stopping me from using the Push Notifications :/
> 
> Any idea ?

Do you have the Android SDK installed? See https://github.com/ably/ably-java/blob/master/settings.gradle#L4

Chris59160:
Hello paddybyers,

haha of course I have the Android SDK installed, I have been developing Android app for years ;-)

To be fair, I have switched to using FCM myself instead of going through Ably for Push Notifications. I will keep ably for real time, until the team has included the Push Notification in the default gradle project.

Thanks for the answer and taking time to read my problem though :-)

paddybyers:
> haha of course I have the Android SDK installed, I have been developing Android app for years ;-)

Well I thought I'd ask :) If `ANDROID_HOME` is defined I can't think of any reason why it wouldn't know about the `android` project. `gradle tasks` lists only the `java` tasks?

paddybyers:
This was merged independently.

Issue #306, opened 2017-01-31
Android System.out: (ERROR): io.ably.lib.transport.WebSocketTransport: No activity for 25000ms, closing connection


abhimanyuraj:

SimonWoolf:
Another customer reported this
their (mildly redacted) code: https://gist.github.com/SimonWoolf/0d98988cd9395d6ef5fbaf4b971d31ae
(mildly redacted) verbose logs: https://gist.github.com/SimonWoolf/bf7024814540d8c19b2590cfcec56650

The lib sends a CLOSE and receives a CLOSED, but the websocket seems to stay open until the 25s activity timeout; which is odd, as realtime definitely disconnects websockets after it sends a CLOSED.
```
(VERBOSE): io.ably.lib.transport.ConnectionManager: onMessage( {"action":8,"flags":0,"count":0,"timestamp":0}
(VERBOSE): io.ably.lib.transport.ConnectionManager: setState(): setting closed
(VERBOSE): io.ably.lib.transport.ConnectionManager: notifyState(): notifying disconnected; id = null
(VERBOSE): io.ably.lib.transport.WebSocketTransport: checkActivity: ok
(ERROR): io.ably.lib.transport.WebSocketTransport: No activity for 25000ms, closing connection
```

(also odd that the connection state would return to `disconnected` after going into `closed`).

That's on 0.8.11, but customer gave a similar log for 1.0.

emilm:
Yes it will remain open forever. And the WebSocketTransport will report no activity etc forever. Seems like that thread maintaining activity just keeps on living.

paddybyers:
I've added https://github.com/ably/ably-java/commit/6899a0240ef18a45edae3e2a3e46f29728c6f725 which eliminates the commonly-seen error message "No activity for 25000ms, closing connection", where the timer was left running even after the WsClient was closed.

I've also added https://github.com/ably/ably-java/commit/934b56282a1103d25f1aaf3ecbbf36129707b788 which has some tests to verify that the connectionmanager thread is exiting cleanly in various situations.

@emilm I tried to reproduce what you're doing in your code but I've not been able to reproduce the situation in which the connectionmanager fails to exit. Are you able to provide any additional information which would help me reproduce it? Or a code sample?

Thanks

emilm:
Hello! All I did was:

`public void open() {
    ExecutorService executor = Executors.newSingleThreadExecutor();
    executor.submit(() -> {
        try {
[...]

PR #315, opened 2017-03-11
ConnectionManager fix attempt

emilm:
An attempt to fix issue #306

The process exits now, seems to work in a short case I tested. it requires some testing, but it should work.

Also I changed the if / else logics so check if what I did there is correct as well, or if you really are supposed to have if stateChange.state != state.state on top.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -600,7 +600,8 @@ public class ConnectionManager implements Runnable, ConnectListener {
 		boolean creating = false;
 		synchronized(this) {
 			if(mgrThread == null) {
-				mgrThread = new Thread(this);
+                mgrThread = new Thread(this, "Connection Manager");
+                mgrThread.setDaemon(true); // So it exists properly
 				state = states.get(ConnectionState.initialized);
 				creating = true;
 			}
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Thanks for the contribution @emilm. I'll get one of the maintainers to review.

emilm:
I hope this is looked into ASAP. I am on a time constraint and I can't have a deamon not shutting down when I signal it to quit. The process will be hanging forever and cause a lot of headache unless this is fixed.

paddybyers:
Hi Emil,

I am looking now

emilm:
Just setting the isDaemon(true) would solve the problem alone, but I guess it's not a proper/clean solution. :)

paddybyers:
> Just setting the isDaemon(true) would solve the problem alone, but I guess it's not a proper/clean solution. :)

Yes, but it is designed to exit cleanly, without being forcibly destroyed. There are environments were you don't have permission to call `setDaemon()`; and if the thread exits abruptly without sending a `CLOSE` message to the server, then it leaves some connection resources on the server for a period of time, which we prefer to avoid.

I will let you know when I have made some progress with it.

paddybyers:
I'm closing this, as we confirmed the other fix resolves the issue (https://github.com/ably/ably-java/commit/6899a0240ef18a45edae3e2a3e46f29728c6f725)

Issue #105, opened 2016-02-26
New Crypto Spec


mattheworiordan:
Please see https://github.com/ably/docs/pull/89

We need to ensure this library complies with the changes described in the PR

mattheworiordan:
@trenouf whilst you are in the Crypto stuff, I know what we don't actually adhere to the Crypto spec correctly as there are missing methods / objects according to [the 0.8 IDL](http://docs.ably.io/client-lib-development-guide/versions/v0.8/features/#idl).

We have had one customer complain about this already and thought that the Java lib was unstable as a result. 

Could you specifically check that `ChannelOptions`, `CipherParams` and `Crypto ` match the IDL?  If there are changes to be made, I suspect there will be no underlying functionality change fortunately.

trenouf:
Will do.

mattheworiordan:
Thanks

trenouf:
ChannelOptions spec points to the java one as a reference!
However the field that the spec thinks should be called `cipher` is instead called `cipherParams`, and there is another private field called `cipher`. That could cause confusion.

CipherParams:
Instead of a binary field `key`, it has a SecretKeySpec field `keySpec`. I guess we could implement the binary field `key` with some code to notice when it has changed and update the `keySpec`.
It does not have a `mode` field.
It has a `ivSpec` public field that is not in the spec.

Crypto:
Missing the `generateRandomKey` static method.

paddybyers:
> Instead of a binary field key, it has a SecretKeySpec field keySpec. I guess we could implement the binary field key with some code to notice when it has changed and update the keySpec.

Is the `key` supposed to be mutable? That certainly wasn't the original intention.

mattheworiordan:
From what I can tell:

* [getDefaultParams](https://github.com/ably/ably-java/blob/master/lib/src/main/java/io/ably/lib/util/Crypto.java#L74-L89) does not conform, see http://docs.ably.io/client-lib-development-guide/features/#RSE1b, it should take a Hashmap or language equivalent so that  any subset of CipherParams fields that includes a key can be provided.  That needs to be fixed.
* I agree it should be `key` not `keySpec`. On @paddybyers's point I am not sure it needs to be mutable, that remains unspecified in the spec.  Why can it not simply be provided as binary and remain as binary as an attribute?
[...]

PR #329, opened 2017-06-21
Crypto default 256 bit length like all other libraries

mattheworiordan:
I am not sure I understand why Java is the only lib with a default key length different to the other libs, so I have changed this to 256 bits like the others and according to the spec.


mattheworiordan:
I assume this should have been covered in https://github.com/ably/ably-java/issues/105 but was not

Issue #166, opened 2016-09-27
Auth header included in HTTP requests


mattheworiordan:
I am seeing a lot of this in the logs from the Java client:

```
16-09-27T11:53:14.235Z lvl=info msg="POST /channels/public:foo/messages" reqid=4e5cee24-def5-4383-bc97-aa2dd7f0f4ef host=52.31.254.190 proto=HTTP/1.1 referer= ua=Java/1.8.0_11 backend=rest-https secs=0.002 status=401 size=166
t=2016-09-27T11:53:14.247Z lvl=info msg="POST /channels/public:foo/messages" reqid=81688bd7-12da-4753-b572-d3bb47b9d17f host=52.31.254.190 proto=HTTP/1.1 referer= ua=Java/1.8.0_11 backend=rest-https app=27d-sg channel=foo:public:foo secs=0.007 status=204 size=0
```

It looks to me like it's plausible that the Java client does not send the Basic authentication headers in the request but instead waits for a challenge each time and resubmitting the request.  Given it is never valid to send a request without either a Token or Basic Auth header, we should look into this.

mattheworiordan:
@trenouf any idea how the underlying HTTP lib works?

trenouf:
No, sorry, never looked at it.

> @trenouf <https://github.com/trenouf> any idea how the underlying HTTP lib works?

mattheworiordan:
Ok, might be one for @paddybyers to comment on later when he's back then.  Seems odd that every Java request is in fact two HTTP requests

paddybyers:
> Seems odd that every Java request is in fact two HTTP requests

Well it's supposed to be doing non-preemptive authentication, which is generally preferred practice, and sending credentials in response to the www-authenticate challenge. This should only happen once; the credentials should be cached by the HTTP client and always re-sent for subsequent requests by the same client to the same origin. If that's not happening then it's a bug.

mattheworiordan:
> Well it's supposed to be doing non-preemptive authentication, which is generally preferred practice, and sending credentials in response to the www-authenticate challenge

Why is that preferred practice when we know 100% of the time we'll need to authenticate?

> his should only happen once; the credentials should be cached by the HTTP client and always re-sent for subsequent requests by the same client to the same origin

I don't think we should assume that REST libraries are stateful.

mattheworiordan:
@paddybyers any thoughts, not sure the advantage of challenge response ever when we know we need to authenticate?

paddybyers:
[...]

PR #339, opened 2017-08-09
Preemptive HTTP authentication

paddybyers:
Expose an argument to http ops indicating whether or not auth will be required for an Ably endpoint; use this to ensure that all relevant requests are preemptively authenticated


paddybyers:
Fixes: https://github.com/ably/ably-java/issues/166

PR #347, opened 2017-08-31
Add ability to generalize over a HTTP request being async or not.

tcard:
This was something I had done in the push, but is independent of it.

Move AsyncHttp to CallbackfulHttp, and instead of inheriting from
ThreadPoolExecutor, take any Executor as a parameter. Then, have
both AsyncHttp and SyncHttp inheriting from it. AsyncHttp injects a
ThreadPoolExecutor, as before, while SyncHttp injects an Executor
that just executes Runnables right away, in the current thread.
Thus, SyncHttp exposes an “asynchronous” (callbackful) interface
while behaving synchronously.

By using CallbackfulHttp, we can abstract over whether an operation
is executed synchronously or asynchronously.

To make this more user-friendly, AnySyncHttp wraps both a SyncHttp
and an AsyncHttp, and lets you specify how to execute (class Execute) a
request (class Request) and then getting its result synchronously or
asynchronously (in another thread), as a method call result or passed
to a callback, with the `sync` and `async` methods respectively.

This design should make the Http class obsolete, once/if all the code
that uses it is ported. It could also be left as a proxy to SyncHttp.

(I did this after I renamed something in a `foo` method but forgot to
do the same in a `fooAsync` method, causing a bug that was reported by
a user.)

As an example, I've refactored Channel.publish(Message[]). Luckily, I've found an instance of the class of errors this is supposed to fix: RTL6g3 was implemented for publish, but not for publishAsync. Now, as they share their code, is implemented by both.


lib/src/main/java/io/ably/lib/http/AnySyncHttp.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.ably.lib.types.AblyException;
+import io.ably.lib.types.Callback;
+import io.ably.lib.types.ErrorInfo;
+
+/**
+ * Created by tcard on 28/04/2017.
+ */
+
+public class AnySyncHttp {
```
mattheworiordan: Should we perhaps add a brief comment as to the purpose of this class for future us?

lib/src/main/java/io/ably/lib/http/CallbackfulHttp.java
```diff
@@ -0,0 +1,423 @@
...
+import io.ably.lib.types.Callback;
+import io.ably.lib.types.ErrorInfo;
+import io.ably.lib.types.Param;
+import io.ably.lib.util.Log;
+
+/**
+ * Created by tcard on 31/8/17.
+ */
+
+public class CallbackfulHttp<Executor extends java.util.concurrent.Executor> {
```
mattheworiordan: What is `CallbackfulHttp` i.e. I am not clear on what the `ful` after `Callback` implies.  Perhaps a high level comment on what this does is needed.

paddybyers:
Overall I like this as a way of rationalising the duplication we had but my reservations at the moment are:

1) names .. I don't like `CallbackfulHttp`, `AnySyncHttp`, `publishAnySync` etc - I'm trying renaming a few things to see how it looks.

2) I would really like to see the migration from `Http` to `SyncHttp` completed  and `AblyRest.http` removed. That can be part of a follow-up PR but we should do it now instead of later.

3) I don't like the passing of results back in array members in https://github.com/ably/ably-java/blob/e60c31ca3e8ced94dc6b82e33be6a1f794c4067a/lib/src/main/java/io/ably/lib/http/AnySyncHttp.java#L30, and especially the fact that it's an untyped `Object` and there need to be unchecked casts. There's surely enough generics to avoid this.

I'll be back with a few naming suggestions.

tcard:
If we get rid of `Http`, `CallbackfulHttp` can be just `Http` (or `AbstractHttp` for extra Javaness). `CallbackfulHttp` is named that way as opposed to "callbackless" `Http`, because that's the difference between them.

I also don't love `AnySyncHttp`. I've thought about `HttpProxy`, because it's just a GoF proxy for `AsyncHttp` and `SyncHttp`, but it could be confused with an HTTP proxy in the network sense. `HttpRequester` maybe?

tcard:
f782a9e addresses @paddybyers 's point 3.

lib/src/main/java/io/ably/lib/http/AnySyncHttp.java
```diff
@@ -0,0 +1,65 @@
...
+    private final AsyncHttp asyncHttp;
+    private final SyncHttp syncHttp;
+
+    public AnySyncHttp(AsyncHttp asyncHttp, SyncHttp syncHttp) {
+        this.asyncHttp = asyncHttp;
+        this.syncHttp = syncHttp;
+    }
+
+    public static class Request<Result> {
```
paddybyers: I don't see why this is `static` and yet it has a `anySyncHttp` member. I think you can remove that member and remove the `static` declaration.

lib/src/main/java/io/ably/lib/http/AnySyncHttp.java
```diff
@@ -0,0 +1,69 @@
...
+        private final AnySyncHttp anySyncHttp;
+        private final Execute<Result> execute;
+
+        Request(AnySyncHttp anySyncHttp, Execute<Result> execute) {
+            this.anySyncHttp = anySyncHttp;
+            this.execute = execute;
+        }
+        
+        public Result sync() throws AblyException {
+            final SyncExecuteResult<Result> result = new SyncExecuteResult<>();
```
paddybyers: Yes, I was looking at just having `syncResult` and `syncErr` members in the `Result` type, instead of a new type to hold them, but what you have is a bit cleaner.

lib/src/main/java/io/ably/lib/http/AnySyncHttp.java
tcard: Done at 08f3c98.

paddybyers:
OK, I think we're heading towards a messy state where the existing `Http` class is a mixture of various configuration options, some ancillary classes (`RequestBody` etc) and some of the request primitives that are used elsewhere in the API. I think we should be refactoring that into something like `HttpOptions` or `HttpConfig` which contains the options-derived state (`hosts` etc); and separate out the ancillary types into a separate class (or an `http.types` package) and the lower primitives can either be removed to eliminate layer duplication, or we can rename that class as `HttpImpl` or `HttpEngine` or something.

That will potentially free up the `Http` name for something that actually represents the API that the rest of the library uses.

Doing that refactor isn't trivial because we also have a lot of tests that rely on those primitive APIs.

Anyway, if we're not doing that yet, how about renaming things as:

`CallbackfulHttp` -> `AbstractHttpExecutor`

`AsyncHttp`, `SyncHttp` can stay the same or be renamed `AsyncHttpExecutor`, `SyncHttpExecutor`

`AnySyncExecutor` -> `HttpExecutor` maybe.

`publishAnySync` -> `executePublish` or `publishImpl` maybe.

lib/src/main/java/io/ably/lib/http/AnySyncHttp.java
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+    }
+
+    public <Result> Request<Result> request(Execute<Result> execute) {
+        return new Request(execute);
+    }
+
+    public interface Execute<Result> {
+        public void execute(CallbackfulHttp http, Callback<Result> callback) throws AblyException;
```
paddybyers: I've getting a raw types warning here and elsewhere; it should be `CallbackfulHttp<? extends Executor> http`.

tcard:
@paddybyers PTAL. I should have split 080a386 in at least a couple of steps but unfortunately I didn't, and since renaming and refactoring went hand in hand it's impossible to do after the fact.

See the comments in 080a386 's commit message.

paddybyers:
This looks very good, thanks.

Are you planning to update the tests in this PR?

tcard:
@paddybyers Yes, I'll do that tomorrow.

tcard:
There are some test that are failing, but I took a look and they didn't seem related to these changes. I'm raising an issue to fix them and merging this PR.

PR #353, opened 2017-09-28
Omit TTL in TokenRequest as JSON if unset.

### LINKED ISSUE #349: TokenRequest.asJson should omit TTL if default
If I call `createTokenRequest` with no additional parameters in Java and JS I get different results.

To compare the two directly I'm using `.asJson` on the Java `TokenRequest`.

The Java version has a key `ttl` that JS does not have.

The JS version has a key `capability` that Java does not have.

tcard:
Fixes #349.


paddybyers:
LGTM, thanks

PR #355, opened 2017-10-16
Push tests.

tcard:
The rest will follow shortly, but this can get work on iOS started.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,23 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:2.2.3'
+    }
+}
+
 apply plugin: 'com.android.library'
 
 ext {
     artifactId = 'ably-android'
+}
+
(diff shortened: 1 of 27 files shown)


mattheworiordan:
Can you confirm what is still outstanding on this?

tcard:
@mattheworiordan This is the stable part only. Then there's the activation state machine, and the PushChannel. Those aren't stable yet.

mattheworiordan:
> @mattheworiordan This is the stable part only. Then there's the activation state machine, and the PushChannel. Those aren't stable yet.

What is the completeness of those sections then?  I was under the impression from the weekly catchups that the Java work was often nearly finished.  If this PR covers `Push.*`, I would assume that there is at least as much work to be done in the "Activate state machine" & `PushChannel`, especially the former given it's the one area with a lot of state complexity no?

tcard:
It's all complete now in a separate PR on top of this one: https://github.com/ably/ably-java/pull/356

tcard:
I've added cherry-picked everything from #356 into this PR, so everything is here now.

PR #360, opened 2018-01-10
Fix dependency resolution for support libraries

funkyboy:
As per Google doc: https://developer.android.com/topic/libraries/support-library/setup.html


funkyboy:
@paddybyers I take this back. There's no need for this change if you use `gradlew`, as you should.
Tried and tested this on a pristine machine.

PR #362, opened 2018-01-30
Update to latest gradle and tools plugins

paddybyers:
Update to
- gradle 4.5;
- android build tools 27.0.3;
- android gradle plugin 2.3.2.

Minor reorganisation of gradle files to make projects more amenable
to import into IDEs.


paddybyers:
@funkyboy PTAL

PR #372, opened 2018-03-13
Add/fix Travis tests

### LINKED ISSUE #364: Add (or fix) CI tests on different platforms
- Java
  - SDK7
  - SDK8

- Android
  - TBD

funkyboy:
Note to self: you **cannot** run x86 images on Travis.
[Related issue](https://github.com/travis-ci/travis-ci/issues/1419)

- [X] Make tests running on `android-22`, java8 and openjdk
- [x] Add Slack and email notifications

Fix #364


funkyboy:
@paddybyers Stuck here: https://travis-ci.org/ably/ably-java/jobs/353722078
Fixing is becoming impractical. 
Will follow [these instructions](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image) and work locally with docker to check what's the problem.

gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.jvmargs=-Xmx1536M
```
funkyboy: Needed to avoid To run dex in process, the Gradle daemon needs a larger heap. message in Travis.

funkyboy:
@paddybyers status so far:
- Android tests are working. API version is 22.
- the `gradlew java:*` command somehow calls stuff that is related to the android “context”. That leads to https://travis-ci.org/ably/ably-java/jobs/353969106#L2000

paddybyers:
> the gradlew java:* command somehow calls stuff that is related to the android “context”. That leads to https://travis-ci.org/ably/ably-java/jobs/353969106#L2000

Are you sure that's the correct link? I see a bytecode version error there

funkyboy:
@paddybyers Yes, that's the link I meant. Also [this](https://travis-ci.org/ably/ably-java/jobs/353969106#L1981) (same error). 

I think that when running `java:*` tests the file `/home/travis/build/ably/ably-java/android/build.gradle` shouldn't be read and could even not exist.

paddybyers:
> I think that when running java:* tests the file /home/travis/build/ably/ably-java/android/build.gradle shouldn't be read and could even not exist.

It might not exist but, since it does, it will be processed. The top-level `build.gradle` is the thing that's being processed to determine the tasks for the project (which include java-specific tasks). When you say `gradle java:xxx` then you're asking for that task in that subproject, but it will still process the whole project.

funkyboy:
@paddybyers sometimes the [simplest solution](https://github.com/ably/ably-java/pull/372/commits/4f33cd77fac32d73a48217d1ca205f8ca742eaa7) works :) https://travis-ci.org/ably/ably-java/jobs/354199784

ci/run-android-tests.sh
```diff
@@ -0,0 +1 @@
+./gradlew connectedAndroidTest --info
```
paddybyers: Do we still need the `--info` ?

funkyboy: No

PR #376, opened 2018-03-16
Use exception instead of deprecation notice

### LINKED ISSUE #139: For 0.9, replace deprecation notice by an exception in BaseMessage.encode.
Basically, [this line](https://github.com/ably/ably-java/blob/16a865928919cae5482793f37a5cf44069e0c95e/lib/src/main/java/io/ably/lib/types/BaseMessage.java#L122) should be removed and [this one](https://github.com/ably/ably-java/blob/16a865928919cae5482793f37a5cf44069e0c95e/lib/src/main/java/io/ably/lib/types/BaseMessage.java#L120) should run unconditionally, ie. removing [this condition](https://github.com/ably/ably-java/blob/16a865928919cae5482793f37a5cf44069e0c95e/lib/src/main/java/io/ably/lib/types/BaseMessage.java#L119).

See #137.

funkyboy:
Fix #139

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/BaseMessage.java
+++ b/lib/src/main/java/io/ably/lib/types/BaseMessage.java
@@ -131,11 +131,7 @@ public class BaseMessage implements Cloneable {
 					encoding = ((encoding == null) ? "" : encoding + "/") + "utf-8";
 				}
 			} else if(!(data instanceof byte[])) {
-				if (opts != null && opts.encrypted) {
-					throw AblyException.fromErrorInfo(new ErrorInfo("Invalid message data or encoding", 400, 40013));
-				} else {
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/types/BaseMessage.java
```diff
@@ -131,11 +131,7 @@ public void encode(ChannelOptions opts) throws AblyException {
 					encoding = ((encoding == null) ? "" : encoding + "/") + "utf-8";
 				}
 			} else if(!(data instanceof byte[])) {
-				if (opts != null && opts.encrypted) {
 					throw AblyException.fromErrorInfo(new ErrorInfo("Invalid message data or encoding", 400, 40013));
```
paddybyers: Surely we also need a whitespace change here as well?

funkyboy: Yup. Fixed.

funkyboy:
@paddybyers Yup. Fixed.

PR #385, opened 2018-05-03
Fix invalid data test

funkyboy:
Fixes this kind of failures https://travis-ci.org/ably/ably-java/jobs/363024037#L2177

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
@@ -825,27 +825,29 @@ public class RealtimeMessageTest extends ParameterizedTest {
 	public void reject_invalid_message_data() throws AblyException {
 		HashMap<String, Integer> data = new HashMap<String, Integer>();
 		Message message = new Message("event", data);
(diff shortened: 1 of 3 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
```diff
@@ -825,27 +825,29 @@ private void expectDataToMatch(MessagesEncodingDataItem fixtureMessage, Message
...
+			@Override
+			public void println(int severity, String tag, String msg, Throwable tr) {
+				capturedLog.add(new LogLine(severity, tag, msg, tr));
+			}
+		});
 
+		try {
 			message.encode(null);
-
+		} catch (AblyException e) {
```
paddybyers: I think there should also be a catch block for the case that a `Throwable` is thrown that isn't an `AblyException` and fail the test.

funkyboy: Done https://github.com/ably/ably-java/pull/385/commits/69995806b33671777dd68e989a2aa3ef56e1994d

PR #386, opened 2018-05-03
Fix connection fail test

funkyboy:
In principle it should never get neither into the connected state, nor into the disconnected one. Correct?


funkyboy:
I misunderstood. There's a bug in the code. The test is correct.

Issue #358, opened 2017-11-10
Implement connection state freshness check


SimonWoolf:
Spec PR: https://github.com/ably/docs/pull/331/files

This is quite high priority; see https://github.com/ably/realtime/issues/1289#issuecomment-341152556 for why, or ask me or Matt in #client-libraries on slack.

See https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L702-L712 for example implementation

paddybyers:
Fixed by https://github.com/ably/ably-java/pull/390

PR #389, opened 2018-05-08
Add connectionStateTtl

funkyboy:
In preparation for https://github.com/ably/ably-java/issues/358 

- [x] Remove `TIMEOUT_SUSPEND`
- [x] add `connectionStateTtl` and use it instead of `TIMEOUT_SUSPEND`

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
@@ -531,7 +531,8 @@ public class ConnectionManagerTest extends ParameterizedTest {
 
 	/**
 	 * RTN15g3. Connect, attach some channels, disconnect, reconnect after (ttl + idle interval) period has passed,
-	 * check that the client reconnects to the same channels using the new connection;
+	 * check that the client reconnects with a different connection and that the channels attached during the first
+	 * connection are correctly reattached;
 	 */
 	@Test
 	public void channels_are_reattached_after_reconnecting_when_statettl_plus_idleinterval_has_passed() {
@@ -565,17 +566,11 @@ public class ConnectionManagerTest extends ParameterizedTest {
 
 			/* Attach to channel */
 			final Channel channel = ably.channels.get(channelName);
-			channel.on(new ChannelStateListener() {
-				@Override
-				public void onChannelStateChanged(ChannelStateChange stateChange) {
-					System.out.println("======================= state is " + stateChange.current);
-				}
-			});
 			channel.once(ChannelState.attached, new ChannelStateListener() {
 				@Override
 				public void onChannelStateChanged(ChannelStateChange stateChange) {
 					System.out.println("Channel attached for the first time");
-					assertEquals("Channel is attached", stateChange.current, ChannelState.attached);
+					assertEquals("Channel was not attached", stateChange.current, ChannelState.attached);
 				}
 			});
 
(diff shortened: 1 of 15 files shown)


lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -388,4 +389,32 @@ public void run() {
...
+	 */
+	@Test
+	public void connection_details_has_ttl() {
+		try {
+			ClientOptions opts = createOptions(testVars.keys[0].keyStr);
+			final AblyRealtime ably = new AblyRealtime(opts);
+			ably.connection.on(ConnectionEvent.connected, new ConnectionStateListener() {
+				@Override
+				public void onConnectionStateChanged(ConnectionStateChange state) {
+					assertEquals(1, 1);
```
paddybyers: What's the point of this?

mattheworiordan: Why is this still here?

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -388,4 +389,32 @@ public void run() {
...
+					try {
+						Field field = ably.connection.connectionManager.getClass().getDeclaredField("connectionStateTtl");
+						field.setAccessible(true);
+						assertEquals("Verify connectionStateTtl has the default value", field.get(ably.connection.connectionManager), 120000L);
+					} catch (NoSuchFieldException|IllegalAccessException e) {
+						fail("Unexpected exception in checking connectionStateTtl");
+					}
+					ably.close();
+				}
+			});
```
mattheworiordan: There is no wait as far as I can tell so the `onConnectionStateChanged` is not necessarily waited for?  How do you know this test even run and assertion succeeded?

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
funkyboy: @paddybyers @mattheworiordan Removed https://github.com/ably/ably-java/pull/389/commits/26c18f67a27d90d292ad53e3fed881f645bde190

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
funkyboy: @mattheworiordan I made the test more explicit https://github.com/ably/ably-java/pull/389/commits/850eaf9d2a7f186bcd73ad1594903142eccd0086

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -408,9 +408,10 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
 					} catch (NoSuchFieldException|IllegalAccessException e) {
 						fail("Unexpected exception in checking connectionStateTtl");
 					}
-					ably.close();
 				}
 			});
+			new Helpers.ConnectionManagerWaiter(ably.connection.connectionManager).waitFor(ConnectionState.connected);
```
paddybyers: I agree this is an improvement, but even better would have been to have an explicit check that the callback ran

funkyboy: @paddybyers done https://github.com/ably/ably-java/pull/389/commits/9de807f16a0dc59c228b2701db163f23f752dfc8

PR #393, opened 2018-06-07
Add jwt tests

### LINKED ISSUE #384: Add test for JWT token

funkyboy:
Fix #384 

- [X] Add tests that use `ClientOptions`
- [X] Add tests that use `authURL`
- [X] Add tests that use `authCallback`
- [X] Add tests with JWT wrapping an Ably native token `x-ably-token`
- [X] Add tests with encrypted token
- [X] Add example when token is returned with `application/jwt` Content-Type
- [X] Add tests with JWT that includes client_id
- [X] Add tests with JWT that includes publish/subscribe capabilities
- [X] Add tests for automatic reauth with and without disconnection
- [X] Add refs to spec items

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
@@ -193,6 +193,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 		try {
 			final String[] tokens = new String[1];
 			final boolean[] authMessages = new boolean[] { false };
+			final boolean[] updateEvents = new boolean[] { false };
 
 			/* create ably realtime with authUrl and params that include a ttl of 35 seconds */
 			DebugOptions options = new DebugOptions(testVars.keys[0].keyStr);
@@ -239,6 +240,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 				public void onConnectionStateChanged(ConnectionStateChange state) {
 					assertNotEquals("Token should not be the same", tokens[0], ablyRealtime.auth.getTokenDetails().token);
 					assertTrue("Auth protocol message has not been received", authMessages[0]);
+					updateEvents[0] = true;
 					ablyRealtime.close();
 				}
 			});
@@ -250,6 +252,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 
 			/* wait for closed state */
 			connectionWaiter.waitFor(ConnectionState.closed);
+			assertTrue("Update event not received", updateEvents[0]);
 		} catch (AblyException e) {
 			e.printStackTrace();
 			fail();
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
@@ -22,7 +22,6 @@ public class RealtimeJWTTest extends ParameterizedTest {
 
 	private AblyRest restJWTRequester;
 	private ClientOptions jwtRequesterOptions;
-	private ClientOptions realtimeOptions;
 	private Key key = testVars.keys[0];
 	private final String clientId = "testJWTClientID";
 	private final String channelName = "testJWTChannel" + UUID.randomUUID().toString();
@@ -38,7 +37,6 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	public void setUpBefore() throws Exception {
 		jwtRequesterOptions = createOptions(testVars.keys[0].keyStr);
 		jwtRequesterOptions.authUrl = echoServer;
-		realtimeOptions = createOptions();
 	}
 
 	/**
@@ -49,7 +47,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	public void auth_clientid_match_the_one_requested_in_jwt() {
 		try {
 			/* create ably realtime with JWT token */
-			realtimeOptions.token = getToken(mergeParams(keys, clientIdParam), null);
+			ClientOptions realtimeOptions = buildClientOptions(mergeParams(keys, clientIdParam), null);
 			assertNotNull("Expected token value", realtimeOptions.token);
 			AblyRealtime ablyRealtime = new AblyRealtime(realtimeOptions);
 
@@ -76,7 +74,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	public void auth_jwt_with_subscribe_only_capability() {
 		try {
 			/* create ably realtime with JWT token that has subscribe-only capabilities */
-			realtimeOptions.token = getToken(keys, susbcribeOnlyCapability);
+			ClientOptions realtimeOptions = buildClientOptions(keys, susbcribeOnlyCapability);
 			assertNotNull("Expected token value", realtimeOptions.token);
 			final AblyRealtime ablyRealtime = new AblyRealtime(realtimeOptions);
 
@@ -120,7 +118,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	public void auth_jwt_with_publish_capability() {
 		try {
 			/* create ably realtime with JWT token that has publish capabilities */
-			realtimeOptions.token = getToken(keys, publishCapability);
+			ClientOptions realtimeOptions = buildClientOptions(keys, publishCapability);
 			assertNotNull("Expected token value", realtimeOptions.token);
 			final AblyRealtime ablyRealtime = new AblyRealtime(realtimeOptions);
 
@@ -163,7 +161,7 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	public void auth_jwt_with_token_that_expires() {
 		try {
 			/* create ably realtime with JWT token that expires in 5 seconds */
-			realtimeOptions.token = getToken(mergeParams(keys, shortTokenTtl), null);
+			ClientOptions realtimeOptions = buildClientOptions(mergeParams(keys, shortTokenTtl), null);
 			assertNotNull("Expected token value", realtimeOptions.token);
 			final AblyRealtime ablyRealtime = new AblyRealtime(realtimeOptions);
 
@@ -263,9 +261,9 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	}
 
 	/**
-	 * Helper to fetch a token with params and capability via authUrl
+	 * Helper to create ClientOptions with a JWT token fetched via authUrl according to the parameters
 	 */
-	private String getToken(Param[] params, String capability) {
+	private ClientOptions buildClientOptions(Param[] params, String capability) {
 		jwtRequesterOptions.authParams = params;
 		try {
 			restJWTRequester = new AblyRest(jwtRequesterOptions);
@@ -274,9 +272,11 @@ public class RealtimeJWTTest extends ParameterizedTest {
 				t.capability = capability;
 			}
 			TokenDetails tokenDetails = restJWTRequester.auth.requestToken(t, null);
-			return tokenDetails.token;
+			ClientOptions realtimeOptions = createOptions();
+			realtimeOptions.token = tokenDetails.token;
+			return realtimeOptions;
 		} catch (AblyException e) {
-			fail("Failure in fetching a JWT token" + e);
+			fail("Failure in fetching a JWT token to create ClientOptions " + e);
 			return null;
 		}
 	}
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
@@ -371,17 +371,22 @@ public class RealtimeJWTTest extends ParameterizedTest {
 	 */
 	private ClientOptions buildClientOptions(Param[] params, String capability) {
 		try {
-			jwtRequesterOptions = createOptions(testVars.keys[0].keyStr);
-			jwtRequesterOptions.authUrl = echoServer;
-			jwtRequesterOptions.authParams = params;
-			restJWTRequester = new AblyRest(jwtRequesterOptions);
-			TokenParams t = new TokenParams();
-			if (capability != null) {
-				t.capability = capability;
-			}
-			TokenDetails tokenDetails = restJWTRequester.auth.requestToken(t, null);
+			final String[] resultToken = new String[1];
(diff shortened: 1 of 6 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+	Param[] mediumTokenTtl = new Param[] { new Param("expiresIn", 35) };
+	private final String susbcribeOnlyCapability = "{\"" + channelName + "\": [\"subscribe\"]}";
+	private final String publishCapability = "{\"" + channelName + "\": [\"publish\"]}";
+
+	@Before
+	public void setUpBefore() throws Exception {
+		jwtRequesterOptions = createOptions(testVars.keys[0].keyStr);
+		jwtRequesterOptions.authUrl = echoServer;
+		realtimeOptions = new ClientOptions();
+		realtimeOptions.environment = "sandbox";
```
paddybyers: This should not be specified here; the test setup sets it from the environment.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+	/**
+	 * Request a JWT that specifies a clientId
+	 * Verifies that the clientId matches the one requested
+	 */
+	@Test
+	public void auth_clientid_match_the_one_requested_in_jwt() {
+		try {
+			/* create ably realtime with JWT token */
+			realtimeOptions.token = getToken(mergeParams(keys, clientIdParam), null);
```
paddybyers: I don't like reusing the same `realtimeOptions` for multiple tests. Please create a new options for each test.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+	Param[] clientIdParam = new Param[] { new Param("clientId", clientId) };
+	Param[] shortTokenTtl = new Param[] { new Param("expiresIn", 5) };
+	Param[] mediumTokenTtl = new Param[] { new Param("expiresIn", 35) };
+	private final String susbcribeOnlyCapability = "{\"" + channelName + "\": [\"subscribe\"]}";
+	private final String publishCapability = "{\"" + channelName + "\": [\"publish\"]}";
+
+	@Before
+	public void setUpBefore() throws Exception {
+		jwtRequesterOptions = createOptions(testVars.keys[0].keyStr);
+		jwtRequesterOptions.authUrl = echoServer;
```
paddybyers: I don't like the fact that the same options instance is being used for every test. If necessary add a helper method to create an options and populate it, so you can easily create a new one each time.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+			Channel channel = ablyRealtime.channels.get(channelName);
+			channel.attach();
+			new ChannelWaiter(channel).waitFor(ChannelState.attached);
+
+			/* publish and verify that it fails */
+			channel.publish(messageName, null, new CompletionListener() {
+				@Override
+				public void onSuccess() {
+					fail("It should not succeed");
+				}
```
paddybyers: I would add a `ablyRealtime.close()` in here

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+				@Override
+				public void onSuccess() {
+					System.out.println("Message " + messageName + " published successfully");
+					ablyRealtime.close();
+				}
+
+				@Override
+				public void onError(ErrorInfo reason) {
+					fail("Publish should not fail");
+				}
```
paddybyers: I would add `ablyRealtime.close()` in here

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+	 */
+	@Test
+	public void auth_jwt_with_client_than_reauths_without_disconnecting() {
+		try {
+			final String[] tokens = new String[1];
+			final boolean[] authMessages = new boolean[] { false };
+
+			/* create ably realtime with authUrl and params that include a ttl of 35 seconds */
+			DebugOptions options = new DebugOptions(testVars.keys[0].keyStr);
+			options.environment = "sandbox";
```
paddybyers: Please don't specify the environment explicitly. That stops the tests from being run against any other environment. Use `createOptions()` or something to get options where the `environment` is set according to the test setup.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+			/* Once receiving the update event check that the token is a new one
+			 * and verify the auth protocol message has been received. */
+			ablyRealtime.connection.on(ConnectionEvent.update, new ConnectionStateListener() {
+				@Override
+				public void onConnectionStateChanged(ConnectionStateChange state) {
+					assertNotEquals("Token should not be the same", tokens[0], ablyRealtime.auth.getTokenDetails().token);
+					assertTrue("Auth protocol message has not been received", authMessages[0]);
+					ablyRealtime.close();
+				}
```
paddybyers: Ideally we'd have something that confirms at the end of this test that this code executed.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+			});
+
+			/* wait for connected state */
+			ConnectionWaiter connectionWaiter = new ConnectionWaiter(ablyRealtime.connection);
+			connectionWaiter.waitFor(ConnectionState.connected);
+			assertEquals("Connected state was NOT reached", ConnectionState.connected, ablyRealtime.connection.state);
+
+			/* wait for closed state */
+			connectionWaiter.waitFor(ConnectionState.closed);
+		} catch (AblyException e) {
```
paddybyers: We don't have anything that confirms that the `authURL` was called. I'm not sure that there's an easy way to do that. We could of course have an authCallback analogue of this test, where we can easily confirm that the callback is called.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeJWTTest.java
```diff
@@ -0,0 +1,285 @@
...
+			e.printStackTrace();
+			fail();
+		}
+	}
+
+	/**
+	 * Helper to fetch a token with params and capability via authUrl
+	 */
+	private String getToken(Param[] params, String capability) {
```
paddybyers: I don't think we should be instancing the Rest library just in order to make an http request to the echoserver. Just make an Http request - there is a helper available on the library, which is used by some other tests.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+	private Key key = testVars.keys[0];
+	Param[] validKeys = new Param[]{ new Param("keyName", key.keyName), new Param("keySecret", key.keySecret) };
+	Param[] invalidKeys = new Param[]{ new Param("keyName", key.keyName), new Param("keySecret", "invalidinvalid") };
+	Param[] tokenEmbedded = new Param[]{ new Param("jwtType", "embedded") };
+	Param[] tokenEmbeddedAndEncrypted = new Param[]{ new Param("jwtType", "embedded"), new Param("encrypted", 1) };
+	Param[] jwtReturnType = new Param[]{ new Param("returnType", "jwt") };
+
+	@Before
+	public void setUpBefore() throws Exception {
+		jwtRequesterOptions = createOptions(testVars.keys[0].keyStr);
```
paddybyers: Same comment as for the realtime tests. Please don't use the same options instances for every test - create new ones; if necessary add a helper method to do that.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+		jwtRequesterOptions.authUrl = echoServer;
+		options = new ClientOptions();
+		options.environment = "sandbox";
```
paddybyers: Again, please don't hard-code the environment.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/**
+	 * Base request of a JWT token (RSA8g RSA8c)
+	 */
+	@Test
+	public void authjwtrequest() {
```
paddybyers: Why here no snakes or camels in the test names?

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+		} catch (AblyException e) {
+			e.printStackTrace();
+			fail("authjwtrequestauthcallback: Unexpected exception");
+		}
+	}
+
+	/**
+	 * Helper to fetch a token with params via authUrl
+	 */
+	private String getTokenWithParams(Param[] params) {
```
paddybyers: Again, I don't think we should be instancing the REST library just in order to make an HTTP request to the echoserver. Make an HTTP request directly.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	/**
+	 * Request of a JWT token that is returned with application/jwt content type (RSA4f, RSA8c)
+	 */
+	@Test
+	public void authjwtrequestreturntype() {
+		try {
+			options.token = getTokenWithParams(mergeParams(validKeys, jwtReturnType));
```
paddybyers: I thought `getTokenWithParams()` was a helper method to obtain a token, but the AblyRest instance created in the test itself was the instance that's the subject of the test. If that's the case, then this test should be using `authURL` in this instance, and asking the echoserver to reply with `application/jwt` here, rather than doing that in `getTokenWithParams()`.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+				public Object getTokenRequest(TokenParams params) throws AblyException {
+					return restJWTRequester.auth.createTokenRequest(params, null);
+				}
+			};
+			ClientOptions optionsWithCallback = new ClientOptions();
+			optionsWithCallback.authCallback = authCallback;
+			optionsWithCallback.environment = "sandbox";
+			AblyRest client = new AblyRest(optionsWithCallback);
+			PaginatedResult<Stats> stats = client.stats(null);
+			assertNotNull("Stats should not be null", stats);
```
paddybyers: I think we need an assertion that proves that the callback was called.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+			restJWTRequester = new AblyRest(createOptions(testVars.keys[0].keyStr));
+			TokenCallback authCallback = new TokenCallback() {
+				@Override
+				public Object getTokenRequest(TokenParams params) throws AblyException {
+					return restJWTRequester.auth.createTokenRequest(params, null);
+				}
+			};
+			ClientOptions optionsWithCallback = new ClientOptions();
+			optionsWithCallback.authCallback = authCallback;
+			optionsWithCallback.environment = "sandbox";
```
paddybyers: Please no hard-coded environment.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
```diff
@@ -0,0 +1,157 @@
...
+	 * Request of a JWT token via authCallback (RSA8g)
+	 */
+	@Test
+	public void authjwtrequestauthcallback() {
+		try {
+			restJWTRequester = new AblyRest(createOptions(testVars.keys[0].keyStr));
+			TokenCallback authCallback = new TokenCallback() {
+				@Override
+				public Object getTokenRequest(TokenParams params) throws AblyException {
+					return restJWTRequester.auth.createTokenRequest(params, null);
```
paddybyers: This is going to be returning a `TokenRequest` and not a JWT, which isn't what we're supposed to be testing.

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
funkyboy: done https://github.com/ably/ably-java/pull/393/commits/c558322d3d248f65b0b9fd2f400d836c334273e1

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
funkyboy: done

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
funkyboy: done https://github.com/ably/ably-java/pull/393/commits/ba5f933d08647660523c29ac259bd5c7de8739ac

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
funkyboy: done https://github.com/ably/ably-java/pull/393/commits/d9c2dd27dbe849be183139af454bfa8e7982028b

lib/src/test/java/io/ably/lib/test/rest/RestJWTTest.java
funkyboy: done https://github.com/ably/ably-java/pull/393/commits/6a02ea7486214c7a0b1295ff8ad8dbbb277e3f36

funkyboy:
@paddybyers addressed all your comments. Ready for another look.

PR #404, opened 2018-06-28
Fix cm thread exit

paddybyers:
Fix the synchronisation around destruction of the connection manager thread and subsequent recreation.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1011,9 +1011,9 @@ public class ConnectionManager implements ConnectListener {
 					transport.abort(e.errorInfo);
 				}
 			} else {
-				/* just abort the transport */
+				/* just close the transport */
 				Log.v(TAG, "Aborting incomplete transport due to close()");
-				transport.abort(REASON_CLOSED);
+				transport.close(false);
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -800,75 +826,72 @@ private void tryWait(long timeout) {
...
+						synchronized(this) {
+							notify();
+						}
 					}
-				}
 
-				while(!state.terminal && stateChange == null) {
-					/* wait for a state change event or for expiry of the current state */
-					tryWait(state.timeout);
+					while(!exiting && stateChange == null) {
```
funkyboy: Aren't we already in a `!exiting` context given https://github.com/ably/ably-java/pull/404/files?utf8=%E2%9C%93&diff=split#diff-060685a947a7032665de7ac9d33bb6ffR838 ? Do we need this double check?

paddybyers: All we know is that we were `!exiting` when we entered the loop at the top. Since then we will have waited for state changes, so it's no longer expected to be true.

funkyboy:
Merging now so I can rebase and see if the `reauth_fail()` test is now passing.

PR #437, opened 2018-11-21
[WIP] RTN20: Add network connectivity support for Android

paddybyers:
This implements support for [RTN20](https://docs.ably.io/client-lib-development-guide/features/#RTN20):

> (RTN20) When the client library can subscribe to the Operating System events for network/internet connectivity changes:
(RTN20a) When `CONNECTED`, `CONNECTING` or `DISCONNECTING`, if the operating system indicates that the underlying internet connection is no longer available, then the client library should immediately transition the state to `DISCONNECTED` with emit a state change with an appropriate reason. This state change will automatically trigger the client library to attempt to reconnect, see RTN15 above
(RTN20b) When `DISCONNECTED` or `SUSPENDED`, if the operating system indicates that the underlying internet connection is now available, the client library should immediately attempt to connect

Registration for these events requires an Android [Context](https://developer.android.com/reference/android/content/Context), and this PR adds a number of methods to the Ably REST/Realtime API:

```
    /**
     * Get the Android Context for this instance
     * @return Context context
     */
    public Context getAndroidContext();

    /**
     * Set the Android Context for this instance
     * @param Context context
     */
    public void setAndroidContext(Context context);

    /**
     * Clear the Android Context for this instance
     */
    public void clearAndroidContext();
```

Setting a context will trigger registration of an intent filter for the Android `CONNECTIVITY_ACTION` events; clearing the context will unregister the filter. These methods should be called in the relevant lifecycle handlers (`onResume()`, `unPause()`, etc) for the Activity or Service in which the the library is instanced.


paddybyers:
This functionality has been merged separately

PR #439, opened 2018-12-06
Add support for bulk rest publish API

paddybyers:
This adds support for the bulk publish API:

    POST /messages

This performs a publish in parallel of messages to channels specified in the request body.

A single `BatchSpec` is an object of the form:

````
{
  channels: <channels>,
  messages: <messages>
}
````

where:
- `<channels>` is a single channel name `String`, or an `Array` of channel name `Strings`; and
- `<messages>` is a single `Message`, or an `Array` of `Message`s.

The bulk publish endpoint accepts a request body containing either a single `BatchSpec` object, or an array of `BatchSpec` objects.

Therefore the following are all valid request bodies for a bulk publish request:

````
{
  channels: ['a channel name, containing a comma', 'another channel name'],
  messages: {data: 'My message contents'}
}
````

````
[
  {
    channels: ['a channel name, containing a comma', 'another channel name'],
    messages: {data: 'My message contents'}
  },
  {
    channels: 'single channel',
    messages: [
      {data: 'My message contents'},
      {name: 'an event', data: 'My event message contents'},
    ]
  }
]
````

There is an obvious mapping from the request body to the array of individual publish requests.

### DIFF
--- a/lib/src/main/java/io/ably/annotation/Experimental.java
+++ b/lib/src/main/java/io/ably/annotation/Experimental.java
@@ -1,0 +1,10 @@
+package io.ably.annotation;
+
+import java.lang.annotation.Documented;
+
+/**
+ * An annotation indicating an experimental API. Any or all detail of this
+ * API are subject to change in the future. Feedback on the API is welcomed.
+ */
+@Documented
+public @interface Experimental { }
--- a/lib/src/main/java/io/ably/lib/http/HttpCore.java
+++ b/lib/src/main/java/io/ably/lib/http/HttpCore.java
@@ -273,10 +273,22 @@ public class HttpCore {
 			throw AblyException.fromErrorInfo(error);
 		}
 
-		if(response.statusCode < 200 || response.statusCode >= 300) {
-			/* get any in-body error details */
-			ErrorInfo error = null;
-			if(response.body != null && response.body.length > 0) {
+		if(response.statusCode >= 200 && response.statusCode < 300) {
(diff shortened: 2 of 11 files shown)


mattheworiordan:
Interesting.  Is this meant to be a working prototype towards adding this to the spec later?  Perhaps in 1.2?

paddybyers:
> Is this meant to be a working prototype towards adding this to the spec later? Perhaps in 1.2?

Yes, that's the question I suppose.

bbeaudreault:
Would it be acceptable to mark the API experimental or beta with an annotation? That is a common pattern in other libraries, which leaves space for early access to APIs while also accounting for future changes when it becomes standard.

paddybyers:
Googling around the consensus seems to be to have an `@experimental` annotation, but there's no standard definition of one. There's also the option of `@deprecated(EXPERIMENTAL)` which is a misuse of `deprecated` but you get compiler/IDE support for it.

bbeaudreault:
Annotations are easy to create, I think it's really more about the documentation it provides paired with javadoc and a disclosure in the API docs.

paddybyers:
> Are we sure adding a rest#publish here isn't going to limit our options later?

I don't know tbh but, give that, I'm ok to name it `publishBatch()`, at least until we've got a decision.

See https://github.com/ably/ably-java/pull/439/commits/98c072f179ac2f9069d6cf3c0bf3bacced1d561a

PR #441, opened 2018-12-07
Call listeners in EventEmitter#emit only once

amsurana:
I have added support for Feature RTE6a in Java.

### DIFF
--- a/lib/src/main/java/io/ably/lib/util/EventEmitter.java
+++ b/lib/src/main/java/io/ably/lib/util/EventEmitter.java
@@ -8,107 +8,126 @@ import java.util.Map;
 
 /**
  * An interface exposing the ability to register listeners for a class of events
- * @author paddy
  *
- * @param <Event> an Enum containing the event names that listeners may be registered for
+ * @param <Event>    an Enum containing the event names that listeners may be registered for
  * @param <Listener> the interface type of the listener
+ * @author paddy
  */
 public abstract class EventEmitter<Event, Listener> {
 
-	/**
-	 * Remove all registered listeners irrespective of type
-	 */
-	public synchronized void off() {
-		listeners.clear();
-		filters.clear();
-	}
+    /**
+     * Remove all registered listeners irrespective of type
+     */
+    public synchronized void off() {
+        listeners.clear();
+        filters.clear();
+    }
 
-	/**
-	 * Register the given listener for all events
-	 * @param listener
-	 */
-	public synchronized void on(Listener listener) {
-		if(!listeners.contains(listener))
-			listeners.add(listener);
-	}
+    /**
+     * Register the given listener for all events
+     *
+     * @param listener
+     */
+    public synchronized void on(Listener listener) {
+        if (!listeners.contains(listener))
+            listeners.add(listener);
+    }
 
-	/**
-	 * Register the given listener for a single occurrence of any event
-	 * @param listener
-	 */
-	public synchronized void once(Listener listener) {
-		filters.put(listener, new Filter(null, listener, true));
-	}
+    /**
+     * Register the given listener for a single occurrence of any event
+     *
+     * @param listener
+     */
+    public synchronized void once(Listener listener) {
+        filters.put(listener, new Filter(null, listener, true));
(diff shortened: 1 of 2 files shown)


lib/src/test/java/io/ably/lib/test/realtime/EventEmitterTest.java
```diff
@@ -175,4 +176,30 @@ public void once_off_event_simple() {
 		assertNull(listener.counts.get(MyEvents.event_1));
 	}
 
+    @Test
```
paddybyers: Did this test fail before you updated the code? Because if it adds a listener to the end of the collection, then I wouldn't expected it to have been called in the previous implementation.

In any case, please can there also be a test also that removes a listener, and verifies it is still called?

lib/src/test/java/io/ably/lib/test/realtime/EventEmitterTest.java
```diff
@@ -175,4 +176,30 @@ public void once_off_event_simple() {
 		assertNull(listener.counts.get(MyEvents.event_1));
 	}
```
paddybyers: There needs to be a comment here please with a reference to the requirement `RTE6a`

lib/src/main/java/io/ably/lib/util/EventEmitter.java
```diff
@@ -82,8 +82,10 @@ public synchronized void off(Event event, Listener listener) {
 	 * @param args the arguments to pass to listeners
 	 */
 	public synchronized void emit(Event event, Object... args) {
-		for (int i = listeners.size() - 1; i >= 0; i--) {
-			apply(listeners.get(i), event, args);
+		List<Listener> clonedListeners = new ArrayList<>();
```
paddybyers: I think here it is worth having a comment with a reference to the spec requirement id `RTE6a`.

lib/src/test/java/io/ably/lib/test/realtime/EventEmitterTest.java
amsurana: The test never failed because there's no Listener added for MyEvents.event_1 in the test case. 
According to what I see, Listener is added for MyEvents.event_0, emitted for MyEvents.event_1, and then removed for MyEvents.event_0. 

So the test does pass before and after. 
I have added one test case `once_off_event` just to add listener, emit, remove listener and emit again.

paddybyers: > The test never failed because there's no Listener added for MyEvents.event_1 in the test case.

OK, so how do you know that your change fixed anything, if you did not have a failing test? It's not always the case, but here I think, with knowledge of how things can fail with the old code, I think you could create a test that fails before the fix is made. Can you think of a way to do that?

amsurana: I tested this in code, but I might not have added a Junit test for this. I will add a test case as well and push commit. Thanks.

paddybyers: OK, but here's a scenario that we know will fail with the old code, but will pass with the fixed code:

- add 2 listeners A, B, for an event;
- the listener for B removes A;
- call `emit()`;

With the old code B will be called twice, but with the fixed code A and B will each be called once.

paddybyers: OK, well lets add a test for this removal case, which shows that you can remove a listener, from within the execution of another listener, without an exception. If we can we really have to have a test that shows the old behaviour is broken, and the new behaviour fixes it.

amsurana:
Updated the PR and added support for:
1. Add listener in listener
2. Remove listener in listener
3. Add event specific listener in listener
4. Add event specific listener in listener
5. Add "once" listener in listener.

paddybyers:
@amsurana your editor has changed all the files it touches from tabs to spaces. We don't want that changed please, and it also makes it impossible to review the changes. Please revert those changes and force-push the change so we can see the actual diff.

PR #444, opened 2018-12-12
Implemented RTE6a specification

amsurana:
Created the PR for implementing support for RTE6a.

Test cases added in EventEmitterTest.java

- Add listener in listener
- Remove listener in listener
- Add event specific listener in listener
- Add event specific listener in listener
- Add "once" listener in listener.


paddybyers:
@amsurana for future reference, please create distinct and descriptive branch names. It's confusing if you use `develop` as the branch name in your repo - ideally if you have any branch in your repo that has the same name as a branch in the upstream repo, then they should be the same - ie yours tracks the branch in the upstream repo.

PR #445, opened 2018-12-19
Implemented Spec: TM3, Message.fromEncoded

amsurana:


lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
```diff
@@ -817,4 +819,19 @@ public LogLine(int severity, String tag, String msg, Throwable tr) {
 			this.tr = tr;
 		}
 	}
+
+	@Test
```
paddybyers: We need to improve the tests. There are two approaches:

- construct a `Message`, and set various attributes, and then encode(), and then attempt to decode with these methods;

- create some fixture data and add it to the `ably-common` repo with the other fixture data.

PR #465, opened 2019-03-05
Removed code that logs firebase token

amsurana:


paddybyers:
Hi, sorry, I think maybe I should have been clearer. I think the log line should remain - it is going to be useful to have a log entry when the token is renewed - just don't include the token in there. Or, better still, include the first 16 characters of the token, perhaps, so at least that log line can be correlated with any token that is persisted, or used in an API request.

Also - please use descriptive branch names. I know this is in your repo, so the branch names are up to you, but using the same name as the target branch isn't going to scale well. Didn't we give you rights to push branches into our repo?

Issue #470, opened 2019-04-23
Expired token never renewed


paddybyers:
A customer error report:

TL;DR I’ve noticed a pretty significant bug in the ably-java sdk.  I’ve built in a work around, but essentially if you have something like this line in the Auth class of your sdk you may be at risk.
https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/rest/Auth.java#L969

bottom line is that the token is always considered valid if the `timeDelta` has not been set and in the ably-java sdk that property is not set for TokenDetails objects (it is set for TokenRequests objects which we don’t use client side).

Prerequisites
- auth callback in ClientOptions returns a TokenDetails object that expires after 5 minutes

Steps
1. connect to Ably with an auth callback
2. disconnect from Ably
3. attempt to reconnect to Ably after 6 minutes (just has to be longer than token expiry)

Expected Result
Ably should use the auth callback to retrieve a new token and reconnect successfully.

Actual Result
Ably continues to use expired token and fails to reconnect. 
[ErrorInfo message=Error verifying JWT; err = jwt expired. (See https://help.ably.io/error/40144 for help.) code=40144 statusCode=401 href=https://help.ably.io/error/40144]

Cause
- `Auth.timeDelta` is only set to a value (other than `Long.MAX_VALUE`) in `createTokenRequest` which is only called if the auth callback returns a TokenRequest object, not a TokenDetails object.
- When checking if a token is expired, the `tokenValid` method is called, but because `Auth.timeDelta` will always be `Long.MAX_VALUE` this method always returns `true`.
- With the `tokenValid` method always returning `true`, the cached token is never cleared and used in all subsequent connection attempts.

paddybyers:
Note that https://github.com/ably/ably-java/pull/471 needs to be applied also to `develop`.

paddybyers:
Closing; this fix has also now been merged for 1.1

PR #471, opened 2019-04-23
ConnectionManager: ensure that cached token details are cleared on any connection error

paddybyers:
Fixes: https://github.com/ably/ably-java/issues/470


mattheworiordan:
@paddybyers does this not need to be applied to 1.1.*?

paddybyers:
> does this not need to be applied to 1.1.*?

yes it will

Issue #478, opened 2019-06-12
Push: LocalDevice.reset() doesn't clear persisted device state


paddybyers:
Found using the Android push example app

amsurana:
Addressed the issue. Updated PR #479

paddybyers:
Fixed by https://github.com/ably/ably-java/pull/479

PR #479, opened 2019-06-18
Local device reset fix

amsurana:
Clear persisted local storage when LocalDevice.reset() is called.
Refer: https://github.com/ably/ably-java/issues/478


paddybyers:
LGTM, but I think `ActivationStateMachine` has the same issue: https://github.com/ably/ably-java/blob/47d60db2aa9af7c60cbb4c41a8c3e4177abe7b15/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java#L479

PR #496, opened 2019-09-17
[WIP]Initial attempt on codec infrastructure

codemerx:


mattheworiordan:
@codemerx please can you a) ensure PRs have a description, b) if it's a WIP, then [use a Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)

paddybyers:
> if it's a WIP, then [use a Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/)

Can you revert a regular PR into a draft PR? I didn't find a way. I've found a couple of times I needed to put `[WIP]` in the subject line instead if, after creating a PR, it became necessary to put it back into a draft state.

mattheworiordan:
>  I've found a couple of times I needed to put [WIP] in the subject line instead if, after creating a PR, it became necessary to put it back into a draft state.

I don't think you can revert it.  The flow is create a Draft PR > create a PR ready for review.  This PR was created with just https://github.com/ably/ably-java/pull/496/commits/67a2148f30caa46cfbf67d0a98c90cec3ea9dc53 and no description, so was draft from the start.  I appreciate if you need go back to draft that is different, but in that case closing it and creating another PR may be better, or exceptionally we can prefix `[WIP]`.

lib/src/main/java/io/ably/lib/types/ClientOptions.java
```diff
@@ -194,4 +195,7 @@ public ClientOptions(String key) throws AblyException {
 	 * before responding.
 	 */
 	public boolean pushFullWait = false;
+
+
+	public final Map<String, AblyCodec> userProvidedCodecs = new HashMap<>();
```
paddybyers: Didn't we agree that this would simply be `codecs` ?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
```diff
@@ -13,6 +13,7 @@
 import java.util.Locale;
 
 import com.google.gson.*;
+import io.ably.lib.types.*;
```
paddybyers: I realise that the IDE did this on your behalf, but preferably keep PRs to just the changes that are the subject of the PR.

codemerx:
> 
> 
> This looks good, but needs bringing up to date with the latest API proposal, at least in respect of `ChannelOptions.codecs`.
> 
> We need tests - we've discussed the various edge cases: we need to test each of the transports, but the tested cases need to include all relevant/problematic combinations in succession of: binary `data`, string `data`, JSON `data`, empty data, etc.
> 
> Also, what about negative tests - eg mock the codec and throw a decode exception and see what happens?
> 
> Also, we need the README to make some reference to this functionality, with an example.

As discussed tests and CI postponed until after library structure gets finalized

paddybyers:
Closed in favour of https://github.com/ably/ably-java/pull/544

Issue #495, opened 2019-08-20
After loss of connectivity, and transport closure due to timeout, the ConnectionManager still thinks the transport is active


paddybyers:
An extract from a log:

```
[Ably internal] Unexpected exception in WsClient Exception: java.net.UnknownHostException: realtime.ably.io

[Ably internal] Unexpected exception attempting connection; wsUri = wss://realtime.ably.io:443/ Exception: io.ably.lib.types.AblyException: io.ably.lib.types.AblyException$HostFailedException: java.net.UnknownHostException: <redacted>
[Ably internal] Unexpected exception in WsClient Exception: java.net.UnknownHostException: realtime.ably.io


[Ably internal] No activity for 25000ms, closing connection

...

Error sending request <redacted> to ably. io.ably.lib.types.AblyException: org.java_websocket.exceptions.WebsocketNotConnectedException
```

It looks from `WebsocketNotConnectedException` that the `ConnectionManager` is still attempting to send on a closed connection.

paddybyers:
Related (internal):
https://support.ably.io/a/tickets/102157
https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/conversation/23324733872#part_id=comment-23324733872-3592513662

tomczoink:
<div>Agent Tom Camp linked Freshdesk ticket <a href="https://ably.freshdesk.com/a/tickets/102157">102157 </a>for this issue</div>

paddybyers:
Related (internal): https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/conversation/23402073164#part_id=comment-23402073164-3667214320

```
Found one Java-level deadlock:
=============================
"Timer-3":
  waiting to lock monitor 0x64403278 (object 0x6a70c150, a io.ably.lib.transport.ConnectionManager),
  which is held by "Thread-16"
"Thread-16":
  waiting to lock monitor 0x64402e18 (object 0x6a70c238, a io.ably.lib.realtime.Channel),
  which is held by "Timer-3"

Java stack information for the threads listed above:
===================================================
"Timer-3":
	at io.ably.lib.transport.ConnectionManager.send(ConnectionManager.java:1070)
	- waiting to lock <0x6a70c150> (a io.ably.lib.transport.ConnectionManager)
	at io.ably.lib.realtime.Channel.attachImpl(Channel.java:145)
	at io.ably.lib.realtime.Channel.attachWithTimeout(Channel.java:318)
[...]

PR #497, opened 2019-09-27
Connectionmanager deadlock fix

paddybyers:
Related: https://github.com/ably/ably-java/issues/495#issuecomment-527105942

This particular deadlock is a result of connection state change notifications being made whilst holding the connectiomanager lock. This PR ensures that all state changes are enacted in the CM thread without holding the lock; any transitions that previously occurred whilst in the process of making another transition are now effectively queued as further requested state changes, and dealt with the CM thread after having finished with the present transition.

@amsurana are you able please to re-test the intermittent connectivity scenario with this branch?


mattheworiordan:
@paddybyers what's the status on this?

paddybyers:
> what's the status on this?

It was stalled because I was looking at a regression, but it is now released.

Issue #505, opened 2019-10-28
Upload to Maven Central


paddybyers:
We need the library and artifacts to be available on Maven Central.

We need to review what exists (https://github.com/ably/ably-java/blob/develop/java/maven.gradle) and decide the best way to enable maven deploys.

mattheworiordan:
@paddybyers shouldn't this issue mention:

> Finally the ws client library we're using has been resurrected and is now being maintained: https://github.com/TooTallNate/Java-WebSocket

paddybyers:
Yes, instead of also pushing `io.ably.Java_Websocket` to Maven, we will aim to migrate to use the latest `org.java-websocket.Java_Websocket`, which is already there.

amihaiemil:
@paddybyers @mattheworiordan first things first, you need to create a namespace/groupid for your company. You need to register with Sonatype and open an Issue about that. Check this out:

https://www.youtube.com/watch?v=0gyF17kWMLg&feature=youtu.be (this is part of their official tutorial). 

You need to create a company account in their Jira tracker and open an Issue asking for the groupId **io.ably.***

The account and password you create are important as they will be used as credentials when publishing the artifacts. Later, you'll need to create a GPG key to sign the artifacts, but let's have the groupId first :)

QuintinWillison:
v.1.1.8 was pushed [live](https://search.maven.org/search?q=g:io.ably) by @paddybyers using https://github.com/ably/ably-java/commit/e42bb9733cb33c2c8618d2ff1b2e3b05f43e9a95, which was further refined with https://github.com/ably/ably-java/commit/1ea6746c3d6cc8a86c264c110fc2c2ccaed8df4d and https://github.com/ably/ably-java/commit/d6fb580762cf24f1f56c9705b4bc6c9536e9706a (all via #560).

Thus, while there is other work that could be done, the core issue has been solved.

PR #506, opened 2019-10-28
Initial setup for Maven Central

amihaiemil:
Changes for #505 

* specify Sonatype snapshot and release repos
* specify signing task
* tested with my own sonatype account (I specified my own coordonates and GPG key) -- the release was succesfully transfered to my staging repo, 

image

**TODO**

* Ably needs a Sonatype account and groupId for Maven Central (as specified [here](https://github.com/ably/ably-java/issues/505#issuecomment-546906599)).
* Ably needs a GPG key (see [here](https://central.sonatype.org/pages/working-with-pgp-signatures.html))
* The following stuff needs to be specified in **gradle.properties**
    ```
    signing.keyId=blaId
    signing.password=blabla
    signing.secretKeyRingFile=/path/to//secring.gpg

    nexusUsername=AblyUsername
    nexusPassword=AblyPassword
    ```
    **IMPORTANT** the above configs have to be encrypted before being commited to the Git repo, and decrypted on-the-fly when performing the release.

* integrate Nexus plugin for releasing to Maven Central (so far the config only uploads the artifacts to  the Sonatype Staging Repo from where they have to be "promoted" to Maven Central)

**OTHER**

- I'm not sure what branching convensions are used. Personally, I always work on a separate branch, for each ticket. The branch's name is the ticket's number (e.g. this PR comes from ``amihaiemil:505``).

### DIFF
--- a/android/maven.gradle
+++ b/android/maven.gradle
@@ -5,10 +5,6 @@ def groupId = "io.ably"
 def artifactId = "ably-android"
 def localReleaseDest = "${buildDir}/release/${version}"
 
-task promoteRepo {
-	println("HELLO PROMOTE REPO ANDROID")
-}
-
 /*
  * Task which signs and uploads the Android artifacts to Nexus OSSRH.
  */
@@ -20,11 +16,11 @@ uploadArchives {
 		beforeDeployment { MavenDeployment deployment -> signing.signPom(deployment) }
 
 		repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
-			authentication(userName: ossrhUsername, password: ossrhPassword)
+			authentication(userName: nexusUsername, password: nexusPassword)
 		}
 
 		snapshotRepository(url: "https://oss.sonatype.org/content/repositories/snapshots/") {
-			authentication(userName: ossrhUsername, password: ossrhPassword)
+			authentication(userName: nexusUsername, password: nexusPassword)
 		}
         pom.groupId = groupId
         pom.artifactId = artifactId
--- a/android/maven.gradle
+++ b/android/maven.gradle
@@ -5,6 +5,10 @@ def groupId = "io.ably"
 def artifactId = "ably-android"
 def localReleaseDest = "${buildDir}/release/${version}"
 
+task promoteRepo {
+	println("HELLO PROMOTE REPO ANDROID")
+}
+
 /*
  * Task which signs and uploads the Android artifacts to Nexus OSSRH.
  */
--- a/android/maven.gradle
+++ b/android/maven.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'maven'
 apply plugin: 'signing'
 
-def groupId = "com.amihaiemil"
+def groupId = "io.ably"
 def artifactId = "ably-android"
 def localReleaseDest = "${buildDir}/release/${version}"
 
--- a/android/maven.gradle
+++ b/android/maven.gradle
@@ -6,7 +6,19 @@ def artifactId = "ably-android"
 def localReleaseDest = "${buildDir}/release/${version}"
 
 uploadArchives {
+	signing {
+		sign configurations.archives
+	}
     repositories.mavenDeployer {
+		beforeDeployment { MavenDeployment deployment -> signing.signPom(deployment) }
+
+		repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
+			authentication(userName: ossrhUsername, password: ossrhPassword)
+		}
+
+		snapshotRepository(url: "https://oss.sonatype.org/content/repositories/snapshots/") {
+			authentication(userName: ossrhUsername, password: ossrhPassword)
+		}
         pom.groupId = groupId
         pom.artifactId = artifactId
         pom.version = version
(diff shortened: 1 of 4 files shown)


amihaiemil:
@paddybyers @mattheworiordan This is an initial PR for the Maven Central task. Please see the TODO list. The encryption part is mandatory if you want this process to be automated. As I suggested initially, we can use Rultor for that (read [this](https://www.yegor256.com/2014/08/19/how-to-release-to-maven-central.html)). 

Or if you are aware of other mechanisms, please let me know :)

amihaiemil:
@paddybyers 

> Does this also upload the android release?

Nope, not yet. I'll check on that as well.

amihaiemil:
@paddybyers I've added the configs for android as well. Tested again with my personal coordonates and it seems to work fine, the artefacts are sent to Sonatype:

Screenshot 2019-11-01 at 18 40 12

Again, this was tested with my accound; the **TODO** list above is still open :)

android/maven.gradle
```diff
@@ -54,10 +66,10 @@ uploadArchives {
         }
 
 		// Export to local Maven cache
-		repository(url: repositories.mavenLocal().url)
+		//repository(url: repositories.mavenLocal().url)
```
paddybyers: Why leave this commented-out?

amihaiemil:
@paddybyers 

> Also, is there a way to have more common-sourced between the java and android uploads? There's a lot of duplication.

I'm not sure. To be honest, I'm just starting to work with Gradle, so I don't know all the tricks. But I would leave the build/CI/Release configs as plain as possible, even if it means duplication sometimes... it's easier to understand, I believe. 

> I would prefer that every user uses their own Sonatype account and key, and we enable those user's access to the project as required

I am not sure this is going to work. The groupId has been assigned to the account you created. Other Sonatype users won'd be able to publish under that groupId. I've actually hit that problem -- I was trying to upload with my account but forgot to change the groupId from ``io.ably`` to ``com.amihaiemil`` (the one I own) and so I kept receiving ``Http 400 BAD REQUEST``

paddybyers:
> I am not sure this is going to work. The groupId has been assigned to the account you created. Other Sonatype users won't be able to publish under that groupId. I've actually hit that problem -- I was trying to upload with my account but forgot to change the groupId from `io.ably` to `com.amihaiemil` (the one I own) and so I kept receiving `Http 400 BAD REQUEST`

OK, but when I set up the groupId I was given the option of which sonatype users were initially given access to the groupId. I assume I can add other users.

amihaiemil:
@paddybyers I've added the configuration for [nexus-staging-plugin](https://github.com/Codearte/gradle-nexus-staging-plugin). 

This plugin will cose the Staging Repository from Nexus (created when we upload archive) and promote the artifacts to Maven Central. As per their documentation, it introduces 3 Gradle Tasks:

```
"
closeRepository - closes an open repository with the uploaded artifacts.
There should be just one open repository available in the staging profile
(possible old/broken repositories can be dropped with Nexus GUI)

releaseRepository - releases a closed repository (required to put artifacts to
Maven Central aka The Central Repository)

closeAndReleaseRepository - closes and releases a repository
(an equivalent to closeRepository releaseRepository
"
```

So we can use ``closeAndReleaseRepository``. We should make sure, though, that there is only 1 Staging Repository active, the one one we'd like to promote. 


This is basically it. I need your help with something: can you please check the order in which these plugins are executed? I'm new to Gradle and this is my only concern at the moment: how and when will these task run and will they run in the correct order (first ``uploadArchives`` then ``closeAndReleaseRepository``).

java/maven.gradle
```diff
@@ -6,6 +6,10 @@ def groupId = "io.ably"
 def artifactId = "ably-java"
 def localReleaseDest = "${buildDir}/release/${version}"
 
+task promoteRepo {
+	println("HELLO PROMOTE REPO")
+}
+
```
amihaiemil: @paddybyers I'm trying to write those ``releaseToMaven`` tasks. You see, I have the same task, ``promoteRepo`` declared for both ``andoird`` and ``java`` subprojects.

I'm running ``gradle :java:promoteRepo`` and gradle also runs the android task, when I would expect only the ``java`` one to run. Running ``gradle :android:promoteRepo`` behaves the same.

On the other hand, playing with the existing ``uploadArchives`` task behaves as expected:

* ``gradle uploadArchives`` runs both subprojects and both ``ably-java`` and ``ably-android`` are uploaded to nexus.

* ``gradle :java:uploadArchives`` only uploads ``ably-java``
 
* ``gradle :android:uploadArchives`` only uploads ``ably-android``.

Do you happen to know what am I missing? Or someone from the team who knows gradle better?

paddybyers: > I'm running `gradle :java:promoteRepo` and gradle also runs the android task, when I would expect only the `java` one to run. Running `gradle :android:promoteRepo` behaves the same.

I think that's because what you wrote was the task definition, which will run as part of the gradle build always. That's not the same as executing the task. That seems to be possible as a relatively simple sequence of subtasks via eg https://stackoverflow.com/a/32909428. However, lets focus on getting the existing release task(s) working first.

amihaiemil:
@paddybyers in the latest commit (https://github.com/ably/ably-java/pull/506/commits/0d276c4caa7ae0a83576bbaf0eee7640e951b901) I have changed the username and password variables to ``nexusUsername`` and ``nexusPassword`` so the nexus-staging-plugin can see them. 

I've tested the command ``./gradlew uploarArchives closeAndReleaseRepository`` and it yields the following log: 


image


*HOWEVER*, the release was performed correctly. It went into Nexus Staging Repo, then it was closed and released to Maven Central (I've refreshed the Nexus page and it happened indeed). The Maven Central UI is not yet refreshed, but we can see the artifacts here: https://repo1.maven.org/maven2/io/ably/

QuintinWillison:
I'm going to land this via #560

Issue #474, opened 2019-05-23
Lib is not re-sending pending messages on new transport after a resume


SimonWoolf:
Log posted by a customer [here](https://ably-real-time.slack.com/archives/C030C5YLY/p1558623828295800) (cf [conversation](https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/all/conversations/22156922233)) shows several issues:

1. At one point it disconnects the transport for inactivity at 14:29:08.086Z, 16s after getting a detached from realtime at 14:28:52.151. That shouldn't happen.
1. After it reconnects and resumes, it does not seem to be re-sending messages on the new transport that were pending on the previous transport, as required by RTN15f
1. Related to the previous one: messages called back with a failure due to getting an ack for a higher msgSerial than the pending queue are called back with "Unknown error / 50000 / 500", which could be more informative

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-430) by [Unito](https://www.unito.io)

paddybyers:
The customer log contains some sensitive information so we're not able to attach it here.

However, I suggest that you start by creating a test for https://docs.ably.io/client-lib-development-guide/features/#RTN15f and see what happens, specifically relating to sending pending messages on a reestablished transport.

amihaiemil:
@paddybyers I did a little research and I landed on [RealTimeResumeTest](https://github.com/ably/ably-java/blob/59100ab140c137f385dbb727c14e2643f4eb2cd7/lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java) -- as far as I understand, this is where we should add the test you propsed. Am I right, or at least close to it?  :D

amihaiemil:
@SimonWoolf This ticket is rather old, but maybe you remember about it. I'm new to the project and currently trying to figure out the architecture/concepts etc. From what I understand, by "transport" you actually mean "channel", right?

paddybyers:
@SimonWoolf btw we discussed ^

paddybyers:
The bug is here, clear as day: https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1209

Instead, queued messages should be reattempted, as per https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c7

paddybyers:
[...]

PR #511, opened 2019-10-31
Added test for #474

amihaiemil:
Added test ``RealtimeResumeTest.resume_publish_queue()`` to cover #474 

1. connect a ``sender`` and a ``receiver`` (AblyRealtime instances, same as other tests)
2. ``sender`` sends some messages and we check if the ``receiver`` got them
3. We cut the transport of the ``sender`` (disconnect it)
4. We send another batch of messages, which should be queued because ``sender``'s transport is down.
5. Reconnect the ``sender``
6. Check if the receiver got the second batch of messages. 

Other changes: 
- unlike other tests, we use the term ``sender`` and ``receiver`` instead of ``*tx`` and ``*rx`` (makes more sense to me)
- The assertion's messages should be the negation of what we are asserting, because they will apear as errors if the assertion does not pass. Other tests do not use this message correctly...

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
@@ -4,6 +4,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import io.ably.lib.types.Message;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
@@ -20,6 +21,8 @@ import io.ably.lib.test.common.ParameterizedTest;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
 import io.ably.lib.types.ErrorInfo;
+
+import java.util.List;
 
 public class RealtimeResumeTest extends ParameterizedTest {
 
@@ -73,8 +76,9 @@ public class RealtimeResumeTest extends ParameterizedTest {
 			e.printStackTrace();
 			fail("init0: Unexpected exception instantiating library");
 		} finally {
-			if(ably != null)
+			if(ably != null) {
 				ably.close();
+			}
 		}
 	}
 
@@ -87,7 +91,8 @@ public class RealtimeResumeTest extends ParameterizedTest {
 	 */
 	@Test
 	public void resume_simple() {
-		AblyRealtime ablyRx = null, ablyTx = null;
+		AblyRealtime ablyTx = null;
+		AblyRealtime ablyRx = null;
 		String channelName = "resume_simple";
 		int messageCount = 5;
 		long delay = 200;
@@ -158,10 +163,12 @@ public class RealtimeResumeTest extends ParameterizedTest {
 			e.printStackTrace();
 			fail("init0: Unexpected exception instantiating library");
 		} finally {
-			if(ablyTx != null)
+			if(ablyTx != null) {
 				ablyTx.close();
-			if(ablyRx != null)
+			}
+			if(ablyRx != null) {
 				ablyRx.close();
+			}
 		}
 	}
 
@@ -174,7 +181,8 @@ public class RealtimeResumeTest extends ParameterizedTest {
 	 */
 	@Test
 	public void resume_disconnected() {
-		AblyRealtime ablyRx = null, ablyTx = null;
+		AblyRealtime ablyTx = null;
+		AblyRealtime ablyRx = null;
 		String channelName = "resume_disconnected";
 		int messageCount = 5;
 		long delay = 200;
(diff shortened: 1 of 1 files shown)


amihaiemil:
@paddybyers After going through the documentation and studying the other tests, I came up with this one. To my understanding, this is the case reported in the Issue. The test seems to pass just fine. 

What do you think?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -120,7 +120,7 @@ public void resume_simple() {
 
 			/* wait for the publish callback to be called */
 			ErrorInfo[] errors = msgComplete1.waitFor();
-			assertTrue("Verify success from all message callbacks", errors.length == 0);				
+			assertTrue("Verify success from all message callbacks", errors.length == 0);
```
paddybyers: I assume that your IDE made this change automatically. We do want these changes, but for future reference, please make any such changes (whitespace-only, or style-only) in a separate commit, so we can more easily review the commits that do have code changes.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -524,4 +524,113 @@ public void resume_verify_publish() {
...
+			ClientOptions opts = createOptions(testVars.keys[0].keyStr);
+			receiver = new AblyRealtime(opts);
+			sender = new AblyRealtime(opts);
+
+			/* create and attach channel to send on */
+			final Channel senderChannel = sender.channels.get(channelName);
+			senderChannel.attach();
+			(new ChannelWaiter(senderChannel)).waitFor(ChannelState.attached);
+			assertEquals(
+			    "The sender's channel should be attached!",
```
paddybyers: Just as a minor point of style: we don't need the `!`. Simple, factual, undramatic statements are fine :)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -524,4 +524,113 @@ public void resume_verify_publish() {
...
+			messageWaiter.waitFor(messageCount);
+			assertEquals(
+			    "Did not receive the entire second round of messages (queued)!",
+				messageWaiter.receivedMessages.size(), messageCount
+			);
+		} catch (AblyException e) {
+			e.printStackTrace();
+			fail("init0: Unexpected exception instantiating library");
+		} finally {
+			if(sender != null)
```
paddybyers: As a point of style, I'm not really a fan of a single-line if clause without braces. There are some instances of this in the library and the tests for historical reasons, and we should try to fix them if a file is being modified as part of other work.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -524,4 +524,113 @@ public void resume_verify_publish() {
...
+			/* wait for the publish callback to be called.*/
+			errors = msgComplete2.waitFor();
+			assertTrue(
+			    "Second round of messages (queued) has errors!",
+				errors.length == 0
+			);
+
+			/* wait for the subscription callback to be called */
+			messageWaiter.waitFor(messageCount);
+			assertEquals(
```
paddybyers: In this assertion you're checking that the correct number of messages is received - I think it would have been better to check that the messages received are in fact the same messages as those that were sent. This could be done easily by making the message `name` contain just the index `i`, or `"test_event:" + i` or something like that.

amihaiemil:
@paddybyers I've made the following changes

1. Added braces to all the ``if``s in the file which didn't have them.
2. Declared the 2 AblyRealtime instances separately - I find it a smell just like ``if``s without braces :D 
3. Removed the ``!`` from error messages.
4. Enhanced the last assertion as you asked. 

If you don't mind, I'll make a few PRs with cosmetic/style changes when I have some time. For instance, I noticed there are classes which declare their attributes as the last part of the file. It seems to me the original developers were coming from C and learned Java by writing this library :D

paddybyers:
> If you don't mind, I'll make a few PRs with cosmetic/style changes when I have some time

In general I'm happy with that, but please don't commit effort to a general update without agreeing the nature of the changes first.

> For instance, I noticed there are classes which declare their attributes as the last part of the file. It seems to me the original developers were coming from C and learned Java by writing this library :D

There was a general approach of keeping private methods and field declarations at the bottom of the class, and public API at the top. That's a legitimate choice, I think, even if it's not your preferred choice. That said, it's never that simple, and sometimes overall organisation and readability is improved if related methods are kept together (including both public and private).

amihaiemil:
@paddybyers Thanks for your comments. What happens now? Will this PR be merged? 

About #474, I suggest you close it and open a new one with your thoughts (you said you think the problem is something else). WDYT?

Issue #492, opened 2019-07-12
fromCipherKey does not match spec


tomczoink:
In the Java library there's a method `fromCipherKey`, however according to the spec it should instead be `withCipherKey` (https://docs.ably.io/client-lib-development-guide/features/#TB3). We should ensure the method matches the spec.

paddybyers:
@amihaiemil are you able to look at this please? Check the spec for exactly what needs to be implemented. If an existing method in the library has the wrong name or signature, then it should be kept but annotated with `@deprecated`. Tests should exist for both the original and corrected API.

amihaiemil:
@tomczoink PR #513 has been merged. I'm closing this. Thanks!

PR #513, opened 2019-11-01
ChannelOptions.withCipherKey + tests

amihaiemil:
PR for #492 
Changes:

* Commit 1: moved attributes of ChannelOptions at the top + javadocs
* Commit 2: formatted method ``getCipher()`` (we should avoid inline ``if``s)
* Commit 3: 
    - Added method ``withChiperKey`` as per spec and deprecated ``fromCipherKey``.
    - Added tests.
* Commit 4: cosmetic changes on ``RealtimeCryptoTest``:
     - added braces to ``if``s;
     - renamed ``tx`` and ``rx`` to ``sender`` and ``receiver`` respectively
     - fixed asserts' error messages
     - removed multiple inline variable initializations

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/ChannelOptions.java
+++ b/lib/src/main/java/io/ably/lib/types/ChannelOptions.java
@@ -33,7 +33,37 @@ public class ChannelOptions {
 		}
 	}
 
+	/**
+	 * <b>Deprecated. Use withCipherKey(byte[]) instead.</b><br><br>
+	 * Create ChannelOptions from the given cipher key.
+	 * @param key Byte array cipher key.
+	 * @return Created ChannelOptions.
+	 * @throws AblyException If something goes wrong.
+	 */
+	@Deprecated
 	public static ChannelOptions fromCipherKey(byte[] key) throws AblyException {
+		return withCipherKey(key);
+	}
+
+	/**
+	 * <b>Deprecated. Use withCipherKey(String) instead.</b><br><br>
+	 * Create ChannelOptions from the given cipher key.
+	 * @param base64Key The cipher key as a base64-encoded String,
+	 * @return Created ChannelOptions.
+	 * @throws AblyException If something goes wrong.
+	 */
+	@Deprecated
+	public static ChannelOptions fromCipherKey(String base64Key) throws AblyException {
+		return fromCipherKey(Base64Coder.decode(base64Key));
+	}
+
+	/**
+	 * Create ChannelOptions with the given cipher key.
+	 * @param key Byte array cipher key.
+	 * @return Created ChannelOptions.
+	 * @throws AblyException If something goes wrong.
+	 */
+	public static ChannelOptions withCipherKey(byte[] key) throws AblyException {
 		ChannelOptions options = new ChannelOptions();
 		options.encrypted = true;
 		options.cipherParams = Crypto.getDefaultParams(key);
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/types/ChannelOptions.java
```diff
@@ -5,9 +5,22 @@
...
+	 */
+	private ChannelCipher cipher;
+
+	/**
+	 * Parameters for the cipher.
+	 */
 	public Object cipherParams;
 
+	/**
+	 * Are these options encrypted or not?
```
paddybyers: "Whether or not this ChannelOptions is encrypted"

I don't like a question here.

lib/src/main/java/io/ably/lib/types/ChannelOptions.java
```diff
@@ -22,9 +22,15 @@
 	public boolean encrypted;
 
 	public ChannelCipher getCipher() throws AblyException {
-		if(!encrypted) return null;
-		if(cipher != null) return cipher;
-		return (cipher = Crypto.getCipher(this));
+		if(!this.encrypted) {
+			return null;
```
paddybyers: I agree with the change to add braces, but why the insistence on using `this` ?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeCryptoTest.java
```diff
@@ -589,6 +594,63 @@ public void channel_options_from_cipher_key() {
...
+
+			/* check the correct plaintext recovered from the message */
+			assertTrue("Verify correct plaintext received", messageText.equals(messageWaiter.receivedMessages.get(0).data));
+
+		} catch (AblyException e) {
+			e.printStackTrace();
+			fail("init0: Unexpected exception instantiating library");
+		} finally {
+			if(ably1 != null)
+				ably1.close();
```
paddybyers: Braces here pls

lib/src/main/java/io/ably/lib/types/ChannelOptions.java
amihaiemil: @paddybyers fixed.

lib/src/main/java/io/ably/lib/types/ChannelOptions.java
amihaiemil: @paddybyers I use ``this.`` all the time when dealing with class variables. It just seems cleaner to me and its clear from the start that it is a class attribute. I don't like relying on colours :) 

I also do it automatically... in previous projects, the build would fail if ``this,`` was not used.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeCryptoTest.java
amihaiemil: @paddybyers braces are already added. You commented on the 3rd commit (3482631), but the latest is the 4th (d04cc195a5403b4f88fad4416307c7661119656b) where the braces are added everywhere.

amihaiemil:
@paddybyers please see my answers :)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeCryptoTest.java
paddybyers: ok

lib/src/main/java/io/ably/lib/types/ChannelOptions.java
paddybyers: Personally I don't like that. I prefer using `this` where necessary for disambiguation, but not unconditionally.

amihaiemil:
@paddybyers so can we merge it? :D

paddybyers:
> so can we merge it?

Yes, if it's approved, then it can be merged.

amihaiemil:
@paddybyers I'm just asking because I don't have the rights to merge PRs :) Thanks!

PR #515, opened 2019-11-05
#514 AblyRealtime implements Autocloseable

### LINKED ISSUE #514: AblyRealtime should implement Autocloseable
I noticed all or most of our tests for ``AblyRealtime`` have the ``try-catch-finally`` pattern, becuase the instantiated ``AblyRealtime`` objects need to be closed. I suspect that also in "real" usage, the object has to be closed when its no longer needed.

In order to simplify things, this class should implement [Autocloseable](https://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html) and be used in ``try-with-resources``, which was introduced in Java 7. Then, the code can be simplified a lot for our clients and also in our tests. Instead of this:

```java
try {
    AblyRealtime ably = ...
} catch (AblyException ex) {
    fail(...);
} finally {
    ably.close();
}
```

we will have this:

```java
try (AblyRealtime ably = ...) {
    ...
} // we don't care about the exception anymore
//since the test will fail automatically and log
//the exception anyway, if it is thrown.
```

Note that this change will not affect backwards compatibility!

amihaiemil:
Fixes #514 

AblyRealtime now implements Autocloseable and can be used in ``try-with-resources``.

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -16,8 +16,11 @@ import io.ably.lib.util.Log;
 /**
  * AblyRealtime
  * The top-level class to be instanced for the Ably Realtime library.
+ *
+ * This class implements {@link AutoCloseable} so you can use it in
+ * try-with-resources constructs and have the JDK close it for you.
  */
-public class AblyRealtime extends AblyRest {
+public class AblyRealtime extends AblyRest implements AutoCloseable {
 
 	/**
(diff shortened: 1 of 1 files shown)


mattheworiordan:
LGTM. Should we update tests to reduce noise then where we close?

paddybyers:
> Should we update tests to reduce noise then where we close explicitly?

I'm not sure we should embark on a wholesale update of existing tests but, as with other style changes, we should use this for new tests, and take the opportunity to update existing code when we're making other changes.

amihaiemil:
@paddybyers can you merge this? :D

Issue #516, opened 2019-11-07
Http -> HttpScheduler -> HttpCore


amihaiemil:
We have the classes ``Http``, ``HttpScheduler`` and ``HttpCore``. 

It is my understanding that class ``Http`` is the highest abstraction and the API for making HTTP requests throughout the library. 

``HttpScheduler``'s Java Doc reads:

 ```
 * HttpScheduler schedules HttpCore operations to an Executor, exposing a generic async API.
 *
 * Internal; use Http instead.
```
and ``HttpCore``'s Java Doc reads:

```
HttpCore performs authenticated HTTP synchronously.
Internal; use Http or HttpScheduler instead.
```

So, an instance of ``Http`` encapsulates ``HttpScheduler`` which in turn encapsulates and works with ``HttpCore``. The last 2 are supposed to be "internal" yet they are present and used directly throughout the library, which means the architecture is a bit broken. These 2 classes should be **package-protected** and only usable inside the package ``io.ably.lib.http``. 

I had a look and the refactoring can be done punctually, 1 step at a time (no need to refactor the whole lib at once). 

What do you think? I'd say it's better to have only a visible ``Http`` instance and avoid having more different ways of performing http calls throughout the codebase.

**Comment by amihaiemil** [comment]
For instance, the constructor of ``Http`` could be redesigned to simply take an ``Auth`` and ``ClientOptions`` and instantiate ``HttpCore``, ``AsyncHttpScheduler`` + SynHttpScheduler`` inside. 

``Http`` constructor now:

```java
public Http(AsyncHttpScheduler asyncHttp, SyncHttpScheduler syncHttp) {
    this.asyncHttp = asyncHttp;
    this.syncHttp = syncHttp;
}
```
and current instantiation in ``AblyBase``'s constructor

```java
public AblyBase(ClientOptions options) throws AblyException {
    ...
    auth = new Auth(this, options);
    httpCore = new HttpCore(options, auth);
    http = new Http(new AsyncHttpScheduler(httpCore, options), new SyncHttpScheduler(httpCore));
    ...
}
```

Redesigned ``Http`` constructore:

```java
public Http(Auth auth, ClientOptions options) {
    this.asyncHttp = new AsyncHttpScheduler(
        new HttpCore(options, auth),
        options
    );
    this.syncHttp = new SyncHttpScheduler(
        new HttpCore(options, auth)
    );
}
```

and simplified ``AblyBase`` constructor:

```java
[...]

PR #522, opened 2019-11-15
More http encapsulation

amihaiemil:
For #516 

* made HttpScheduler abstract (as I understand it has to be extended always, in order for the Executor to be povided).

* made AsyncHttpScheduler and SyncHttpScheduler package protected

* refactored ``Http``'s constructor to accept ``HttpCore`` and ``ClientOptions`` from which it will build the encapsulated ``AsyncHttpScheduler`` and ``SyncHttpScheduler`` itself.

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
@@ -1,9 +1,6 @@
 package io.ably.lib.http;
 
-import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
-import io.ably.lib.types.ErrorInfo;
-
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


amihaiemil:
@paddybyers This is the little I could do so far. Next step, but slowly (1 method at a time), I'd like to try and get rid of [HttpHelpers](https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/http/HttpHelpers.java).

This utility class is a clear indicator that HTTP requests are being made "ad-hoc" throughout the library, in places where they should not be made, where the developer hurried or didn't know how to access the ``Http`` instance in use.

paddybyers:
> This utility class is a clear indicator that HTTP requests are being made "ad-hoc" throughout the library, in places where they should not be made, where the developer hurried or didn't know how to access the `Http` instance in use.

I'm sorry, but I don't think you can legitimately say that.

There are various ways that HTTP requests need to be made, and these are explicitly indicated in each of the methods in this class, and those differences exist for valid reasons. Perhaps you're being misled by the class name, but it's not as random and disorganised as you're suggesting.

lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
```diff
@@ -12,23 +9,32 @@
  * A HttpScheduler that uses a thread pool to run HTTP operations.
  */
 class AsyncHttpScheduler extends HttpScheduler<ThreadPoolExecutor> {
-	AsyncHttpScheduler(HttpCore httpCore, ClientOptions options) {
-		super(httpCore, new ThreadPoolExecutor(options.asyncHttpThreadpoolSize, options.asyncHttpThreadpoolSize, KEEP_ALIVE_TIME, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>()));
-		executor.allowsCoreThreadTimeOut();
-	}
+    AsyncHttpScheduler(HttpCore httpCore, ClientOptions options) {
```
paddybyers: Why have you changed from tabs to spaces here?

amihaiemil: @paddybyers so Github renders the code with the correct indentation. For instance, look how [SyncHttpScheduler](https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/http/SyncHttpScheduler.java) is displayed now: the Github web view renders TAB as 8 spaces instead of 4.

amihaiemil:
> I'm sorry, but I don't think you can legitimately say that

@paddybyers well, static utility methods are in general a bad smell, that's what I was trying to say. I mean, requests performed via HttpHelpers should be possible via Http. If they are not, it means class Http needs more work. 

Anyway, I'm not going to change anything if you don't agree :D

QuintinWillison:
We will not be merging this pull request.

Issue #529, opened 2019-11-22
ProGuard warnings emitted by Android build against 1.1.6


QuintinWillison:
On Android,  a customer is seeing v.1.1.6 of ably-java results in below ProGuard error:

```
Warning: org.slf4j.LoggerFactory: can't find referenced class org.slf4j.impl.StaticLoggerBinder
Warning: org.slf4j.LoggerFactory: can't find referenced class org.slf4j.impl.StaticLoggerBinder
Warning: org.slf4j.LoggerFactory: can't find referenced class org.slf4j.impl.StaticLoggerBinder
Warning: org.slf4j.LoggerFactory: can't find referenced class org.slf4j.impl.StaticLoggerBinder
Warning: org.slf4j.LoggerFactory: can't find referenced class org.slf4j.impl.StaticLoggerBinder
Warning: org.slf4j.MDC: can't find referenced class org.slf4j.impl.StaticMDCBinder
Warning: org.slf4j.MDC: can't find referenced class org.slf4j.impl.StaticMDCBinder
Warning: org.slf4j.MDC: can't find referenced class org.slf4j.impl.StaticMDCBinder
Warning: org.slf4j.MDC: can't find referenced class org.slf4j.impl.StaticMDCBinder
Warning: org.slf4j.MarkerFactory: can't find referenced class org.slf4j.impl.StaticMarkerBinder
Warning: org.slf4j.MarkerFactory: can't find referenced class org.slf4j.impl.StaticMarkerBinder
Warning: org.slf4j.MarkerFactory: can't find referenced class org.slf4j.impl.StaticMarkerBinder
Warning: org.slf4j.MarkerFactory: can't find referenced class org.slf4j.impl.StaticMarkerBinder
```

It can be suppressed with a `"-dontwarn org.slf4j.**”` proguard rule.

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-437) by [Unito](https://www.unito.io)

PR #530, opened 2019-11-22
Fix SLF4J proguard warnings #529

QuintinWillison:
PR for #529 
It was 1.4.0 of Java-Websocket that introduced SLF4J in a 'breaking change'.

As a temporary fix to rid us of the proguard warnings, this rolls us back to 1.3.9 of Java-Websocket dependency.

We can re-upgrade to 1.4.0 later when we've got the time to look at the proguard configuration at the same time.


amihaiemil:
@QuintinWillison as a good practice, please always link the Issue which the PR is solving in the description. You mentioned the Issue's number only in the PR's title, which doesn't make any formal link between the two.

Issue #531, opened 2019-11-22
ably-java sometimes failing to decrypt Messages


SimonWoolf:
Customer reports that encrypted messages are occasionally unable to be decrypted:
```
2019-11-22 11:48:20.515 E/ABLY_INTERNAL: decrypt()
2019-11-22 11:48:20.515 E/ABLY_INTERNAL: error:1e00007b:Cipher functions:OPENSSL_internal:WRONG_FINAL_BLOCK_LENGTH
2019-11-22 11:48:20.515 E/ABLY_INTERNAL: Unexpected exception: error:1e00007b:Cipher functions:OPENSSL_internal:WRONG_FINAL_BLOCK_LENGTH
```
It appears to be specific messages only, but those messages can apparently still _sometimes_ be successfully decrypted, but not other times.

Unclear whether the messages in question were published from ably-java or ably-cocoa, but it is one or the other. Also unclear whether the problem is limited to messages published by only one of those client libs, or both. So may be an ably-cocoa bug.

See internal channels for example of a raw message that sometimes can't be decrypted

amihaiemil:
@SimonWoolf Can you please give more details about this Issue? What's the exact scenario when this happens (code flow?). Also, you say:

> See internal channels for example of a raw message that sometimes can't be decrypted

What are the "internal channels"?

SimonWoolf:
> What are the "internal channels"?

Sorry if that was unclear -- I just meant our internal company slack. It's a bug report from a customer, the message can't be made public for data privacy reasons.

Re a test case -- no, one does not exist now, once we figure out what happens well enough to write a test that consistently reproduces the issue we'll be most of the way towards solving it :slightly_smiling_face:

@QuintinWillison is coordinating the response for this; liaise with him :slightly_smiling_face:

PR #534, opened 2019-11-27
More testing of the Java crypto implementation

QuintinWillison:
See ably/ably-cocoa#921 for source of the data in 4f2dd0d.

Tests fail and, while that is uexpected, it is the current state of this branch - where it could either be a problem with my testing method or it could be a bug in the crypto code. Specifically:

- io.ably.lib.test.realtime.`RealtimeCryptoMessageTest`:
  - `encrypt_message_128`
  - `encrypt_message_256`
- io.ably.lib.test.realtime.`RealtimeCryptoTest`:
  - `decodeAppleLibrarySequences`

#531


paddybyers:
See https://github.com/ably/ably-java/tree/531-encryption-investigations-paddy. All test cases after the first fail, because the act of using the `CipherParams` in one test mutates the `iv` for subsequent tests. That shouldn't happen of course, but that's the reason for the failures. In practice clients won't specify an `iv`, which is why we've not seen that issue in the wild.

lib/src/main/java/io/ably/lib/util/Crypto.java
```diff
@@ -111,12 +129,10 @@ public static CipherParams getDefaultParams(String base64Key) {
 	}
 
 	/**
-	 * Obtain default CipherParams using key and algorithm from other CipherParams object
-	 * @param params
-	 * @return
+	 * Package scoped method for unit testing purposes.
```
paddybyers: This is a public API: https://docs.ably.io/client-lib-development-guide/features/#RSE1

lib/src/test/java/io/ably/lib/test/realtime/RealtimeCryptoMessageTest.java
```diff
@@ -37,6 +38,12 @@ public void encrypt_message_256() throws IOException, NoSuchAlgorithmException,
 		testEncrypt(testData, "cipher+aes-256-cbc");
 	}
 
+	@Test
+	public void encryptCocoa() throws IOException, NoSuchAlgorithmException, CloneNotSupportedException, AblyException {
```
paddybyers: Why are these identified as Cocoa fixtures? Aren't they just common fixtures?

lib/src/test/java/io/ably/lib/util/CryptoTest.java
```diff
@@ -63,4 +69,108 @@ public void cipher_params() throws AblyException, NoSuchAlgorithmException {
...
+		writeData(writer, encoded, null);
+
+		writer.name("encrypted");
+		writeData(writer, encoded, "cipher+aes-256-cbc");
+
+		writer.endObject();
+	}
+
+	private static final String BASE64 = "base64";
+	private void writeData(final JsonWriter writer, final byte[] data, final String encoding) throws IOException {
```
paddybyers: Doesn't there need to be something in here that writes the msgpack-formatted message as well?

QuintinWillison: Nothing that I've touched yet that has to read these has needed that property.

QuintinWillison: Nothing that I've touched, so far, has needed that property.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeCryptoMessageTest.java
QuintinWillison: Now gone. Any chance you can revisit this PR, please, @paddybyers ? I'm seeking an approval prior to merge. Thanks.

lib/src/test/java/io/ably/lib/util/CryptoMessageTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+		final CipherParams params = Crypto.getParams(algorithm, fixtureSet.key, fixtureSet.iv);
+		final ChannelOptions options = new ChannelOptions() {{encrypted = true; cipherParams = params;}};
+
+		for(final CryptoTestItem item : testData.items) {
+			final Message plain = item.encoded;
+			final Message encrypted = item.encrypted;
+			assertThat(encrypted.encoding, endsWith(fixtureSet.cipherName + "/base64"));
+
+			// if necessary, remove base64 encoding from plain-'text' message
```
paddybyers: Comment syntax is inconsistent with other changes in this commit. In this repo we've been using `/* .... */`-style comments generally.

QuintinWillison: Thanks. I'm going to leave this as is - it'll get picked up, if necessary, when we get around to conforming code style across the codebase. For now I am happy that, albeit in a minority, my commenting style isn't entirely non-standard even within this codebase:

```
ably-java % git grep '/\* ' | wc -l
    2332
ably-java % git grep '// ' | wc -l
     191
```

QuintinWillison: Issue created: #537

QuintinWillison:
I've squashed a subset of the commits and then merged manually.

PR #521, opened 2019-11-12
[WIP] Adding recovery on delta decode failure

codemerx:
Adding recovery mechanism for delta failures - previous message id mismatch and delta decode algorithm failure.

PR #536, opened 2019-12-03
[WIP] Delta decode failure recovery 

codemerx:
Adding recovery mechanism for delta failures - previous message id mismatch and delta decode algorithm failure. This is an updated version of #521 based on delta decoding implemented as library plugin as per ably/wiki#380 (comment)

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
+++ b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
@@ -653,7 +653,7 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
 		Message lastMessage = messages[messages.length - 1];
 
 		if (firstMessage.extras != null && firstMessage.extras.delta != null && !firstMessage.extras.delta.from.equals(this.lastPayloadMessageId)) {
-			Log.w(TAG, "Delta message decode failure - previous message not available.");
+			Log.e(TAG, String.format("Delta message decode failure - previous message not available. Message id = %s, channel = %s", firstMessage.id, name));
 			this.startDecodeFailureRecovery();
 			return;
 		}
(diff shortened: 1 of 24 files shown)


lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -663,11 +663,22 @@ private void onMessage(ProtocolMessage message) {
 			try {
 				msg.decode(options, decodingContext);
 			} catch (MessageDecodeException e) {
-				Log.e(TAG, String.format("%s on channel %s", e.errorInfo.message, name));
 				if (e.errorInfo.code == 40018) {
+					if(msg.id == null) msg.id = message.id + ':' + i;
```
QuintinWillison: I'm surprised that this needs to be done here - why wasn't this set upstream of this method?

Regardless, if we know it's not set then why have you got the `null` check? Either it needs to be set here for every message or it doesn't. This just seems ambiguous.

codemerx: We don't know whether the realtime would have set this or not. If it has not set it we set it [further down the line in this method](https://github.com/ably/ably-java/pull/536/files/2a0c16897db4d67c80338d7642e585d88f3dc804..10bb8be448672640ee3c20b4ee36269a7da8121b#diff-a122de4484dcb1721e213f32ab337af5R684). This is not actually my design and is not part of the work on this PR. That being the case I do not know all the details as of why it is done this way. @paddybyers could explain this I guess.

What I have done at the line your comment pertains to will only set the message id a bit earlier so that it is available for logging.

QuintinWillison: ok, thanks for explaining. I was just interested as it looked a bit out of place.

paddybyers:
Closed in favour of https://github.com/ably/ably-java/pull/544

PR #543, opened 2020-01-03
Push Activation State Machine: validate an already-registered device on activation

paddybyers:
Implements https://github.com/ably/docs/pull/786

### DIFF
--- a/android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java
+++ b/android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java
@@ -10,6 +10,7 @@ import android.test.AndroidTestCase;
 
 import android.util.Log;
 import com.google.gson.JsonObject;
+import io.ably.lib.http.HttpCore;
 import io.ably.lib.push.*;
 import io.ably.lib.push.ActivationStateMachine.AfterRegistrationSyncFailed;
 import io.ably.lib.push.ActivationStateMachine.CalledActivate;
@@ -60,6 +61,7 @@ import io.ably.lib.util.Serialisation;
 import static io.ably.lib.test.common.Helpers.assertArrayUnorderedEquals;
 import static io.ably.lib.test.common.Helpers.assertInstanceOf;
 import static io.ably.lib.test.common.Helpers.assertSize;
+import static io.ably.lib.util.Serialisation.gson;
 
 public class AndroidPushTest extends AndroidTestCase {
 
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
```diff
@@ -415,12 +412,53 @@ public void execute(HttpScheduler http, Callback<Void> callback) throws AblyExce
...
+					machine.handleEvent(new SyncRegistrationFailed(reason));
+				}
+			});
+		}
+	}
+
+	private static void validateRegistration(final ActivationStateMachine machine) {
+		final LocalDevice device = machine.activationContext.getLocalDevice();
+		boolean useCustomRegistrar = machine.activationContext.getPreferences().getBoolean(ActivationStateMachine.PersistKeys.PUSH_CUSTOM_REGISTRAR, false);
+		if (useCustomRegistrar) {
```
tcard: This isn't in the spec changes. Maybe the new spec should look like RSH3b3a/RSH3b3b?

> (RSH3b3a) If a custom registerCallback was provided to Push#activate, pass it the local DeviceDetails updated with the push details.
> (RSH3b3b) Otherwise, make an asynchronous HTTP POST request to /push/deviceRegistrations using the local DeviceDetails updated with the push details as body.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -1017,6 +1017,11 @@ private static final String hmac(String text, String key) {
 	 * @throws AblyException
 	 */
 	public void setClientId(String clientId) throws AblyException {
+		if(clientId == null) {
+			/* do nothing - we received a token without a clientId */
```
tcard: If this is possible, shouldn't it be in the spec and thus have an associated test?

android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java
```diff
@@ -1214,24 +1242,31 @@ public void run() throws Exception {
...
+						assertEquals("fcm", requestRecipient.getAsJsonPrimitive("transportType").getAsString());
+						assertEquals(updatedRegistrationToken, requestRecipient.getAsJsonPrimitive("registrationToken").getAsString());
+
+						if(isExpectingRegistrationValidation) {
+							// RSH3f1a: PUT the entire DeviceDetails
+							assertEquals("PUT", request.method);
+							assertTrue(requestBody.has("deviceSecret"));
+							assertTrue(requestBody.has("clientId"));
+						} else {
+							// RSH3d3b: PATCH the updated members
```
tcard: So, the deleted assertion for method == PATCH should be moved here instead?
```suggestion
							// RSH3d3b: PATCH the updated members
							assertEquals("PATCH", request.method);
```

paddybyers:
@tcard see also https://github.com/ably/ably-java/pull/543/commits/e70483a451eefc015eda8ed7999920ba7ca86a4b.

This implements https://github.com/ably/docs/commit/1b775f3e1caa9e9e8b6315638c6751f2dc86ffcd

(Tests to follow.)

android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
paddybyers: +1, see https://github.com/ably/docs/pull/786/commits/cc8e7def1b356368b5d8601a1f8c9fad6bdb5145

tcard:
I just want to link to https://github.com/ably/docs/pull/786 for cross-reference in the GitHub UI.

tcard:
I've extracted #553 from this and rebased the rest on top. Once #553 is merged, base should be changed back to master.

tcard:
Screenshot 2020-02-24 at 16 42 54

To-do:

- [x] Test changes to RSH3a2a.
- [ ] Test changes to RSH8*.
- [ ] Spec changes derived from 257a6c1.

tcard:
While talking to @QuintinWillison I noticed that now that we've renamed `AfterRegistrationUpdateFailed` to `AfterRegistrationSyncFailed`, if clients update were to update the library where the previous persistent state was `AfterRegistrationUpdateFailed`, loading the state from persistence would fail, as no such state exists now. We need to make sure we upgrade deprecated states to their new counterparts.

tcard:
Screenshot 2020-02-26 at 17 07 52

tcard:
@paddybyers PTAL

QuintinWillison:
🥳

PR #552, opened 2020-02-09
Fix reference to param that wasn't updated when param name changed.

tcard:


paddybyers:
oops, not sure how that happened

QuintinWillison:
> oops, not sure how that happened

That would have been me, @paddybyers ... sorry!

PR #558, opened 2020-02-13
Channels visibility improvements

QuintinWillison:
The quick win was making the `channels` field on realtime instances `final` - preventing that reference from being trampled.

The larger change was a bit more complex, involving (both) the `Channels` interfaces we have. They were classes but are now interfaces with private implementations, solving the following accessibility issues:

1. **Constructor**: It should never have been accessible, however it was also causing an issue for Kotlin developers because the lack of a `new` keyword makes it less obvious that you're creating an instance of an inner class.
2. **Underlying map implementation**: Allowing anybody in possession of a Channels reference to mutate the channels map without the Ably implementation knowing.
3. **Internal methods**: Called by the connection manager but could equally be called by anybody in possession of a Channels reference without the connection manager knowing.

### Aside

I wanted to push further and deprecate the `channels` fields on both `AblyRealtime` and `AblyBase`, however because `AblyRealtime` is a subclass of `AblyBase` it was not as easy as just creating a `getChannels()` method. The problem is that fields can shadow superclass fields and change their type at the same time, even to an incompatible type, but methods can't. And the `Channels` returned by base are of a completely different class to the `Channels` returned by realtime (due to the fact that we also have two forks of `ChannelBase` upon which each is ultimately reliant!).

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -5,6 +5,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import io.ably.lib.rest.AblyRest;
+import io.ably.lib.transport.ConnectionManager;
 import io.ably.lib.transport.ITransport;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ChannelOptions;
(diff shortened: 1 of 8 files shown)


QuintinWillison:
Yes, kids, [it's real](https://github.com/ably/ably-java/pull/558/commits/64b2bb673d54eb210513b08fe34a19f1ca4b3428). Always, always, always run tests before you push!

QuintinWillison:
Sorry, @paddybyers, I've broken all kinds of tests. Shall fix those and then re-add my request for review once done.

PR #559, opened 2020-02-13
Updating behavior to latest 1.2 spec

codemerx:
`M_ATTACH_RESUME` and RTL4k1 full implementation


QuintinWillison:
Ah, that's better! I've rebased this one against the integration branch and it's reviewable again.

PR #560, opened 2020-02-26
Maven deployment: add task for deploy to staging

paddybyers:
This contains the same in-progress work as https://github.com/ably/ably-java/pull/506, squashed and rebased to the 1.1.8 release.

PR #563, opened 2020-02-27
Pull in what's in the release 1.1.8 branch

QuintinWillison:
This change introduces an event queue for serialisation of all
CM-related event handling. In general, state changes, associated
notifications, and other event processing, are performed without
the CM locked, to avoid the deadlocks that have been prevalent in
earlier releases. However, transport state changes that require
a synchronous change in CM behaviour - that is, transport
disconnections - are handled synchronously so that the transport
is not used when it is no longer viable.


QuintinWillison:
So, @paddybyers, this is what I get once I rebase `release-1.1.8` on to `master` ... clearly it's been merged in to master already at some point. The log lines look like debugging noise.

QuintinWillison:
ok... so I hadn't spotted that another PR is yet to merge to this branch. 🤦‍♂

QuintinWillison:
it was #560 - now pointed to merge to master

PR #576, opened 2020-05-18
Add request ids (RSC7c, TO3p)

QuintinWillison:
Creating this as a draft for the moment as I did the work a while ago so need to assess where it got to at the time. There was a reason that I didn't create a PR at the time but I can't recall what it was.


lib/src/main/java/io/ably/lib/types/Message.java
```diff
@@ -112,7 +112,7 @@ Message readMsgpack(MessageUnpacker unpacker) throws IOException {
 			if(fieldName.equals("name")) {
 				name = unpacker.unpackString();
 			} else if (fieldName == "extras") {
```
Amanda1232, line 114: Our static analysis engine detects that your pull request includes a change from a Comparison of String objects using `==` or `!=`. According to [this description](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#es-comparison-of-string-objects-using-or-es-comparing-strings-with-eq), unless both strings are either constants in a source file, or have been interned using the `String.intern()` method, the same string value may be represented by two different String objects. Consider using the `equals(Object)` method instead.

PR #579, opened 2020-05-26
Add example of subscribing to a channel with deltas

QuintinWillison:
I think this does reveal that the channel options API could be slicker, but we've not got time to be changing it now.

I'm also a little disconcerted, if I'm brutally honest, that the `Message` instances still contain the `extras` telling us "this message is a delta" but still present data that is the final 'rendered' value (i.e. delta applied). This is also something we're not going to be changing right now (I expect) but does feel ugly to me. So, I've written that final paragraph around inspection of the instance to turn this in to a feature! Unless we consider it a bug?

cc/ @SimonWoolf @mattheworiordan

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ channel.subscribe(events, new MessageListener() {
 
 ### Subscribing to a channel in delta mode ###
 
-Subscribing to a channel in delta mode enables delta compression. This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.
+Subscribing to a channel in delta mode enables [delta compression](https://www.ably.io/documentation/realtime/channels/channel-parameters/deltas). This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.
 
 Request a Vcdiff formatted delta stream using channel options when you get the channel:
 
--- a/README.md
+++ b/README.md
@@ -137,6 +137,22 @@ channel.subscribe(events, new MessageListener() {
 });
 ```
 
+### Subscribing to a channel with deltas ###
+
+Request a Vcdiff formatted delta stream using channel options when you get the channel:
+
+```java
+Map<String, String> params = new HashMap<>();
+params.put("delta", "vcdiff");
+ChannelOptions options = new ChannelOptions();
+options.params = params;
+Channel channel = ably.channels.get("test", options);
+```
+
+Beyond specifying channel options, the rest is transparent and requires no further changes to your application. The `message.data` instances that are delivered to your `MessageListener` continue to contain the values that were originally published.
+
+If you would like to inspect the `Message` instances in order to identify whether the `data` they present was rendered from a delta message from Ably then you can see if `extras.getDelta().getFormat()` equals `"vcdiff"`.
+
 ### Publishing to a channel ###
 
 ```java
--- a/README.md
+++ b/README.md
@@ -137,7 +137,9 @@ channel.subscribe(events, new MessageListener() {
 });
 ```
 
-### Subscribing to a channel with deltas ###
+### Subscribing to a channel in delta mode ###
+
+Subscribing to a channel in delta mode enables delta compression. This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.
 
 Request a Vcdiff formatted delta stream using channel options when you get the channel:
 


QuintinWillison:
[internal prover](https://github.com/ably/hello-deltas-java/blob/master/src/main/java/io/ably/examples/deltas/Hello.java)

paddybyers:
> I'm also a little disconcerted, if I'm brutally honest, that the `Message` instances still contain the `extras` telling us "this message is a delta" but still present data that is the final 'rendered' value (i.e. delta applied). This is also something we're not going to be changing right now (I expect) but does feel ugly to me.

Yes, in the same way that the library unwraps the `encoding` steps and updates `encoding` accordingly, then it could also strip the metadata/extras rendered superfluous in that process. I think that's worth raising as a spec question, but it's not going to change this release I think.

mattheworiordan:
> I think that's worth raising as a spec question, but it's not going to change this release I think.

Was a spec question raised?

README.md
```diff
@@ -137,6 +137,22 @@ channel.subscribe(events, new MessageListener() {
 });
 ```
 
+### Subscribing to a channel with deltas ###
+
+Request a Vcdiff formatted delta stream using channel options when you get the channel:
```
mattheworiordan, line 142: I thought we were calling this delta compression?
This description assumes way too much.  

* Does a user know what deltas are?
* Why would they use them?
* What benefit do they get?

@MarkWoulfeAbly can you get involved here perhaps as this extends to how we more generally deal with documentation?

MarkWoulfeAbly: Having spoken with @paddybyers I believe the breakdown is as follows:

Delta Compression - The name of the feature as a whole.
Delta Mode - An attribute to your attachment of a channel.
delta - The specific name of the flag that you change.

So the title could be updated to mention delta compression instead, though I don't know if that gives too much more context. It might make sense to link through to the actual documentation here if more information is needed.

Can I ask where this ReadMe will be visible from? Just GitHub? Do we know who will be getting their information from here as opposed to the documentation?

paddybyers: > Just GitHub?

Yes. But it will be what many - perhaps most - developers use to get started.

MarkWoulfeAbly: I would be interested in understanding a bit more about GitHub as an entry point for exposing users to new features. Alongside the documentation and support help pages that make 3 potential sources of truth which can be an issue. In a case like this, it sounds like we just want to give the developer a taster and can then link through to the full documentation if they want to learn more information as opposed to diving into code examples. cc @JackUkleja for a product perspective on this.

paddybyers: As a developer, it's common to go to Github first, because that's the code is, which is the ultimate source of truth. Personally I find it quite frustrating when libraries on Github don't show you simple "getting started" examples, and you have to go and look elsewhere, or look in the tests, or whatever.

I do think that the code snippets in Github should not be extensive, and they obviously need to point to the main documentation for more information.

mattheworiordan: >  Alongside the documentation and support help pages

No one goes to the support help pages to get documentation do they?  I am not aware of that. I am only aware of us linking people to support pages from our docs which is where the confusion lies.

>  that make 3 potential sources of truth

Sure, but as Paddy has said, developers come to Ably via Github frequently when they download an SDK.  So the key features and code samples is needed, and we should be encouraging people to the canonical docs at www.ably.io/documentation.

mattheworiordan: >  It might make sense to link through to the actual documentation here if more information is needed.

Regardless I do think it's important that whatever content we provide to developers is understandable on the basis that they may know nothing about Ably.

QuintinWillison: @MarkWoulfeAbly @mattheworiordan I'm not really clear what the next step is here. Do you know what we should link to in the docs, as in a specific page? This discussion doesn't seem to have reached a conclusion.

Feel free to suggest new words and/or push a commit or two to this branch. Thanks!

MarkWoulfeAbly, line 142:
```suggestion
### Subscribing to a channel in delta mode ###

Subscribing to a channel in delta mode enables delta compression. This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.

Request a Vcdiff formatted delta stream using channel options when you get the channel:
```

MarkWoulfeAbly: change suggested below

README.md
```diff
@@ -137,6 +137,24 @@ channel.subscribe(events, new MessageListener() {
 });
 ```
 
+### Subscribing to a channel in delta mode ###
+
+Subscribing to a channel in delta mode enables delta compression. This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.
```
MarkWoulfeAbly, line 142:
```suggestion
Subscribing to a channel in delta mode enables [delta compression](https://www.ably.io/documentation/realtime/channels/channel-parameters/deltas). This is a way for a client to subscribe to a channel so that message payloads sent contain only the difference (ie the delta) between the present message and the previous message on the channel.
```

MarkWoulfeAbly:
@mattheworiordan @QuintinWillison I've updated my added text with a link through to where the documentation page will live.

QuintinWillison:
Great, thanks @MarkWoulfeAbly - I'll get this merged and duplicated to:

- https://github.com/ably/ably-js/pull/656
- https://github.com/ably/ably-cocoa/pull/1021
- https://github.com/ably/ably-dotnet/pull/414 (which I notice I need to rebase so ignore the noise if you look immediately)

PR #581, opened 2020-06-10
Support outbound message extras

### LINKED ISSUE #580: Address impact of change to interface on extras field on Message
For [1.2.0](https://github.com/ably/ably-java/releases/tag/v1.2.0) we changed `Message.extras` from `JsonObject` to `MessageExtras`.

This was a breaking change which [we acknowledged as a TODO](https://github.com/ably/ably-java/blob/b53dc120ef67999e6a48700fedf3bc5176eaa20f/lib/src/main/java/io/ably/lib/types/MessageExtras.java#L15) but never addressed.

We need to decide the approach for fixing this (either revert or extend the new `MessageExtras` class) in order to comply with [TM2i](https://docs.ably.io/client-lib-development-guide/features/#TM2i).

QuintinWillison:
Fixes #580. Conforms, IMO, with [TM2i](https://docs.ably.io/client-lib-development-guide/features/#TM2i) (with a little bit of idiomatic artistic license).

Tasks outstanding:

- [x] More `MessageExtras` unit tests (both constructors and message pack)
- [x] Integration test conforming to [RSL6a2](https://docs.ably.io/client-lib-development-guide/features/#RSL6a2) (as, clearly, that didn't exist!)


lib/src/main/java/io/ably/lib/types/MessageExtras.java
```diff
@@ -1,64 +1,131 @@
...
 
-	private final DeltaExtras delta; // may be null
-	
+	/**
+	 * Only intended for use with a MessageExtras instance received from Ably's servers (inbound).
+	 */
+	private final DeltaExtras delta;
+
+	/**
+	 * Only intended for use with MessageExtras instances created to be sent to Ably's servers (outbound).
```
paddybyers, line 29: I'm not sure I understand the comment. Supporting `raw` is for content that isn't understood by this library, and I think that content can go in both directions.

QuintinWillison: Fair enough. I think it's evolved since our discussion later yesterday. I shall refine this commentary.

QuintinWillison:
@paddybyers In terms of supporting `PushExtras` in a strongly typed and compiler verifiable manner, then perhaps in future, yes. For now, **No**.

I know that the customer who is waiting on us fixing this omission is already generating the necessary gson object in their own code - according to this mould/example that they gave us:

```java
JsonUtils.`object`().add(
    "push", JsonUtils.`object`().add(
        "fcm", JsonUtils.`object`()
        // more config here
    )
).toJson()
```

(via [internal Slack message](https://ably-real-time.slack.com/archives/CKAPDADC6/p1591714899111700))

As such, they can keep that same code and just utilise the new constructor I've added to pass 'raw' gson. I'm happy to change that terminology if it's considered incorrect or somehow otherwise misleading. 😁

lib/src/main/java/io/ably/lib/types/MessageExtras.java
QuintinWillison: https://github.com/ably/ably-java/pull/581/commits/c56c859fc133f73c559fa7525d9e4409c95700e2

QuintinWillison:
Back to the drawing board again, I'm afraid.

Some of this suite failing:

```
-> Test crypto_publish[text_protocol](io.ably.lib.test.rest.RestCryptoTest)

io.ably.lib.test.rest.RestSuite > io.ably.lib.test.rest.RestCryptoTest.crypto_publish[text_protocol] FAILED
    java.lang.AssertionError at RestCryptoTest.java:75
-> Test crypto_publish_key_mismatch[text_protocol](io.ably.lib.test.rest.RestCryptoTest)

io.ably.lib.test.rest.RestSuite > io.ably.lib.test.rest.RestCryptoTest.crypto_publish_key_mismatch[text_protocol] FAILED
    java.lang.AssertionError at RestCryptoTest.java:207
-> Test crypto_send_encrypted_unhandled[text_protocol](io.ably.lib.test.rest.RestCryptoTest)

io.ably.lib.test.rest.RestSuite > io.ably.lib.test.rest.RestCryptoTest.crypto_send_encrypted_unhandled[text_protocol] FAILED
    java.lang.AssertionError at RestCryptoTest.java:288
-> Test crypto_publish_256[text_protocol](io.ably.lib.test.rest.RestCryptoTest)

io.ably.lib.test.rest.RestSuite > io.ably.lib.test.rest.RestCryptoTest.crypto_publish_256[text_protocol] FAILED
    java.lang.AssertionError at RestCryptoTest.java:123
-> Test crypto_publish_alt[text_protocol](io.ably.lib.test.rest.RestCryptoTest)

io.ably.lib.test.rest.RestSuite > io.ably.lib.test.rest.RestCryptoTest.crypto_publish_alt[text_protocol] FAILED
    java.lang.AssertionError at RestCryptoTest.java:169

```

lib/src/main/java/io/ably/lib/types/MessageExtras.java
```diff
@@ -1,64 +1,150 @@
...
+	 */
+	public MessageExtras(final JsonObject raw) {
+		this.raw = raw;
+		delta = null;
+	}
+
+	/**
+	 * @since 1.2.0
+	 */
 	public MessageExtras(final DeltaExtras delta) {
```
paddybyers, line 43: I'm still not clear on how this API is going to grow to accommodate more known `extras` members. We're going to have `push`, `metadata`, etc. How would you construct a `MessageExtras` if you have multiple of those values to supply?

The kind of thing you're seeing in other languages now is where the type for each of the known members implements a specific interface, eg

```
interface MessageExtra {}

class DeltaExtras implements MessageExtra {
  ...
}

class PushExtras implements MessageExtra {
  ...
}

public MessageExtras(JSONObject raw, MessageExtra ...members) { .... }
```

lib/src/main/java/io/ably/lib/types/MessageExtras.java
```diff
@@ -1,64 +1,150 @@
...
+	}
+
 	public DeltaExtras getDelta() {
 		return delta;
 	}
 
-	/* package private */ JsonElement toJsonElement() {
-		return Serialisation.gson.toJsonTree(this);
+	/* package private */ JsonObject getRaw() {
+		return raw;
```
paddybyers, line 58: Doesn't this need to be `public` ? If someone validly sends me a message containing a member in `extras` that this library doesn't know about, then I'm going to have to process the raw JSON.

If we are going to make it public, then perhaps it needs to be `asJSON()` or something like that.

QuintinWillison: Happy to make that public, yes. I was just trying to minimise exposure but I can see that's a clearly valid use case. I agree that, in that case, the name would need to change too.

lib/src/main/java/io/ably/lib/types/MessageExtras.java
QuintinWillison: Interesting. Depending on pressure we're under to release 1.2.1 it would definitely be worth having a call to discuss this [as you suggested elsewhere](https://ably-real-time.slack.com/archives/DPY0JSV6E/p1591911561044500).

The main thing I don't want to do is make any breaking changes between 1.2.0 and 1.2.1, but what you propose merely widens the type acceptance of this method so I think that's ok.

lib/src/main/java/io/ably/lib/types/MessageExtras.java
QuintinWillison: I've now made that method public: https://github.com/ably/ably-java/pull/581/commits/e41c806d707f77e6755d19c1eb8555aa0688ef0a

lib/src/main/java/io/ably/lib/types/MessageExtras.java
QuintinWillison: @paddybyers I'm currently disinclined to do anything about this in order to get this fix out... we can think how to architect this properly outside of the scope of this PR, if that's ok with you?

In terms of what I [mentioned above](https://github.com/ably/ably-java/pull/581#discussion_r439088702)... If we added a super interface at some point in the future to `DeltaExtras` then this method could be safely refactored to accept any object that implements that interface without that being a breaking API change.

paddybyers: > `public MessageExtras(final DeltaExtras delta)`

who will use this API?

QuintinWillison: For some reason or another it looks like it's only [used by RealtimeDeltaDecoderTest](https://github.com/ably/ably-java/blob/b53dc120ef67999e6a48700fedf3bc5176eaa20f/lib/src/test/java/io/ably/lib/test/realtime/RealtimeDeltaDecoderTest.java#L148) at the moment. I'll have a look as to whether that's needed. Perhaps I can replace it with a mock.

Meanwhile, of course, do we simply ignore the fact that we exposed this for 1.2.0? Maybe the best option is to simply get rid of it from the public API. What do you think, @paddybyers?

(equally I guess you could follow that argument through and just revert this portion of [the change](https://github.com/ably/ably-java/commit/2b5c3a9ed0102d9b12803516ce8bf90ad15fa4d4) to the type of message `extras` wholesale!)

QuintinWillison: I've removed that API: https://github.com/ably/ably-java/pull/581/commits/d6ff228bfb8c4c939ed63d5cf1f699092c8ed143

marto83:
@QuintinWillison I like the idea of keeping an internal `JsonObject` inside MessageExtras and reading Deltas out of it. 

I think we can improve the way we hydrate the DeltaExtras object. 
DeltasExtras is read-only so we can simplify the logic of that class. You don't need to have `/* package private */ void write(MessagePacker packer) throws IOException`. You always populate it out of the json object. 

MessageExtras can become simpler as well. In the constructor you will just take the json and parse it to hydrate the deltas object. No need to have different parsing for json and message pack. You can use the same logic you have when parsing the json. Something like this: 
	   
    public MessageExtras(final JsonObject jsonObject) {
		if (null == jsonObject) {
			this.jsonObject = new JsonObject();
		} else {
			this.jsonObject = jsonObject;
		}
		parseJson();
	}

	private void parseJson() {
		final JsonElement deltaElement = jsonObject.get(DELTA);
		if (deltaElement instanceof JsonObject) {
			delta = DeltaExtras.read((JsonObject)deltaElement);
		} else {
			if (null != deltaElement) {
				// Not sure if we should throw an exception here 
				throw MessageDecodeException.fromDescription("The value under the delta key is of the wrong type \"" + deltaElement.getClass() + "\" when expected a map.");
			}
		}
	}

This way we can easily add more logic there to extract other read-only or read-wrtie objects. Objects that will support writing to the json will have to have an instance of the internal json object passed to them.

The next thing to note is that we should always return a copy of the jsonObject because it is mutable and we don't want people to be changing it under our feet. If somebody wants to update the MessageExtras of a Message they should use the constructor.

QuintinWillison:
Thanks @marto83 ... if this is all implementation detail then I'm inclined to leave things as they are so we can get this release out. I'm happy to work on internal refactoring later and am also happy for you to take an alternative approach when you make the same change to the [.NET client library](https://github.com/ably/ably-dotnet).

marto83:
I agree, they are all internal details. I'm happy with the external interface.

PR #607, opened 2020-09-29
Fix Member Presence

### LINKED ISSUE #467: Presence blocking get sometimes has missing members
From customer (see ticket https://support.ably.io/a/tickets/97899) 

> Intermittently during testing (and pretty consistently once we started detaching after a Channel.presence.get(true) blocking call returned), we saw presence incorrectly omit attached clients (i.e. presence would report no clients connected, but our subsequent messages on the channel were successfully delivered). The issue was not observed during debugging, suggesting a race condition. While I expect we could get around this issue by explicitly attaching to the channel earlier, calling presence on it when necessary, and explicitly detaching later we'd rather not remain attached to the channel longer than necessary and, in any event, it'd be nice if that race condition were resolved (if that is what's going on).

sacOO7:
- Updated wait logic during presence sync
- Added check for the attached channel state while requesting for members
- Added tests to make sure, the client returns the right members during intermittent detach cycles
- Updated a few other tests
- Fixes #467

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -3,7 +3,6 @@ package io.ably.lib.realtime;
 import io.ably.lib.http.BasePaginatedQuery;
 import io.ably.lib.http.HttpCore;
 import io.ably.lib.http.HttpUtils;
-import io.ably.lib.http.PaginatedQuery;
 import io.ably.lib.transport.ConnectionManager;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.AsyncPaginatedResult;
@@ -15,8 +14,16 @@ import io.ably.lib.types.PresenceMessage;
 import io.ably.lib.types.PresenceSerializer;
 import io.ably.lib.types.ProtocolMessage;
 import io.ably.lib.util.Log;
-
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.EnumMap;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * A class that provides access to presence operations and state for the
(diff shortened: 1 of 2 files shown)


QuintinWillison:
Conflicts against main now on this one, @sacOO7 ... likely as a result of my recent changes in main. They were necessary however to get us wholesale moved over from tabs to spaces. Plus we now have Checkstyle! 🎉  (early days, only a few rules, but from acorns grow oak trees!)

sacOO7:
Hi @QuintinWillison I will fix the conflicts and make PR again

lib/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -792,11 +792,15 @@ void setSuspended(ErrorInfo reason) {
...
-            boolean syncIsComplete = false; /* temporary variable to avoid potential race conditions */
-            while((channel.state == ChannelState.attached || channel.state == ChannelState.attaching) &&
-                    /* = (and not ==) is intentional */
-                    !(syncIsComplete = (!syncInProgress && syncComplete)))
+            boolean syncIsComplete = false;    /* temporary variable to avoid potential race conditions */
+            while (channel.state == ChannelState.attaching) {
                 wait();
+            }
+            if (channel.state == ChannelState.attached) {
+                while (!(syncIsComplete = (!syncInProgress && syncComplete))) {
```
QuintinWillison, line 800: I realise you've carried this code over from work that was done before, but I'm really not comfortable with this from a readability standpoint. Could we make this code more readable? It's completely acceptable for it to span more lines if that makes it easier to understand! I don't, for example, think it would pass [InnerAssignment](https://checkstyle.sourceforge.io/config_coding.html#InnerAssignment) as/when we switch that rule on.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1508,7 +1537,7 @@ public void onPresenceMessage(PresenceMessage message) {
             leavePresenceWaiter.waitFor(ably1.options.clientId, Action.leave);
 
             /* Validate that,
-             *	- we received all actions
+             *- we received all actions
```
QuintinWillison, line 1540: This change (and a few others) add noise. And, actually, in this case I think reduces readability of the comment.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -2598,7 +2627,7 @@ public void presence_state_change () {
         try {
             DebugOptions opts = new DebugOptions(testVars.keys[0].keyStr);
             fillInOptions(opts);
-            opts.autoConnect = false;	/* to queue presence messages */
+            opts.autoConnect = false;  /* to queue presence messages */
```
QuintinWillison, line 2630: why this change? It's unrelated to the fix you've made as far as I can see. Same goes for one below. Just adds noise for reviewers. Please try to keep PRs clean. 🙂

QuintinWillison (changes requested):

lib/src/main/java/io/ably/lib/realtime/Presence.java
sacOO7: I believe it's more readable than before.
I did try to make it more simplified, but ```syncIsComplete``` variable ( mentioned to avoid potential race conditions), makes it as little difficult to read. I don't know the original intention of the variable used, so I tried not to change the logic much.
Also, it's easier to understand the code, if you go through the whole waitforsync() method from the code.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
sacOO7: I don't usually add unnecessary code. I think this was added as a part of ```checkstyle``` run.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
sacOO7: Checkstyle related change

lib/src/main/java/io/ably/lib/realtime/Presence.java
sacOO7: But, I will see if I can simplify it more

sacOO7: I have updated with simplified implementation

vzhikserg (approved):

PR #611, opened 2020-10-01
Remove redundant semicolons

vzhikserg:
The removed semicolons are redundant and it is better to remove them

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -43,7 +43,7 @@ public class Auth {
 	 */
 	public enum AuthMethod {
 		basic,
-		token;
+		token
 	}
 
 	/**
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -43,7 +43,7 @@ public class Auth {
      */
     public enum AuthMethod {
         basic,
-        token;
+        token,
     }
 
     /**
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -10,5 +10,9 @@
 
 	<module name="TreeWalker">
 		<module name="RedundantModifier"/>
+		<module name="UnnecessarySemicolonAfterOuterTypeDeclaration"/>
+		<module name="UnnecessarySemicolonAfterTypeMemberDeclaration"/>
+		<module name="UnnecessarySemicolonInEnumeration"/>
+		<module name="UnnecessarySemicolonInTryWithResources"/>
 	</module>
 </module>
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 checkstyle {
-	toolVersion '8.36.2'
+    toolVersion '8.36.2'
 }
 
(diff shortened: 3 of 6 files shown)


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -43,7 +43,7 @@
 	 */
 	public enum AuthMethod {
 		basic,
-		token;
+		token
```
QuintinWillison, line 46: Ideally I would use a comma here, for the benefit of future diffs.
```suggestion
		token,
```

QuintinWillison (changes requested):
Thanks @vzhikserg ... now that we've got Checkstyle in the main branch, do you fancy seeing if you can add a rule to enforce this going forwards?

Also, one minor observation.

vzhikserg:
@QuintinWillison The changes from the main branch were merged and I've added the checkstyle modules. I had to specify the latest version of the checkstyle version for these modules.

QuintinWillison:
@vzhikserg, thanks. We prefer to rebase rather than produce a merge commit so could you please force push up a clean, linear commit history to your branch?

Also, I had seen various Checkstyle Gradle configurations specify the `toolVersion` but hadn't seen an explicit reason why I needed to, at this stage, for this project. I'm going to have a play now from my end as I want to understand the implications.

QuintinWillison:
ok, so for the record, if I run `./gradlew clean checkstyleMain --info` locally then I see a line stating:

```
[ant:checkstyle] Running Checkstyle 8.12 on 86 files
```

We need a minimum of Checkstyle 8.31 for the rules you've added.

My environment:

```
ably-java % ./gradlew --version

------------------------------------------------------------
Gradle 5.2.1
------------------------------------------------------------

Build time:   2019-02-08 19:00:10 UTC
Revision:     f02764e074c32ee8851a4e1877dd1fea8ffb7183

Kotlin DSL:   1.1.3
Kotlin:       1.3.20
Groovy:       2.5.4
Ant:          Apache Ant(TM) version 1.9.13 compiled on July 10 2018
JVM:          1.8.0_241 (Oracle Corporation 25.241-b07)
OS:           Mac OS X 10.15.5 x86_64
```

Given the Checkstyle plugin is bundled with Gradle, I'm going to look to upgrade our Gradle wrapper as my preferred route to get access to rules requiring Checkstyle version 8.31 or higher.

There is no need, in my opinion, to explicitly specify the Checkstyle tool version if we can just upgrade Gradle to get these rules. I will produce a PR for this which I'll land in `main` before we land this one.

So, please bear with me while I get that done.

QuintinWillison:
@vzhikserg could you please:

1. Remove the explicit Checkstyle tool version configuration now that we've upgraded Gradle.
2. Rebase as described in [my previous comment](https://github.com/ably/ably-java/pull/611#issuecomment-702584326).

Thanks. 😁

QuintinWillison:
That close was automatic. Let me fix. Sorry, was 'cause I changed target branch.

QuintinWillison:
@vzhikserg GitHub will not allow me to reopen this PR, nor will it let me change the target branch back to `main` which would make the concept of re-opening make sense again. Grr!

Any chance you can open a new PR, please? Sorry - I didn't realise this would happen if I did things in the order I just did. We're all learning, all the time. 🤔

PR #618, opened 2020-10-04
Defaults: Generate environment fallbacks

### LINKED ISSUE #603: Defaults: Generate environment fallbacks
Per `RSC15g`, introduced in https://github.com/ably/docs/pull/965.

See https://github.com/ably/ably-js/pull/682 and https://github.com/ably/ably-ruby/pull/196 for reference implementations.

vzhikserg:
Fixes #603


QuintinWillison (changes requested):
Tests look sensible, though I think you could be clearer about the array normalisation you're doing for the benefit of those looking in future... perhaps even consider comparing as sets. And, I guess, consider testing for when the set of fallbacks is exhausted(per [RSC15a](https://docs.ably.io/client-lib-development-guide/features/#RSC15a)).

Could you please try moving these tests to the same package (within `src/test`) as the class under test (`io.ably.lib.transport`) as they look like perfect candidates for "pure unit testing", needing no interaction with external systems.

QuintinWillison:
I've asked @sacOO7 to also review this PR as it's critical to the library's operation... multiple spec points were modified and added in the [originating PR](https://github.com/ably/docs/pull/965).

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -61,8 +61,15 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
         }
         fallbackHostsUseDefault = options.fallbackHostsUseDefault;
         if (options.fallbackHosts == null) {
-            fallbackHosts = Arrays.copyOf(Defaults.HOST_FALLBACKS, Defaults.HOST_FALLBACKS.length);
-            fallbackHostsIsDefault = true;
+            if (!fallbackHostsUseDefault && options.environment != null && !options.environment.equalsIgnoreCase("production")) {
```
sacOO7, line 64: Also need to add check if port and tlsport are set in options
https://github.com/ably/ably-js/blob/8ba626804f68e262c81d9304fca9a05db1a5d7c6/common/lib/util/defaults.js#L138

sacOO7: Also, https://github.com/ably/ably-ruby/pull/196#issue-378182770

sacOO7:
Actually, I was going through the source of truth for the PR (https://github.com/ably/ably-js/pull/682). It seems some of the code is inconsistent with the spec for fallbacks.

vzhikserg:
@QuintinWillison I moved the Hosts tests under the unit tests. These tests were also refactored a bit to remove some boilerplate code.
@sacOO7 I could not find any info about port ports and this implementation should reflect the specification. The [pull request](https://github.com/ably/ably-ruby/pull/196#issue-378182770) you mentioned is not even merged yet. Please check if this implementation follows the existing specification.

sacOO7:
Hi @vzhikserg as I mentioned https://github.com/ably/ably-js/blob/8ba626804f68e262c81d9304fca9a05db1a5d7c6/common/lib/util/defaults.js#L138
This implementation is already a part of the js client.
And ruby mentions about implementing the same (if you noticed).
I agree we have some discrepancies between spec and client implementations, but we surely don't want to deviate more from other client implementations.

sacOO7:
I add links to code on master branch (in case it's easier to understand)
https://github.com/ably/ably-js/blob/8ba626804f68e262c81d9304fca9a05db1a5d7c6/common/lib/util/defaults.js#L138

QuintinWillison:
This is valuable discussion so thanks to both @vzhikserg and @sacOO7 for your attention to detail and continue tenacity! 😁 

We'll get to the bottom of this, but bear in mind that @mattheworiordan's [changes to `ably-ruby`](https://github.com/ably/ably-ruby/pull/196) (for now ignoring the fact that they're yet to be merged) may well "win over" as he originated this requirement and modelled it originally in Ruby to illustrate the requirement for the wider team.

jamienewcomb:
@vzhikserg this ready for a further review now?

vzhikserg:
@jamienewcomb I've added a check for port and tls port, but it is hard to compare the implementations between different libraries. @QuintinWillison @sacOO7 please check the implementation and let me know if anything is missing.

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -60,9 +60,19 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
             setPrimaryHost(defaultHost);
         }
         fallbackHostsUseDefault = options.fallbackHostsUseDefault;
-        if (options.fallbackHosts == null) {
-            fallbackHosts = Arrays.copyOf(Defaults.HOST_FALLBACKS, Defaults.HOST_FALLBACKS.length);
-            fallbackHostsIsDefault = true;
+        if (options.tlsPort != 0 || options.port != 0) {
```
sacOO7, line 63: Is default value for tlsPort & port 0 🤔 ? Ideally, it should be either null or standard values.

sacOO7: Generally, port 0 is reserved, so it shouldn't cause any problem...maybe we can add a minor check for port values to be positive and between 1 to 65535 ( as a part of client-side error handling )

vzhikserg: I agree that it would be better to have null as the default value, but this is how it was defined [in the code](https://github.com/ably/ably-java/blob/b69cfe70d403b022546cea6123528552319e24b5/lib/src/main/java/io/ably/lib/types/ClientOptions.java#L78). If I change it now then it can break code in clients' libraries.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,215 @@
...
+    @Test
+    public void hosts_fallback() throws AblyException {
+        ClientOptions options = new ClientOptions();
+        Hosts hosts = new Hosts(Defaults.HOST_REALTIME, Defaults.HOST_REALTIME, options);
+        String host = hosts.getFallback(Defaults.HOST_REALTIME);
+        /* Expect given fallback host string is (relatively) valid */
+        assertThat(host, is(not(isEmptyOrNullString())));
+        /* Expect multiple calls will provide different (relatively) valid fallback hosts */
+        String host2 = hosts.getFallback(host);
+        assertThat(host2, is(not(allOf(isEmptyOrNullString(), equalTo(host)))));
```
sacOO7, line 36: Can we make assertions more simple to read

sacOO7: I think it's okay to have multiline assertions

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,215 @@
...
+        /* Expect a null, when we requested more than available fallback hosts */
+        for (int i = Defaults.HOST_FALLBACKS.length - 1; i > 0; i--) {
+            assertThat(host2, is(not(equalTo(null))));
```
sacOO7, line 39: Can be made more easier to read using Assert.assertNotNull(host2);

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * Expect a null, when we provide empty array of fallback hosts
+     */
+    @Test
+    public void hosts_fallback_empty_array() throws AblyException {
+        ClientOptions options = new ClientOptions();
+        options.fallbackHosts = new String[]{};
+        Hosts hosts = new Hosts(Defaults.HOST_REALTIME, Defaults.HOST_REALTIME, options);
+        String host = hosts.getFallback(Defaults.HOST_REALTIME);
+        assertThat(host, is(equalTo(null)));
```
sacOO7, line 63: Assert.assertNull(host);

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -60,9 +60,19 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
...
+            fallbackHosts = new String[]{};
```
sacOO7, line 64: As stated in, https://github.com/ably/ably-ruby/pull/196 we will not use defaultFallbackHosts or EnvFallbackHosts when port and tlsport are set. But, we will set fallbackHosts if passed in options. The current implementation ignores options.fallbackHosts when port and tlsport are set.

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -60,9 +60,19 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
...
+            fallbackHostsIsDefault = false;
+        } else if (options.fallbackHosts == null) {
+            if (!fallbackHostsUseDefault && options.environment != null && !options.environment.equalsIgnoreCase("production")) {
+                /* RSC15g2: If ClientOptions#environment is set to a value other than "production"
+                 * and ClientOptions#fallbackHosts is not set, use the environment fallback hosts */
+                fallbackHosts = Defaults.getEnvironmentFallbackHosts(options.environment);
+                fallbackHostsIsDefault = false;
+            } else {
+                fallbackHosts = Arrays.copyOf(Defaults.HOST_FALLBACKS, Defaults.HOST_FALLBACKS.length);
```
sacOO7, line 73: RSC15b, RTN17b: only use fallback if the hostname has not been overridden.
This might set default fallbackHosts even if a custom host is set. We will not use default fallbackhosts if the custom host is set. This is just my hunch by observing the code. You can check by writing tests.

lib/src/main/java/io/ably/lib/transport/Hosts.java
sacOO7: You might like to take a look at https://github.com/ably/ably-js/pull/689/files for reference

sacOO7:
I feel https://github.com/ably/ably-java/blob/647575906f716598174f0fac3ee6a9fbb7f805d5/lib/src/main/java/io/ably/lib/transport/Hosts.java#L150 is a redundant check. Since we already derived fallbackHosts based on spec, check duplicate efforts. I would like to hear your thoughts @QuintinWillison .

lib/src/main/java/io/ably/lib/transport/Hosts.java
sacOO7: I didn't check the latest PR, based on a mutual understanding on old js PR.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        while ((host = hosts.getFallback(host)) != null){
+            returnedEnvironmentFallbackHosts.add(host);
+        }
+
+        assertThat(returnedEnvironmentFallbackHosts, containsInAnyOrder(Defaults.HOST_FALLBACKS));
+    }
+
+    @Test
+    public void hosts_no_fallback_when_port_is_defined() throws AblyException {
```
sacOO7, line 213: Can we add more tests for ports, 
1. It should return fallback when custom fallbacks are provided. 
2. If the host is provided along with fallbacks, it should return fallbacks.

sacOO7:
The code looks good to me. Maybe we can more warnings/errors for critical cases as per https://github.com/ably/ably-js/pull/689/files# (for conflicting scenarios). I didn't get a chance to go through all tests, but I see a good opportunity to cover more cases.

sacOO7:
Also, being on a safer side, can you add @lmars for the review (He has a more deep context about setting env. fallbacks).

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -59,20 +59,38 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
...
+             * fallbackHostsUseDefault to true */
+            throw AblyException.fromErrorInfo(new ErrorInfo("cannot set both fallbackHosts and fallbackHostsUseDefault options", 40000, 400));
+        }
         fallbackHostsUseDefault = options.fallbackHostsUseDefault;
-        if (options.fallbackHosts == null) {
-            fallbackHosts = Arrays.copyOf(Defaults.HOST_FALLBACKS, Defaults.HOST_FALLBACKS.length);
-            fallbackHostsIsDefault = true;
+        if (primaryHostIsDefault) {
+            if (options.fallbackHosts == null && options.port == 0 && options.tlsPort == 0) {
+                if (!fallbackHostsUseDefault && options.environment != null && !options.environment.equalsIgnoreCase("production")) {
```
lmars, line 71: Given this is inside a block where `primaryHostIsDefault` is `true`, I think this code is redundant (since if the primary host is the default, then `options.environment` is guaranteed not to be set)

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -59,20 +59,38 @@ public Hosts(String primaryHost, String defaultHost, ClientOptions options) thro
...
-            fallbackHostsIsDefault = false;
-            if (options.fallbackHostsUseDefault) {
-                /* TO3k7: It is never valid to configure fallbackHost and set
-                 * fallbackHostsUseDefault to true */
-                throw AblyException.fromErrorInfo(new ErrorInfo("cannot set both fallbackHosts and fallbackHostsUseDefault options", 40000, 400));
+            if (fallbackHostsUseDefault && options.port == 0 && options.tlsPort == 0) {
+                fallbackHosts = Defaults.HOST_FALLBACKS.clone();
+                fallbackHostsIsDefault = true;
+            } else {
+                fallbackHosts = (options.fallbackHosts == null) ? new String[] {} : options.fallbackHosts.clone();
```
lmars, line 89: The check for `options.environment` and subsequently using the environment fallbacks needs to happen here.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,229 @@
...
+    @Test
+    public void hosts_fallback_use_environment() throws AblyException {
+        ClientOptions options = new ClientOptions();
+        options.environment = "sandbox";
+        String[] expectedEnvironmentFallbackHosts = Defaults.getEnvironmentFallbackHosts(options.environment);
+        Hosts hosts = new Hosts(null, Defaults.HOST_REALTIME, options);
+
+        String host = "sandbox-" + Defaults.HOST_REALTIME;
+        List<String> returnedEnvironmentFallbackHosts = new ArrayList<>();
+        while ((host = hosts.getFallback(host)) != null){
```
lmars, line 186: It looks to me like `getFallback(host)` with `host` set to `sandbox-<default-host>` as it is here just returns `null`, and therefore `returnedEnvironmentFallbackHosts` remains empty and this test should fail? Perhaps there needs to be an explicit check that it is not empty?

lib/src/main/java/io/ably/lib/transport/Hosts.java
sacOO7: As per code, there can be a case where ```primaryHostIsDefault``` is ```true``` if non-production env is set.

sacOO7: But I agree ```primaryHostIsDefault``` is always false when ```non-production``` env is set

sacOO7: @vzhikserg I think you should move else block under ```if (primaryHost != null) {``` and remove this redundant check

sacOO7: Idk. I feel there's still a good scope for refactoring to keep things simple. We can separate business logic for calculating a primary host and fallbacks for the respective hosts.

sacOO7:
Also, can you mark ```fallbackHostsUseDefault``` as deprecated as per https://github.com/ably/docs/pull/965
@lmars I am not sure why it's not reflected in [TO3k7](https://docs.ably.io/client-lib-development-guide/features/#TO3k7)

sacOO7:
@vzhikserg I think you can copycat logic as per https://github.com/ably/ably-js/pull/689/files. That way, if we need to remove ```fallbackHostsUseDefault``` in the near future, we can just delete that particular block without touching other parts of the code.

lmars:
@sacOO7 

> I am not sure why it's not reflected in [TO3k7](https://docs.ably.io/client-lib-development-guide/features/#TO3k7)

The docs.ably.io site seems to be out of date, I've asked about that internally, but it is up-to-date on the website:

https://www.ably.io/documentation/client-lib-development-guide/features#TO3k7

vzhikserg:
@lmars @sacOO7 thanks for your valuable feedback. Could you please check the latest changes in this PR?

lib/src/test/java/io/ably/lib/transport/HostsTest.java
vzhikserg: Good catch ;) The business logic was updated and this test should be green

lmars (approved):
LGTM

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -40,39 +40,38 @@
...
         this.defaultHost = defaultHost;
-        if (primaryHost != null) {
+        this.fallbackHostsUseDefault = options.fallbackHostsUseDefault;
+        boolean hasCustomPrimaryHost = primaryHost != null && !primaryHost.equalsIgnoreCase(defaultHost);
+        String[] tempFallbackHosts = options.fallbackHosts;
+        if (options.fallbackHostsUseDefault) {
+            if (options.fallbackHosts != null) {
+                throw AblyException.fromErrorInfo(new ErrorInfo("fallbackHosts and fallbackHostsUseDefault cannot both be set", 40000, 400));
+            }
+            if (options.port != 0 && options.tlsPort != 0) {
```
sacOO7, line 50: if (options.port != 0 || options.tlsPort != 0)

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -40,39 +40,38 @@
...
+                throw AblyException.fromErrorInfo(new ErrorInfo("fallbackHostsUseDefault cannot be set when port or tlsPort are set", 40000, 400));
+            }
+            tempFallbackHosts = Defaults.HOST_FALLBACKS;
+        }
+
+        boolean isProduction = (options.environment == null) || ("production".equalsIgnoreCase(options.environment));
```
sacOO7, line 56: You can also add a check for empty environment string

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -40,39 +40,38 @@
...
+            tempFallbackHosts = Defaults.HOST_FALLBACKS;
+        }
+
+        boolean isProduction = options.environment == null || options.environment.isEmpty() || "production".equalsIgnoreCase(options.environment);
+
+        if (!hasCustomPrimaryHost && tempFallbackHosts == null && options.port == 0 && options.tlsPort == 0) {
+            tempFallbackHosts = isProduction ? Defaults.HOST_FALLBACKS : Defaults.getEnvironmentFallbackHosts(options.environment);
+        }
+
+        if (hasCustomPrimaryHost) {
```
sacOO7, line 62: !hasCustomPrimaryHost

lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -40,39 +40,38 @@
...
-            fallbackHostsIsDefault = false;
-            if (options.fallbackHostsUseDefault) {
-                /* TO3k7: It is never valid to configure fallbackHost and set
-                 * fallbackHostsUseDefault to true */
-                throw AblyException.fromErrorInfo(new ErrorInfo("cannot set both fallbackHosts and fallbackHostsUseDefault options", 40000, 400));
-            }
+            setPrimaryHost(isProduction ? defaultHost : options.environment + "-" + defaultHost);
         }
+
+        fallbackHostsIsDefault = Arrays.equals(Defaults.HOST_FALLBACKS, tempFallbackHosts);
```
sacOO7, line 73: Does this do deepEquals 🤔?

sacOO7:
I have added a few comments, I didn't get a chance to go through tests, but I think you can copycat tests from js client too.

lib/src/main/java/io/ably/lib/transport/Hosts.java
vzhikserg: Nice catch! The code was updated and one unit test was added.

lib/src/main/java/io/ably/lib/transport/Hosts.java
vzhikserg: The check was extended

lib/src/main/java/io/ably/lib/transport/Hosts.java
vzhikserg: Could you please elaborate on why it should be inverted?

lib/src/main/java/io/ably/lib/transport/Hosts.java
vzhikserg: Yes, it will check the number of elements as well as if all of them are equal

lib/src/main/java/io/ably/lib/transport/Hosts.java
sacOO7: Wow, I didn't see line no 59. That explains the logic. It should work now : )
I will mark as resolved.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,271 @@
...
+        host = hosts.getFallback(host);
+        assertThat(host, nullValue());
+    }
+
+    /**
+     * Expect that returned host is contained within default host list
+     * when realtimeHost is non-default and fallbackHostsUseDefault is set
+     */
+    @Test
+    public void hosts_fallback_use_default() throws AblyException {
```
sacOO7, line 138: Can we make the test a bit simple. I think Array. Equals should do the job if it does a deep check. The test looks a bit complex to understand at first look.

sacOO7: Or maybe take out common logic in test helpers. But, if abstraction already exists, then we should reuse it.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -0,0 +1,271 @@
...
+        Hosts hosts = new Hosts(Defaults.HOST_REALTIME, Defaults.HOST_REALTIME, options);
+        String host = hosts.getFallback(Defaults.HOST_REALTIME);
+        /* Expect given fallback host string is (relatively) valid */
+        assertThat(host, not(isEmptyOrNullString()));
+        /* Expect multiple calls will provide different (relatively) valid fallback hosts */
+        String host2 = hosts.getFallback(host);
+        assertThat(host2, not(isEmptyOrNullString()));
+        assertThat(host2, not(host));
+        /* Expect a null, when we requested more than available fallback hosts */
+        for (int i = Defaults.HOST_FALLBACKS.length - 1; i > 0; i--) {
```
sacOO7, line 38: I would generally avoid int iterations as much as possible. 
Can you check if the java version supports?

```for (String host : Defaults.HOST_FALLBACKS) { }```
or you can use collections API for iterations

sacOO7: Also, I am not sure, what this for loop exactly does 🤔

lib/src/test/java/io/ably/lib/transport/HostsTest.java
vzhikserg: There are was a bunch of tests which were there before and it seems that I wrote them, but I've just moved this file. Yes, the logic is more complicated than it should be, but I checked the tests and they look good to me. I would be glad to make these tests nicer but in a separate pull request.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
vzhikserg: There are was a bunch of tests which were there before and it seems that I wrote them, but I've just moved this file. Yes, the logic is more complicated than it should be, but I checked the tests and they look good to me. I would be glad to make these tests nicer but in a separate pull request.

sacOO7: Okay, makes sense now. But, I am afraid if we will have the capacity to raise another PR (just to fix tests). I think it's better if we fix it as a part of this PR ( Till the soup is HOTT 😋). Also, I strongly feel the test file is super relevant with modifications made to the code, so makes sense to modify existing tests by any means necessary 😃

sacOO7: Also, make sure you add missing tests with respect to Js PR.

sachinsh76:
@vzhikserg I had a brief discussion with @jamienewcomb . 
 It's not that I am not happy about the code. 
It's just that I'm not confident about it.

I believe tests should be treated as integral parts of the code. We generally don’t follow patterns where features and related tests are treated differently. As a general practice, they should be kept as a part of the same codebase.
Also, makes it easier for the reviewer to properly review the PR. 

I think It’s better we start treating tests as spec aligned implementation (in current context).
Because that belief changes how we look at the code and implicitly solves most of our problems.

sacOO7:
@vzhikserg I had a brief discussion with @jamienewcomb . 
 It's not that I am not happy about the code. 
It's just that I'm not confident about it.

I believe tests should be treated as integral parts of the code. We generally don’t follow patterns where features and related tests are treated differently. As a general practice, they should be kept as a part of the same codebase.
Also, makes it easier for the reviewer to properly review the PR. 

I think It’s better we start treating tests as spec aligned implementation (in current context).
Because that belief changes how we look at the code and implicitly solves most of our problems.

sacOO7:
Also, I feel we don't need to make lot of code changes. Just need to add/modify few tests.

QuintinWillison:
Thanks for the input, @sacOO7 ... I'm going to land this PR as it clearly moves the product forward, even if the testing isn't perfect to everybody's preferences. Could you please create an issue in this repository, referencing this PR, and summarising the gaps you feel we have in testing for this feature. Thanks!

PR #619, opened 2020-10-04
Make Ably version more robust

vzhikserg:
Some of the tests on my machine were failing due to the incorrect version - "1,2" instead of "1.2". The English locale was added to DecimalFormatter to enforce the proper format.


QuintinWillison (approved):
Great spot. Thanks! 😁

PR #622, opened 2020-10-13
Fix param and return javadoc statements

vzhikserg:
This fix helps to limit the number of warnings displayed while building the Android-specific library

```
...\java\io\ably\lib\types\Callback.java:19: warning - @param argument "reason:" is not a parameter name.
...\java\io\ably\lib\types\Capability.java:29: warning - @param argument "capability:" is not a parameter name.
...\java\io\ably\lib\types\ClientOptions.java:27: warning - @param argument "key:" is not a parameter name.

```


QuintinWillison (approved):
Apologies for the late review, @vzhikserg ... I should have looked at this much sooner. Thanks for doing this.

I'm going to save myself time by not reviewing in depth. But, given it's only commentary changes, I'm going to trust you and assume it improves things. 😁

PR #623, opened 2020-10-20
Add verbose logs in push notification related code

QuintinWillison:
To assist in debugging an active issue one of our customers is having.

### DIFF
--- a/android/src/main/java/io/ably/lib/types/RegistrationToken.java
+++ b/android/src/main/java/io/ably/lib/types/RegistrationToken.java
@@ -33,4 +33,12 @@ public class RegistrationToken {
             return name().toLowerCase();
         }
     }
+
+    @Override
+    public String toString() {
+        return "RegistrationToken{" +
+            "type=" + type +
+            ", token='" + token + '\'' +
+            '}';
+    }
 }
--- a/android/src/main/java/io/ably/lib/platform/Platform.java
+++ b/android/src/main/java/io/ably/lib/platform/Platform.java
@@ -10,6 +10,7 @@ import io.ably.lib.transport.NetworkConnectivity;
 import io.ably.lib.transport.NetworkConnectivity.DelegatedNetworkConnectivity;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ErrorInfo;
+import io.ably.lib.util.Log;
 
 import java.util.WeakHashMap;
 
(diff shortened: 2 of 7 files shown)


QuintinWillison:
@paddybyers could you please approve this with a view to landing this PR so I can get a 1.2.3 release out to include this additional logging in production. To assist with active customer issue.

FYI, @cruickshankpg 😁

android/src/main/java/io/ably/lib/types/RegistrationToken.java
```diff
@@ -33,4 +33,12 @@ public String toName() {
             return name().toLowerCase();
         }
     }
+
+    @Override
```
paddybyers, line 37: We've been using `[ ]` instead of `{ }`: https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/types/Message.java#L103

paddybyers (approved):
LGTM, thanks

android/src/main/java/io/ably/lib/types/RegistrationToken.java
QuintinWillison: Looking at this and from hazy memory, I'm pretty sure I used IntelliJ IDEA's boilerplate via auto-complete to create this method... I'm happy to change and will push a commit to do so. Is this a formal Ably standard, or just convention within this codebase?

QuintinWillison: We have very few `toString()` implementations across this codebase, however those already differ in terms of style. e.g.:

- perhaps more illustrative of the style you point out: [ErrorInfo](https://github.com/ably/ably-java/blob/0e4da0f8ef63524b6fdc72e3d1a80561acf329d1/lib/src/main/java/io/ably/lib/types/ErrorInfo.java#L68)
- more recently illustrative of the style I used: [MessageExtras](https://github.com/ably/ably-java/blob/0e4da0f8ef63524b6fdc72e3d1a80561acf329d1/lib/src/main/java/io/ably/lib/types/MessageExtras.java#L126)

My preference would be to align with IntelliJ IDEA's default, thus leave this PR in current state and address square bracket "contraventions" later. Any reason that's a bad choice, @paddybyers ?

paddybyers: > Any reason that's a bad choice

No, so long as we're consistent I'm fine

PR #624, opened 2020-10-21
Gradle conform and reformat

QuintinWillison:
This is the first eager step I'm taking towards refactoring our Gradle files.

For now I am just enabling the formatting rules so there is quite a lot of noise here in respect of the necessary fixes in respect of indentation and tabs to spaces etc.. The meat lies in the files within `gradle-lint`.

## The Journey

Getting to this point was a little bit arduous. The CodeNarc plugin adds some default tasks for checking source sets (e.g. `codenarcMain`), however I simply could not get them to check anything but files with a `.groovy` extension, even when I categorically knew the source set was capturing the Groovy-based files I wanted (just all with `.gradle` extension).

For this reason the way CodeNarc is integrated here is breaking with convention based upon how most projects use it, in that we have an Ably-specific task called `checkWithCodenarc`. I am completely comfortable with this, however, because I am yet to find an example "in the wild" of anybody using CodeNarc in the way that I am here to introspect on Gradle build script files. Most projects are using it to inspect source code for a library or application that itself has been written in Groovy.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,8 +9,7 @@ charset = utf-8
 indent_style = space
 indent_size = 2
 
-[*.java]
-indent_style = space
+[*.{java,groovy,gradle}]
 indent_size = 4
 
 [*.md]
--- a/README.md
+++ b/README.md
@@ -489,13 +489,28 @@ The Android-specific library AAR is built with:
 
 ## Code Standard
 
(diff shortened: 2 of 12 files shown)


paddybyers (approved):
LGTM, thanks

vzhikserg (approved):

owenpearson (approved):

PR #625, opened 2020-10-21
Adopt more Groovy conventions in Gradle scripts

QuintinWillison:
Still focussed on conforming what we have for now, with an eye towards the next step being the refactoring of our Gradle files.

This establishes a good baseline, following on from #624, on top of which to make that refactor.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,11 +35,11 @@ android {
         targetSdkVersion 24
         versionCode 1
         versionName version
-        setProperty("archivesBaseName", "ably-android-$versionName")
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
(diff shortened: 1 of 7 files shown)


paddybyers (approved):
LGTM, thanks

vzhikserg (approved):
LGTM, I am just curious why the pipeline was not triggered for this pull request (I know that it would be red anyway)?

owenpearson (approved):

QuintinWillison:
Hi @vzhikserg - well spotted. For reasons unknown to me we have Travis configured to only build `main` and branches specifically given the suffix `-ci`. This is not approach I agree with and is something we will change soon. Once I've finished my work on the Gradle side of things then I am either going to work to fix our Travis builds or, perhaps preferable, switch to using GitHub's native CI (workflows / actions).

**TL;DR:** Yup. 😁  It's on our list!

Issue #628, opened 2020-11-03
Add/modify generate environment fallback tests


sacOO7:
- Related to https://github.com/ably/ably-java/pull/618
- Should be covered as per https://github.com/ably/ably-java/pull/618#issuecomment-720498053
- Refactor test logic and make it simple to understand

PR #629, opened 2020-11-04
Refactor unit tests related to hosts and environmental fallbacks

vzhikserg:
Fixes: #628

### DIFF
--- a/lib/src/test/java/io/ably/lib/transport/HostsTest.java
+++ b/lib/src/test/java/io/ably/lib/transport/HostsTest.java
@@ -251,10 +251,7 @@ public class HostsTest {
         options.fallbackHostsUseDefault = true;
 
         // When
-        Hosts hosts = new Hosts(null, Defaults.HOST_REALTIME, options);
-
-        // Then
-        assertThat(hosts.getFallback(Defaults.HOST_REALTIME), nullValue());
+        new Hosts(null, Defaults.HOST_REALTIME, options);
     }
 
     private List<String> collectFallbackHosts(Hosts hosts) {
--- a/lib/src/test/java/io/ably/lib/transport/HostsTest.java
+++ b/lib/src/test/java/io/ably/lib/transport/HostsTest.java
@@ -1,44 +1,37 @@
 package io.ably.lib.transport;
 
+import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.isEmptyOrNullString;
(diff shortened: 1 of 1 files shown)


lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -69,65 +67,52 @@ public void hosts_fallback_empty_array() throws AblyException {
...
-        }
-        assertThat(idx, is(customHosts.length));
-        assertTrue(allFound);
-        assertTrue(shuffled);
+
+        // Then
+        List<String> fallbackHosts = collectFallbackHosts(hosts);
+        assertThat(hosts.getPrimaryHost(), is(Defaults.HOST_REALTIME));
+        // the returned fallback hosts should have the same elements as custom host fallbacks
+        assertThat(fallbackHosts, containsInAnyOrder(customHosts));
```
sacOO7, line 81: Nicely implemented 💙

lib/src/test/java/io/ably/lib/transport/HostsTest.java
```diff
@@ -251,21 +233,36 @@ public void hosts_fallback_when_fallback_hosts_and_port_are_defined() throws Abl
...
-        ClientOptions options = new ClientOptions();
+        // Given
         options.tlsPort = 8081;
         options.fallbackHostsUseDefault = true;
+
+        // When
         Hosts hosts = new Hosts(null, Defaults.HOST_REALTIME, options);
 
+        // Then
         assertThat(hosts.getFallback(Defaults.HOST_REALTIME), nullValue());
```
sacOO7, line 257: It will never reach this statement right

sacOO7:
Btw, I forgot to mention, while naming git branch, we also include issue id, that way it's easy to refer to the exact issue.
Like current branch can be renamed as, ```fix/628-refactor-hosts-unit-tests```
You can start following the convention from next time onwards : )

sacOO7:
Regarding ```private final ClientOptions options = new ClientOptions();```

Just a thought. Normally I would be more careful while using common fields for executing tests. It might create interdependency between tests ( leading to flaky test behavior). **WET** can be preferred over **DRY** in this case : ). But, current tests should work, since it creates ```new HostsTest``` object, every time a new test is executed.

lib/src/test/java/io/ably/lib/transport/HostsTest.java
vzhikserg: That's right! I will remove this obsolete statement

sacOO7 (approved):
LGTM 😃

QuintinWillison (approved):

Issue #621, opened 2020-10-12
Capture context rich error message


sacOO7:
- Current ErrorInfo class just captures the exception messages instead of the whole error context.
- Exception should be fully wrapped while being consumed.
- It shouldn't leave any missing trail behind for captures exceptions.
Link to read about exception handling practices - https://stackify.com/best-practices-exceptions-java/

sacOO7:
Proposed solution 
- Extend ```ErroInfo``` with native ```Exception``` class.
(as stated, ```ErrorInfo``` is an exception type encapsulating error information containing an Ably-specific error code and generic status code, but it's not a java exception ... yet)
- Pass original exception as a **cause** while creating new ```ErrorInfo``` exception.
- Include exception **cause** while logging ErrorInfo object.

mattheworiordan:
I suppose this proposal, this is what I have done in Ruby, see https://github.com/ably/ably-ruby/blob/main/lib/ably/models/error_info.rb#L33-L57 and https://github.com/ably/ably-ruby/blob/main/lib/ably/exceptions.rb#L16

QuintinWillison:
We already have an `AblyException` class which encapsulates an `ErrorInfo` instance, which is an architectural choice - using composition - that had already been made (rightly or wrongly) in the past.

Modifying the inheritance tree above `ErrorInfo` (essentially widening its public interface) feels like a hammer to crack a nut, and is not the right place in the stack to do this.

There is already a [well-known, idiomatic constructor](https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html#Exception(java.lang.String,%20java.lang.Throwable)) on the base `Exception` class allowing us to pass a cause. Therefore this requirement may be solved by simply adding a constructor to our `AblyException` class that calls that.

Also, @sacOO7, can you please provide a breadcrumb link back to the original discussion that prompted you to create this issue so those passing by in future can gain more context as to the requirement. Thanks.

sacOO7:
Yes, I think we already have ```AblyException`` class. I was looking into the use cases where we use ```ErrorInfo`` class and ```AblyException``` derived from ```ErrorInfo```. It seems maybe at some of the places, we are not using ```AblyException``` class effectively (Like the above case). I was also looking into the way exceptions are handled.
https://github.com/ably/ably-java/blob/647575906f716598174f0fac3ee6a9fbb7f805d5/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L279
and 
[...]

Issue #606, opened 2020-09-29
Devices do not reconnect after handshake failure


QuintinWillison:
_We've had a report from a customer, [via Intercom](https://app.intercom.com/a/apps/ua39m1ld/inbox/inbox/conversation/38483300001110). Conversation detail, with redactions:_

CUSTOMER: hello. I get handshake failure:

```
[Ably internal] setState(): setting suspended; reason [ErrorInfo message=Received fatal alert: handshake_failure (See https://help.ably.io/error/80000) code=80000 statusCode=503 href=https://help.ably.io/error/80000] on all my devices being restarted
```

CUSTOMER: they run java 11 under ARMv7

**Jo from Ably**: Hi, sorry to hear you're having issues, could you give me a little more information on how you're using ably?

CUSTOMER: we use the java client

CUSTOMER: version 1.1.10

**Jo from Ably**: 80000 is a network connectivity error, do your clients all have network access?

CUSTOMER: yeah I tried using curl on https, and that went fine

CUSTOMER: java client works on windows as well through java here

**Jo from Ably**: and the connection does not retry after 15 seconds?

CUSTOMER: so it must be some kind of an SSL handshake failure happening only on ARM but I am not sure if you changed the certificate for your web services

CUSTOMER: 

```
[Ably internal] HTTP request: https://internet-up.ably-realtime.com/is-the-internet-up.txt GET
[Ably internal] X-Ably-Lib: java-1.1.10
[Ably internal] Accept: application/json
[Ably internal] X-Ably-Version: 1.0
[Ably internal] HTTP response:
[Ably internal] CF-RAY: REDACTED
[Ably internal] Server: cloudflare
[Ably internal] Connection: keep-alive
[Ably internal] Last-Modified: Wed, 14 Jan 2015 16:12:25 GMT
[Ably internal] x-amz-version-id: null
[Ably internal] Date: Thu, 27 Aug 2020 15:03:41 GMT
[Ably internal] Accept-Ranges: bytes
[Ably internal] CF-Cache-Status: DYNAMIC
[Ably internal] ETag: "REDACTED"
[Ably internal] Set-Cookie: REDACTED; expires=Sat, 26-Sep-20 15:03:40 GMT; path=/; domain=.ably-realtime.com; HttpOnly; SameSite=Lax
[Ably internal] x-amz-request-id: REDACTED
[Ably internal] x-amz-id-2: REDACTED
[Ably internal] Content-Length: 4
[...]

PR #632, opened 2020-11-21
Log error message to get clear understanding of exception

sacOO7:
- Related to #606 
- Based on #621 
- Logged error context before passing it to the client.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -223,7 +223,7 @@ public class WebSocketTransport implements ITransport {
 
         @Override
         public void onError(final Exception e) {
-            Log.e(TAG, "Unexpected exception ", e);
+            Log.e(TAG, "Connection error ", e);
             connectListener.onTransportUnavailable(WebSocketTransport.this, new ErrorInfo(e.getMessage(), 503, 80000));
         }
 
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -223,6 +223,7 @@ public void onClose(final int wsCode, final String wsReason, final boolean remot
 
         @Override
         public void onError(final Exception e) {
+            Log.e(TAG, "Unexpected exception ", e);
```
paddybyers, line 226: This is presumably called if there is any error on the websocket connection, and as such it's not "Unexpected" - it's a routine event. I would have the error message say "Connection error" or something like that.

sacOO7: Agree! But I was hoping, connection-related errors will be logged as a part of the exception message anyway. Also, I found some of the cases where it can throw IO related exceptions (https://github.com/TooTallNate/Java-WebSocket/blob/a76e0524ece10062ca98ffa74e20a9f9d8bbd51d/src/main/java/org/java_websocket/client/WebSocketClient.java#L964)
- But, connection-related errors might occur most of the time, so makes sense to keep it specific & relevant.

paddybyers (approved):
LGTM, thanks

PR #643, opened 2021-01-21
Add workflows

QuintinWillison:
Establishing lightweight foundations.


QuintinWillison:
Failures expected at this stage as this pull request simply moves us off Travis and onto GitHub workflows.

This gives us a more solid foundation to grow from hereon. Going forward our intention is to:

1. Fix failing / flakey tests
2. Increase runtime test environment use beyond workflow defaults (different JRE, Android emulators and perhaps even building and running against JRE on macOS and Windows)
3. Sort out library signing and release process
4. Increase code quality with more checks/lints and unit tests

All of which will need discussion to varying degrees - not only in terms of prioritisation but also in terms of approach.

vzhikserg (approved):
LGTM, thanks!

PR #642, opened 2021-01-14
Fix CI pipeline

vzhikserg:
Fix CI pipeline by ignoring failing tests and adding retry policies for flaky tests.

PR #645, opened 2021-01-29
Add workflow status badges

QuintinWillison:
Hoping to land this once https://github.com/ably/ably-java/pull/642 has been progressed so that checks pass.


vzhikserg (approved):

QuintinWillison:
It is currently expected that the assemble workflow should fail. I am hoping to move onto fixing that soon as part of a Gradle refactor. In the meantime I am comfortable that the badge makes it really clear to all that we have work to do in this area.

PR #646, opened 2021-02-10
Update references from 1 -> l to match client spec

natdempk:
👋 Hopefully a small one. I was looking at the client spec and it appears these should actually be `l` instead of `1`:
* https://www.ably.io/documentation/client-lib-development-guide/features#TO3l3
* https://www.ably.io/documentation/client-lib-development-guide/features#TO3l4


vzhikserg (approved):
Thanks for updating the comments!

sacOO7 (approved):

paddybyers:
Thanks @natdempk

QuintinWillison:
I acknowledge that checks are failing, however that is clearly unrelated to the changes being made by this PR. Thanks, all!

PR #648, opened 2021-02-17
Remove empty capability query parameter

### LINKED ISSUE #647: capability tokendetails adds to HTTP Request as a query parameter
I'm trying to request a token from an authUrl `authUrl = "htttps://sample.com/auth"`
For some reason when initiating the call Ably HTTP call adds `capability` as a query parameter. 
`https://sample.com/auth?capability=`
Is there anyway to remove this query parameter? Thanks

vzhikserg:
Fixes #647 

The check was extended to transparently handle null and empty strings. If the value is null we should also return null and in this case, it shouldn't add this query parameter.

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/Capability.java
+++ b/lib/src/main/java/io/ably/lib/types/Capability.java
@@ -27,7 +27,7 @@ public class Capability {
      * (if for example it is not valid JSON)
      */
     public static String c14n(String capability) throws AblyException {
-        if (capability == null || capability.isEmpty()) return capability;
+        if (capability == null || capability.isEmpty()) return null;
         try {
             JsonObject json = (JsonObject)gsonParser.parse(capability);
             return (new Capability(json)).toString();
--- a/lib/src/main/java/io/ably/lib/types/Capability.java
+++ b/lib/src/main/java/io/ably/lib/types/Capability.java
@@ -19,21 +19,19 @@ import com.google.gson.JsonParser;
 public class Capability {
 
     /**
-     * Convenience method to canonicalise a JSON capability expression
+     * Convenience method to canonicalise a JSON capability expression.
      *
      * @param capability a capability string, which is the JSON text for the capability
      * @return a capability string which has been canonicalised
      * @throws AblyException if there is an error processing the given string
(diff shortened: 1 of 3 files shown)


QuintinWillison (approved):

lib/src/main/java/io/ably/lib/types/Capability.java
```diff
@@ -19,21 +19,19 @@
...
      *
      * @param capability a capability string, which is the JSON text for the capability
      * @return a capability string which has been canonicalised
      * @throws AblyException if there is an error processing the given string
      * (if for example it is not valid JSON)
      */
-    public static final String c14n(String capability) throws AblyException {
-        if(capability == null || capability.isEmpty()) return "";
+    public static String c14n(String capability) throws AblyException {
+        if (capability == null || capability.isEmpty()) return capability;
```
sacOO7, line 30: I feel it should return null when the capability is empty too, doesn't make sense to send empty capability to the server 🤔

QuintinWillison: @sacOO7, can you reference something in the spec or perhaps another client library implementation that does exhibit the behaviour you describe?

sacOO7: Okay, let me check.

sacOO7: https://github.com/ably/ably-go/blob/1df14a2afaad30eaac619fb36a88bafef654f0b8/ably/token.go#L22
says it to be `omitempty`
- Means it will not include the key if the capability is `empty` or `null`  (https://www.sohamkamani.com/golang/omitempty/)

vzhikserg: Updated the code to return null for empty strings

sacOO7 (approved):

lib/src/main/java/io/ably/lib/types/Capability.java
QuintinWillison: Great, thanks both. 😁

sacOO7 (approved):

PR #652, opened 2021-03-01
Unregister ConnectionWaiter listeners once connected

### LINKED ISSUE #651: Many instances of ConnectionWaiter spawned while app is running, with authentication token flow
Reported by a customer using version 1.2.1 of this library:

> There are thousands of instances of ConnectionWaiter.
> ConnectionWaiter is instantiated for each onAuthUpdated(): https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L885
It adds itself as a listener here: https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L465
> But I couldn't find where it was cleaned up / removed?

The customer has also shared us a screenshot:

javass

The customer is suggesting that this behaviour may be new since #539 (for #495). Specifically stating:

> there has always been issues, like deadlocks
> it was attempted to be mitigated by a refactor which created a new set of problems

(where the refactor they refer to is presumably #539).

QuintinWillison:
Fixes #651.

This is a relatively tentative fix in which I'm fairly confident due to the isolated and internal nature of the fix.

The fundamental issue was that we were not removing `ConnectionWaiter` instances as connection listeners, however I've also refactored the code a little for clarity.

There is probably plenty more I could have refactored but I wanted to keep the impact of this change minimal as we're aiming to get this out quickly as a bug fix.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -453,14 +453,18 @@ public class ConnectionManager implements ConnectListener {
         return currentState.queueEvents || currentState.sendEvents;
     }
 
-    /*************************************
-     * a class that listens for currentState change
-     * events for in-place authorization
-     *************************************/
-
+    /**
+     * Listens for connection state changes.
+     *
+     * The close() method must be called when the ConnectionWaiter is no longer needed.
+     */
     private class ConnectionWaiter implements ConnectionStateListener {
         private ConnectionStateChange change;
+        private boolean closed = false;
 
+        /**
+         * Create a ConnectionWaiter as a connection listener.
+         */
         private ConnectionWaiter() {
             connection.on(this);
         }
@@ -469,6 +473,10 @@ public class ConnectionManager implements ConnectListener {
          * Wait for a currentState change notification
          */
         private synchronized ErrorInfo waitForChange() {
+            if (closed) {
+                throw new IllegalStateException("Already closed.");
+            }
+
             Log.d(TAG, "ConnectionWaiter.waitFor()");
             if (change == null) {
                 try { wait(); } catch(InterruptedException e) {}
@@ -483,11 +491,22 @@ public class ConnectionManager implements ConnectListener {
          * ConnectionStateListener interface
          */
         @Override
-        public void onConnectionStateChanged(ConnectionStateChange state) {
-            synchronized(this) {
-                change = state;
-                notify();
+        public synchronized void onConnectionStateChanged(ConnectionStateChange state) {
+            change = state;
+            notify();
+        }
+
+        /**
+         * Remove this ConnectionWaiter as a connection listener.
+         */
+        private void close() {
+            // This method is explicitly not synchronized. There may be a case for this in the
+            // future, however its addition is designed to be lightweight with minimal impact.
+            if (closed) {
+                return;
             }
+            closed = true;
+            connection.off(this);
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -483,11 +491,22 @@ private synchronized ErrorInfo waitForChange() {
...
+        public synchronized void onConnectionStateChanged(ConnectionStateChange state) {
+            change = state;
+            notify();
+        }
+
+        /**
+         * Remove this ConnectionWaiter as a connection listener.
+         */
+        private void close() {
+            // This method is explicitly not synchronized. There may be a case for this in the
```
paddybyers, line 503: In this repo we've tended to use `/* .... */` comment style

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -483,11 +491,22 @@ private synchronized ErrorInfo waitForChange() {
...
+            // future, however its addition is designed to be lightweight with minimal impact.
+            if (closed) {
```
paddybyers, line 505: I think adding `synchronized` would be safe, but it this case it happens not to be necessary; if in the race the `.off()` is called twice, the second call would have no effect. Also, the calling code cannot end up calling `close()` more than once.

paddybyers (approved):
LGTM, thanks

vzhikserg (approved):

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -483,11 +491,22 @@ private synchronized ErrorInfo waitForChange() {
...
+                return;
             }
+            closed = true;
```
sacOO7, line 508: I feel `closed=true` can be set after `connection.off(this);`, that way it's logically closed after removing it from connection

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -881,64 +900,74 @@ public void run() {
...
+            while (waitingForConnected) {
+                final ErrorInfo reason = waiter.waitForChange();
+                final ConnectionState connectionState = currentState.state;
+                switch (connectionState) {
+                    case connected:
+                        Log.v(TAG, "onAuthUpdated: got connected");
+                        waitingForConnected = false;
+                        break;
+
+                    case connecting:
```
sacOO7, line 958: Not sure, but if it's `connecting`, it can transition to `connected` right ... that way it shouldn't break and just continue to check if the next state is connected

sacOO7: Maybe we can add explicit continue ... WDYT

sacOO7 (approved):

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
QuintinWillison: For interface commentary I would agree, however within a method implementation I tend to prefer `//` as it's more compact - typically for single line comments but also for multi-line. I should document this somewhere.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
QuintinWillison: Thanks for the suggestion, but I still prefer it this way around.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
QuintinWillison: I'm not quite sure what you mean, as this whole `switch` statement is validating a single state. So, `connected` is transitory and `connected` will get encountered on a subsequent iteration of the `while`.

QuintinWillison:
Thanks all for the prompt reviews. 😁

PR #653, opened 2021-03-02
Release/1.2.4

QuintinWillison:
I think it's worth me mentioning that I do still expect the assemble workflow to fail, as it is not yet implemented. However, my preference is to have it in place as a really obvious reminder that we need to work on that soon! 😁 

(i.e. I'm not blocking this release on that being fixed... as I have done previously I'll be building and releasing from my local workstation)


vzhikserg (approved):

owenpearson (approved):
LGTM

PR #655, opened 2021-03-02
Improve channel map operations in respect of thread-safety

### LINKED ISSUE #649: AblyBase.InternalChannels is not thread-safe
in the AblyBase class, the InternalChannels class holds a map of channels, which gets auto-populated on get.  however, the underlying data structure is a simple HashMap (which is not thread-safe).  when you do a get for a channel which does not currently exist in the map, a new one is created and then inserted into the underlying map.  this has three problems.  

1. it isn't atomic, i.e. two different threads making this call for a new channel the first time could end up with two different instances of Channel.  

2. the put() call on the underlying HashMap isn't thread safe, risking corrupting the underlying map.

3. modifying the options of the channel before returning it is not safe if the channel is being used by multiple threads concurrently.

note that 1,2 can be fairly simply and efficiently solved by using a ConcurrentHashMap and the computeIfAbsent method.

also note that AblyRealtime.InternalChannels has two of these problems as well (it does use a ConcurrentHashMap so 2 isn't an issue).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-650) by [Unito](https://www.unito.io/learn-more)

QuintinWillison:
Fixes #649.

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -130,10 +130,6 @@ public class AblyRealtime extends AblyRest implements AutoCloseable {
     }
 
     private class InternalChannels extends InternalMap<String, Channel> implements Channels, ConnectionManager.Channels {
-        private InternalChannels() {
-            super(new ConcurrentHashMap<String, Channel>());
-        }
-
         /**
          * Get the named channel; if it does not already exist,
          * create it with default options.
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -2,7 +2,6 @@ package io.ably.lib.realtime;
 
 import java.util.Iterator;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import io.ably.lib.rest.AblyRest;
 import io.ably.lib.transport.ConnectionManager;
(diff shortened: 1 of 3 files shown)


vzhikserg:
@QuintinWillison could you please remove [unused imports](https://github.com/ably/ably-java/pull/655/checks?check_run_id=2015348160)?

QuintinWillison:
Thanks, @vzhikserg ... that was rather sloppy of me! 🤦 
Fixing that now. Commit to follow.

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -148,21 +143,24 @@ public Channel get(String channelName) {
...
         @Override
-        public Channel get(String channelName, ChannelOptions channelOptions) throws AblyException {
-            Channel channel = map.get(channelName);
-            if (channel != null) {
+        public Channel get(final String channelName, final ChannelOptions channelOptions) throws AblyException {
+            // We're not using computeIfAbsent because that requires Java 1.8.
+            // Hence there's the slight inefficiency of creating newChannel when it may not be
+            // needed because there is an existingChannel.
+            final Channel newChannel = new Channel(AblyRealtime.this, channelName, channelOptions);
+            final Channel existingChannel = map.putIfAbsent(channelName, newChannel);
```
sacOO7, line 151: You can use `map.get(channelName) `for concurrentHashmap too right?
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html#get(java.lang.Object)

QuintinWillison: I don't understand your point. That is the method that this refactor is pretty much focussed on avoiding the use of. I've replaced two calls (one to `get`, followed by a subsequent conditional call to `put`, with a single call to atomic `putIfAbsent` as backed by the concurrent hash map).

Perhaps you need to re-read the issue to see what we're aiming to solve here. 🤔

paddybyers (approved):
LGTM, thanks

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
sacOO7: Actually, I was looking at the comment `slight inefficiency of creating newChannel`, if we use `map.get(channelName)`, we can avoid it. I feel it's a tradeoff, in most cases, the client will get an existing channel rather than creating a new one.

PR #657, opened 2021-03-02
Makes the Hosts class safe to be called from any thread

### LINKED ISSUE #650: Hosts class is not thread safe
The Hosts class is used by various threads when making api calls, however, the internals of this class are not handled in a thread safe way.  the various calls to manipulate the primary and preferred hosts and the expirations will not work correctly for more than one thread at a time.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-651) by [Unito](https://www.unito.io/learn-more)

QuintinWillison:
Fixes #650.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/Hosts.java
+++ b/lib/src/main/java/io/ably/lib/transport/Hosts.java
@@ -12,10 +12,10 @@ import java.util.Collections;
  * Object to encapsulate primary host name and shuffled fallback host names.
  */
 public class Hosts {
-    private String primaryHost;
+    private final String primaryHost;
     private String prefHost;
     private long prefHostExpiry;
-    boolean primaryHostIsDefault;
+    private final boolean primaryHostIsDefault;
     private final String defaultHost;
     private final String[] fallbackHosts;
     private final boolean fallbackHostsIsDefault;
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/transport/Hosts.java
```diff
@@ -10,18 +10,19 @@
...
-    boolean primaryHostIsDefault;
+    private final String primaryHost;
+    private final boolean primaryHostIsDefault;
     private final String defaultHost;
     private final String[] fallbackHosts;
     private final boolean fallbackHostsIsDefault;
     private final boolean fallbackHostsUseDefault;
     private final long fallbackRetryTimeout;
 
+    private final Preferred preferred = new Preferred();
```
paddybyers, line 25: I'm a bit surprised you didn't go for the option of `Preferred` being immutable, with `preferred` non-final.

paddybyers (approved):
LGTM, thanks

lib/src/main/java/io/ably/lib/transport/Hosts.java
QuintinWillison: Yeah, this was movement of state and the logic around mutating that state, as much for my own benefit to break the moving parts down a bit further... but, you're right, I could still have explored an immutable `Preferred`, as you say, with methods that return a new instance or the same instance as appropriate.

For speed of delivery right now I'm going to leave it as is but I appreciate the observation. 😁

PR #671, opened 2021-05-18
Add Ably-Agent header

KacperKluka:
Adds the `Ably-Agent` header together with name and version validation to both `ably-java` and `ably-android`. Regardless of the SDK variant (java or android) the header will always be `ably-java`. For Android the OS version number is also added as a `android/XX` header entry.


lib/src/main/java/io/ably/lib/rest/AblyBase.java
```diff
@@ -74,6 +71,11 @@ public AblyBase(ClientOptions options) throws AblyException {
             Log.e(getClass().getName(), msg);
             throw AblyException.fromErrorInfo(new ErrorInfo(msg, 400, 40000));
         }
+        if (options.agents != null && !AblyAgentValidator.areAllValid(options.agents)) {
+            String msg = "invalid agent provided";
+            Log.e(getClass().getName(), msg);
+            throw AblyException.fromErrorInfo(new ErrorInfo(msg, 400, 40000));
```
KacperKluka, line 77: I've based this on the error handling from above's `if(options == null)`, not sure if that's the right way to handle it :thinking:

lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -11,15 +11,15 @@
     /* versions */
     public static final float ABLY_VERSION_NUMBER   = 1.2f;
     public static final String ABLY_VERSION         = new DecimalFormat("0.0", new DecimalFormatSymbols(Locale.ENGLISH)).format(ABLY_VERSION_NUMBER);
-    public static final String ABLY_LIB_VERSION     = String.format("%s-%s", BuildConfig.LIBRARY_NAME, BuildConfig.VERSION);
+    public static final String ABLY_AGENT_VERSION   = String.format("%s/%s", "ably-java", BuildConfig.VERSION);
```
KacperKluka, line 14: Since the format agent name should always be `ably-java` I've decided to use a hardcoded string, is it okay or should I create some constant for that? :thinking:

lib/src/main/java/io/ably/lib/util/AblyAgentValidator.java
```diff
@@ -0,0 +1,45 @@
+package io.ably.lib.util;
+
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public class AblyAgentValidator {
+    /**
+     * Agent name validation regex.
+     * Allow only lowercase letters and '-'.
```
KacperKluka, line 9: I looked at some examples and the javascript PR and it looked like the names only have lowercase letters and dashes :wink: Not sure if that's a correct assumption though :thinking:

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
```diff
@@ -0,0 +1,61 @@
+package io.ably.lib.util;
+
+import android.os.Build;
```
KacperKluka, line 3: That worries me a bit because I'm not sure if that will cause troubles for the `ably-java` variant which doesn't have any android dependencies, right? :thinking:

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final String AGENT_DIVIDER = "/";
+    private static final String ANDROID_LIBRARY_NAME = "android";
+
+    public static String create(Map<String, String> additionalAgents) {
+        StringBuilder agentStringBuilder = new StringBuilder();
+        if (!additionalAgents.isEmpty()) {
+            agentStringBuilder.append(getAdditionalAgentEntries(additionalAgents));
+        }
+        agentStringBuilder.append(Defaults.ABLY_AGENT_VERSION);
+        if (BuildConfig.LIBRARY_NAME == ANDROID_LIBRARY_NAME) {
```
KacperKluka, line 30: It's a rather naive check but I didn't have any better idea :thinking:

lib/src/main/java/io/ably/lib/rest/AblyBase.java
QuintinWillison: I've asked around and haven't seen any suggestions for alternatives so this generic code suffice.

lib/src/main/java/io/ably/lib/transport/Defaults.java
QuintinWillison: Looks good to me.

lib/src/main/java/io/ably/lib/util/AblyAgentValidator.java
QuintinWillison: On this point I've just worked up a new docs PR to address this question, in response to https://github.com/ably/ably-cocoa/pull/1119 also. See https://github.com/ably/docs/pull/1096.

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
QuintinWillison: Yes, this is probably a problem. This project has a _unique_ structure, so we need to tread carefully. There is an Android specific folder so you may need to create an API to surface additional headers for whatever platform we're running on.

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
QuintinWillison: It would be better if you used `.equals()` rather than `==`.

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
```diff
@@ -0,0 +1,61 @@
...
+                additionalAgentsBuilder.append(AGENT_DIVIDER);
+                additionalAgentsBuilder.append(additionalAgentVersion);
+            }
+            additionalAgentsBuilder.append(AGENT_ENTRY_SEPARATOR);
+        }
+        return additionalAgentsBuilder.toString();
+    }
+
+
+    private static List<String> getSortedAgentNames(Map<String, String> agents) {
```
QuintinWillison, line 52: Nice, but there's no need to sort them.

If you're sorting them to make testing easier then I would handle that test-side.

lib/src/main/java/io/ably/lib/http/HttpCore.java
```diff
@@ -20,12 +20,8 @@
 import io.ably.lib.rest.Auth;
 import io.ably.lib.transport.Defaults;
 import io.ably.lib.transport.Hosts;
-import io.ably.lib.types.AblyException;
-import io.ably.lib.types.ClientOptions;
-import io.ably.lib.types.ErrorInfo;
-import io.ably.lib.types.ErrorResponse;
-import io.ably.lib.types.Param;
-import io.ably.lib.types.ProxyOptions;
+import io.ably.lib.types.*;
```
QuintinWillison, line 23: I would expect this to fail checkstyle as we've got the `AvoidStarImport` rule enabled. Our preference is to explicitly import required types. I know IDEs can have different opinions on this, but for us it's about code clarity and reducing namespace pollution.

QuintinWillison (changes requested):
Some change needed as a result of the docs PR I've opened today (see my reply to one of your initial comments). Also, generally, try to keep it as simple as possible - so things like sorting, given they're not required, just make it more difficult to review based on the specified requirements. Also, would be nice to see some links or references to the spec point(s) being implemented - comments are fine for this as we don't have any formal attributes in this codebase.

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
owenpearson: The precise origin of this comes from @paddybyers comment [here](https://github.com/ably/docs/pull/1034#discussion_r604345679), although it wasn't subsequently discussed in the thread. I do basically agree with it though; I think we should recommend an ordering just because it's nice to be able to say "this is precisely what the ably-agent header should look like" rather than having a bunch of equally correct possibilities. To me, dependency order makes the most sense semantically but I do also realise it might be ambiguous at times. I wouldn't personally be against changing the recommendation to something like alphabetical order just so that it's totally unambiguous for every implementation.

QuintinWillison: I think I'm ok with the idea of the library codebase getting pride of place (i.e. `ably-java`), but I'm not sure I see the value in any ordering beyond that... especially when it can be ambiguous as you say. Alphabetical, perhaps, but what value would that really add? There's got to be some concrete information we convey to the receiver (Ably's service) by doing this, as otherwise I don't really get the value in ordering at all.

lmars: There is an expectation that the Ably SDK version comes first in the list, and we're relying on that in the router to maintain the current metrics that power the [client library statistics dashboard](https://grafana.ably.io:8443/d/000000029/client-library-statistics?orgId=1) by emitting a metric for the first entry in the list, as it was considered unacceptable to break that dashboard.

There has been [discussion in Slack](https://ably-real-time.slack.com/archives/CURL4U2FP/p1621854952027600) about how we expose the rest of the entries in the list for later querying, since we aren't comfortable with just turning every entry into its own metric. The current plan is for the router to log the header which can then be queried in Athena, but there has been some debate on whether that is the correct approach.

lib/src/main/java/io/ably/lib/http/HttpCore.java
KacperKluka: Changed it back to how it was https://github.com/ably/ably-java/pull/671/commits/9b683325229128f70ab53335012647e206978262

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
KacperKluka: Too much Kotlin and I've forgotten how to write in Java :sweat_smile: Thanks for catching that :wink: https://github.com/ably/ably-java/pull/671/commits/ddd279cd03120efbf46b3f76f09e2fec7938bdcb

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
KacperKluka: I've removed the sorting https://github.com/ably/ably-java/pull/671/commits/522f29be7c671272604b4014d8c7d51002dc5e73 and placed the library agent as the first one in the agent list https://github.com/ably/ably-java/pull/671/commits/22258979d0770976c06c9e689d193214b817c78b

lib/src/main/java/io/ably/lib/util/AblyAgentValidator.java
KacperKluka: I've changed the regex https://github.com/ably/ably-java/pull/671/commits/3fb4f4978826d5db3dcc1e1176e5c5b63a582202 but still, I wasn't sure if uppercase letters are allowed so I've stuck to lowercase letters only, is that correct? :thinking:

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
KacperKluka: I've created a solution that works but I'm not very satisfied with how it looks like https://github.com/ably/ably-java/pull/671/commits/2561db3064b8e97b6a9454070038eee1bef6db3e but due to the project structure I haven't had any better idea so if you'll have any suggestions on how to improve it that would be awesome :smile:

QuintinWillison: Yeah, I agree. As discussed at more length with you via Slack DMs, the mutable static field isn't a great solution. We need to move this up to somewhere it can be governed by the compiler. Back to the drawing board. ✏️

QuintinWillison:
Due to the scale of changes that @KacperKluka is going to need to make to this codebase, it feels like we should fix #674 before continuing work on this PR. FYI, @paddybyers @jamienewcomb

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
QuintinWillison: Thanks. @owenpearson is going to lead in terms of bringing this conversation to a conclusive close.

lib/src/main/java/io/ably/lib/util/AgentHeaderCreator.java
KacperKluka: I've changed it so now each of the platforms (java, android) provide their own `PlatformAgentProvider` which returns the correct agent. This shouldn't affect the way of using the SDK and wee will avoid using that mutable static field :wink: https://github.com/ably/ably-java/pull/671/commits/08c9e6281a647e981516dd23ae75452cb23da29c
@paddybyers could you take a look at the above commit in terms of the SDKs architecture? Does it look alright to you?

KacperKluka:
For some reason, the integration test check is failing due to timeouts. Running the same test commands on my local machine works and all tests pass :thinking: Any ideas why this happens?

QuintinWillison (changes requested):
Thanks, @KacperKluka. Now that the dust has settled a bit in respect of our collective understanding of what we're doing here - both from the perspective of client libraries but also from the perspective of how it's handled at the backend - we need to make some changes to this work to fit:

- Remove code which isn't explicitly required - the main example being the agent validator
- Change the `ClientOptions.agents` API commentary to make it clear:
    - This is an internal interface to be used only by Ably-authored SDKs
    - That developers using this interface must add to `ably-common`'s [agents.json](https://github.com/ably/ably-common/blob/main/protocol/agents.json)

KacperKluka:
@QuintinWillison I've removed the agent validator https://github.com/ably/ably-java/pull/671/commits/111628b7196f0409254b951e38ffd0b22c4a17eb and updated javadocs https://github.com/ably/ably-java/pull/671/commits/343095ea30e0f21624a41b2d4615a19016e75ed5 :wink:

android/src/main/java/io/ably/lib/util/AndroidPlatformAgentProvider.java
```diff
@@ -0,0 +1,10 @@
+package io.ably.lib.util;
+
+import android.os.Build;
+
+public class AndroidPlatformAgentProvider implements PlatformAgentProvider {
+    @Override
+    public String createPlatformAgent() {
+        return "android" + AgentHeaderCreator.AGENT_DIVIDER + Build.VERSION.SDK_INT;
```
QuintinWillison, line 8: `android` added in https://github.com/ably/ably-common/pull/68

lib/src/main/java/io/ably/lib/util/PlatformAgentProvider.java
```diff
@@ -0,0 +1,10 @@
+package io.ably.lib.util;
+
+public interface PlatformAgentProvider {
+    /**
+     * Creates the platform agent for agent header {@link AgentHeaderCreator}.
+     *
+     * @return Platform agent string or null if not available.
```
QuintinWillison, line 7: I note that there's an empty implementation for Java.

Taking a quick look around, it seems to me that we must be able to query the Java runtime version. That would be very useful information to us. See [here](https://stackoverflow.com/a/2591122). It should be coded defensively, but could be helpful. I also wonder if it's also something we should query on Android too as it should still be a valid concept in that context also.

QuintinWillison (changes requested):
Also, I also wonder if there's any way that we can contain information about the SDK(s) against which the SDK was built. This may or may not be meaningful but questions we may want to answer, if possible, are what JDK was used and/or which Android SDK was used. Is that possible?

lib/src/main/java/io/ably/lib/util/PlatformAgentProvider.java
KacperKluka: I've added the JRE to the previously empty Java platform header :wink: https://github.com/ably/ably-java/pull/671/commits/2e870ac7a94225ada965fe5d6e962325d8b493f0

KacperKluka: And here's the agents file update PR https://github.com/ably/ably-common/pull/95

QuintinWillison:
As discussed [elsewhere (internal)](https://ably-real-time.slack.com/archives/C019LDWM9UH/p1626690797006300), I think we should land https://github.com/ably/ably-java/pull/684 to `main` before considering this pull request ready to land (i.e. I would like to see Android emulation workflow running in CI against the code changes in this pull request).

lib/src/main/java/io/ably/lib/util/PlatformAgentProvider.java
QuintinWillison: 🚀

QuintinWillison (changes requested):
Tests need updating, please. [e.g.](https://github.com/ably/ably-java/pull/671/checks?check_run_id=3101347300#step:3:452):

```
org.junit.ComparisonFailure: expected:<ably-java/1.2.6[ jre/11.0.11]> but was:<ably-java/1.2.6[]>
```

java/src/main/java/io/ably/lib/util/JavaPlatformAgentProvider.java
```diff
@@ -0,0 +1,8 @@
+package io.ably.lib.util;
+
+public class JavaPlatformAgentProvider implements PlatformAgentProvider {
+    @Override
+    public String createPlatformAgent() {
+        return "jre" + AgentHeaderCreator.AGENT_DIVIDER + System.getProperty("java.version");
```
QuintinWillison, line 6: [getProperty](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getProperty-java.lang.String-) can return `null`. We should defend against that possibility. Perhaps also against the possibility of an empty string coming back. Perhaps even going as far as to trim whitespace from either end of the returned string before checking if it's empty. If empty or `null` then perhaps use a placeholder of `"unknown"`.

QuintinWillison (changes requested):
Sorry, just a little bit more refinement needed I think...

java/src/main/java/io/ably/lib/util/JavaPlatformAgentProvider.java
KacperKluka: https://github.com/ably/ably-java/pull/671/commits/cfe904d1c816d63e3c0ff0f3f9d645a9b87c5944

QuintinWillison: Perfect, thanks.

And no idea what I was going on about with `"unknown"` - thanks for ignoring that! 😆 
(what you've done is better and cleaner)

QuintinWillison (approved):

martin-morek (approved):

AndyNicks (approved):

QuintinWillison:
I decided to land this PR, despite my [previous assertion](https://github.com/ably/ably-java/pull/671#issuecomment-882440329). When we do get Android emulation workflow running, and if that reveals any bugs, then we'll fix them then.

Issue #620, opened 2020-10-12
Fail connection immediately if authorize() called and 403 returned


QuintinWillison:
The first task here is to validate whether this client library behaves per `RSA4d` and `RSA4d1`, as added in https://github.com/ably/docs/pull/961 with reference implementation in https://github.com/ably/ably-js/pull/680.

If not, then implement correct behaviour with appropriate testing.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-449) by [Unito](https://www.unito.io)

PR #672, opened 2021-05-21
Checking if error code is 403 and failing connection

martin-morek:
- If server replies with 403 status code on authorisation attempt, connection state turns to Failed
- Fixes [#620](https://github.com/ably/ably-java/issues/620)

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -978,6 +978,12 @@ public class ConnectionManager implements ConnectListener {
      */
     public void onAuthError(ErrorInfo errorInfo) {
         Log.i(TAG, String.format("onAuthError: (%d) %s", errorInfo.code, errorInfo.message));
+
+        if(errorInfo.code == 403) {
+            this.connection.state = ConnectionState.failed;
+            return;
+        }
+
         switch (currentState.state) {
             case connecting:
                 ITransport transport = this.transport;
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -980,7 +980,14 @@ public class ConnectionManager implements ConnectListener {
         Log.i(TAG, String.format("onAuthError: (%d) %s", errorInfo.code, errorInfo.message));
 
         if(errorInfo.code == 403) {
-            this.connection.state = ConnectionState.failed;
+            ConnectionStateChange failedStateChange =
+                new ConnectionStateChange(
+                    connection.state,
+                    ConnectionState.failed,
+                    0,
+                    errorInfo);
+
+            this.connection.onConnectionStateChange(failedStateChange);
             return;
         }
 
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -979,7 +979,7 @@ public class ConnectionManager implements ConnectListener {
     public void onAuthError(ErrorInfo errorInfo) {
         Log.i(TAG, String.format("onAuthError: (%d) %s", errorInfo.code, errorInfo.message));
 
-        if(errorInfo.code == 403) {
+        if(errorInfo.statusCode == 403) {
             ConnectionStateChange failedStateChange =
                 new ConnectionStateChange(
                     connection.state,
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
@@ -2,6 +2,7 @@ package io.ably.lib.test.realtime;
 
 import io.ably.lib.realtime.*;
 import io.ably.lib.test.common.Setup;
+import io.ably.lib.types.*;
 import io.ably.lib.util.Log;
 
 import io.ably.lib.debug.DebugOptions;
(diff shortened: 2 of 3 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -978,6 +978,12 @@ public void onAuthUpdated(final String token, final boolean waitForResponse) thr
      */
     public void onAuthError(ErrorInfo errorInfo) {
         Log.i(TAG, String.format("onAuthError: (%d) %s", errorInfo.code, errorInfo.message));
+
+        if(errorInfo.code == 403) {
+            this.connection.state = ConnectionState.failed;
```
sacOO7, line 983: I feel you should check if there is a better way to set the state. Generally we don't directly set the state. You can check for method `setState` on connection. 
- SetState is responsible for broadcasting state change to components that need to act upon state change

paddybyers: Correct, just setting the state here will break stuff

sacOO7:
Also, I strongly feel, we need to add couple of tests for the change made

martin-morek:
> Also, I strongly feel, we need to add couple of tests for the change made @sacOO7

I definitely agree on this, however could you please point me what's the correct place to put them? I've seen there are already test classes but I cannot figurate out how to fail auth to get the correct error code.

KacperKluka (approved):
LGTM, given that I have little knowledge of the codebase and specs :wink:

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -74,6 +75,44 @@ public void auth_client_match_tokendetails_null_clientId() {
         }
     }
 
+    /**
+     * RSA4d:  If a request by a realtime client to an authUrl results in an HTTP 403 response,
+     * or any of an authUrl request, an authCallback, or a request to Ably to exchange
+     * a TokenRequest for a TokenDetails result in an ErrorInfo with statusCode 403,
+     * as part of an attempt by the realtime client to authenticate, then the client library
```
paddybyers, line 82: Shouldn't this be testing that the 80019 code is also returned?

paddybyers (changes requested):
It looks like the requirement isn't implemented - it states that the returned error needs to be wrapped in a new error whose `statusCode` is 403 and `code` is 80019.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -74,6 +75,44 @@ public void auth_client_match_tokendetails_null_clientId() {
...
+     * should transition to the FAILED state, with an ErrorInfo (with code 80019, statusCode 403,
+     * and cause set to the underlying cause) emitted with the state change and set as the connection
+     * errorReason
+     *
+     * Verify end connection state is failed
+     */
+    @Test
+    public void auth_client_fails_authorize_server_forbidden() {
```
sacOO7, line 90: You should mention spec that test belongs too, like current one to `RSA4d` and `RSA4d1`, for right naming conventions, check for other tests

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -74,6 +75,44 @@ public void auth_client_match_tokendetails_null_clientId() {
...
+            try {
+                ablyRealtime.auth.authorize(null, null);
+            } catch (AblyException e) {
+                assertEquals(403, e.errorInfo.statusCode);
+            }
+
+            /* wait for failed state */
+            Helpers.ConnectionWaiter connectionWaiter = new Helpers.ConnectionWaiter(ablyRealtime.connection);
+            connectionWaiter.waitFor(ConnectionState.failed);
+            assertEquals("Verify connected state has failed", ConnectionState.failed, ablyRealtime.connection.state);
```
sacOO7, line 109: Refer to this [PR implementation](https://github.com/ably/ably-dotnet/pull/459/files), for doing right assertions

sacOO7: Just a psuedocode for assertions 
```
change.Previous.Should().Be(ConnectionState.Connected);
change.Reason.Code.Should().Be(expectedCode);           
realtimeClient.Connection.ErrorReason.Code.Should().Be(expectedCode);             
realtimeClient.Connection.ErrorReason.StatusCode.Should().Be(HttpStatusCode.Forbidden); // 403

```

sacOO7:
Also, add some details to the active PR, as in which issues it's fixing. 
Adding one more reference [implementation in JS](https://github.com/ably/ably-js/pull/680)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -74,6 +75,66 @@ public void auth_client_match_tokendetails_null_clientId() {
...
+            ablyRealtime.connection.connect();
+
+            Helpers.ConnectionWaiter connectionWaiter = new Helpers.ConnectionWaiter(ablyRealtime.connection);
+            connectionWaiter.waitFor(ConnectionState.connected);
+
+            ablyRealtime.connection.once(ConnectionEvent.failed, new ConnectionStateListener() {
+                @Override
+                public void onConnectionStateChanged(ConnectionStateChange stateChange) {
+                    assertEquals(stateChange.previous, ConnectionState.connected);
+                    assertEquals(stateChange.reason.code, 80019);
```
sacOO7, line 114: This should follow the right format, `expected` and `actual` values should be at the right places

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
sacOO7: You should mention it as a part of test

martin-morek: Sorry but I don't follow, what would you expect to be a test name? Or what is missing?

sacOO7: Check this test, and commented documentation for the test 
https://github.com/ably/ably-java/blob/73d211c1f7da1abd3e583dedbf0431625491568f/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java#L1521-L1540 
It includes spec ID's as to where spec belongs to

martin-morek: I've extended test comments and tried to kept consistent with the rest of the test class

sacOO7 (approved):

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
martin-morek: I believe it does now. On the line 127,128 I'm checking error status code and error code as well. Also in `ConnectionStateListener` where I awaiting `ConnectionEven.failed`. Or do you have something else in mind?

paddybyers: no, lgtm, thanks

PR #675, opened 2021-05-28
Using a clientId should no longer be forcing token auth

martin-morek:
- Fixes [#345](https://github.com/ably/docs/pull/345)
- Basic authentication can be used with specified `clientId`

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -868,8 +868,7 @@ public class Auth {
 
         /* decide default auth method (spec: RSA4) */
         if(authOptions.key != null) {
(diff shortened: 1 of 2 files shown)


QuintinWillison (approved):

QuintinWillison:
@martin-morek, it sounds like it's worth double checking. Thanks for your input, @sacOO7.

martin-morek:
@sacOO7 @QuintinWillison  Sorry, but I'm missing out the context a bit. About what spec changes are we talking about? Could you please point me where can I find them to verify?

sacOO7:
As per original issues, you can see spec changes that happened as a part of PR to the DOC

martin-morek:
I've tried to verify it, `RSA4` spec should be okey, there are tests for each scenario.

Issue #574, opened 2020-05-04
RSC7c TI1 addRequestIds on ClientOptions and requestId on ErrorInfo


QuintinWillison:
Missing for this client library, though implemented [for 1.2](https://github.com/ably/ably-cocoa/pull/982) in ably-cocoa.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-585) by [Unito](https://www.unito.io)

PR #678, opened 2021-06-16
Add request_id query param if addRequestIds is enabled

martin-morek:
- Fixes [574](https://github.com/ably/ably-java/issues/574)
- Implementation of `RSC7c`. Adds `addRequestIds` parameter to client options. If this option is enabled every REST request to Ably should include in a request_id query string parameter a random string obtained by url-safe base64-encoding a sequence of at least 9 bytes obtained from a source of randomness. This request ID must remain the same if a request is retried to a fallback host per RSC15. Any log messages associated with the request should include the request ID. If the request fails, the request ID must be included in the ErrorInfo returned to the user.

### DIFF
--- a/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
+++ b/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
@@ -7,9 +7,6 @@ import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.support.v4.content.LocalBroadcastManager;
 
-import com.google.android.gms.tasks.OnCompleteListener;
-import com.google.android.gms.tasks.Task;
-import com.google.firebase.iid.InstanceIdResult;
 import com.google.gson.JsonObject;
 
 import java.lang.reflect.Constructor;
@@ -21,6 +18,7 @@ import io.ably.lib.http.*;
 import io.ably.lib.rest.AblyRest;
 import io.ably.lib.rest.DeviceDetails;
 import io.ably.lib.types.*;
+import io.ably.lib.util.Crypto;
 import io.ably.lib.util.IntentUtils;
 import io.ably.lib.util.Log;
 import io.ably.lib.util.Serialisation;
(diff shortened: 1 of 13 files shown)


android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
```diff
@@ -142,6 +140,9 @@ public void execute(HttpScheduler http, Callback<JsonObject> callback) throws Ab
                             if(ably.options.pushFullWait) {
                                 params = Param.push(null, "fullWait", "true");
                             }
+                            if(ably.options.addRequestIds) { // RSC7c
```
QuintinWillison, line 143: While I recognise that this repetition was already there (re `pushFullWait` > `fullWait` lines above), could we please DRY this up by perhaps taking this params building code and moving it somewhere common. I think it would be appropriate to do that in the scope of this pull request as the size of this common code now doubles.

android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
```diff
@@ -7,9 +7,6 @@
 import android.content.SharedPreferences;
 import android.support.v4.content.LocalBroadcastManager;
 
-import com.google.android.gms.tasks.OnCompleteListener;
```
QuintinWillison, line 10: Presumably this change does not relate directly to the scope of work being undertaken under this PR. No need to change now but in future please try to make changes like this under a separate PR to make it easier for reviewers to focus on the changes required for this piece of work.

QuintinWillison (changes requested):
Looks good, thanks. Just a bit of refactoring please.

android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
martin-morek: I've just noticed that's unused so I removed it but I agree with you and will do it in new PR next time.

android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
martin-morek: I've pulled common logic out, but not sure if to the best place. Please let me know if you have a better idea.

QuintinWillison: I see you (https://github.com/ably/ably-java/pull/678/commits/89ecf93d93076f20c54eef931f324b20bd445068).

I think the location of this new code is entirely appropriate. What I would like to see now, however, for this new method is:

1. API commentary
2. Unit test(s)

QuintinWillison (approved):

Issue #593, opened 2020-07-22
Support for encryption of shared preferences


QuintinWillison:
We persist [secrets](https://github.com/ably/ably-java/blob/1399a98d6d0e7c5b75218e945c284708940ec1f0/android/src/main/java/io/ably/lib/push/LocalDevice.java#L123) (e.g. `ABLY_DEVICE_SECRET`) and [tokens](https://github.com/ably/ably-java/blob/1399a98d6d0e7c5b75218e945c284708940ec1f0/android/src/main/java/io/ably/lib/push/LocalDevice.java#L113) (e.g. `ABLY_DEVICE_IDENTITY_TOKEN`) to `SharedPreferences` without encryption or any other form of obfuscation, in terms of both values and the keys they are stored under.

This has been raised by a customer who discovered this when doing a security test after they "rooted" their app.

When Ably first implemented `LocalDevice` storage, there wasn't anything we deemed suitable that would give any material increase in security, hence the implementation was kept simple.

Perhaps the best solution going forward would be to allow app developers to override the storage I/O in order to provide their own encryption and/or platform service designed for secrets. Such a provider might be as simple as:

```java
interface Storage {
    void write(String key, String value);
    String read(String key, String defaultValue);
}
```

We would provide a default implementation using shared preferences, as we do currently.

I would also suggest we look at fixing #562 when we work on this as it's in the same area of the codebase.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-589) by [Unito](https://www.unito.io)

ben-xD:
It wasn't very clear to me what this issue or related PR was doing. After reading the diff of the PR, I noticed in `ClientOptions`: 
https://github.com/ably/ably-java/blob/d9ad0e1923fbfa5ad85eecf07be52fcc5b470b09/lib/src/main/java/io/ably/lib/types/ClientOptions.java#L200-L204

So this issue only allows users to specify a custom implementation of `Storage`, and doesn't actually do anything with encryption. Users would have to add logic to encrypt the data themselves. This is probably because the clients shouldn't really encrypt the data using a symmetric key which is available locally, since this can be found and used by a threat actor.

sacOO7:
Actually, there is a official way to encrypt shared pref. With a key, that's not exposed to a user.
https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences
[...]

PR #681, opened 2021-07-02
Support for encryption of shared preferences

martin-morek:
- Fixes [593](https://github.com/ably/ably-java/issues/593)
- Default SharedPreferences could be replaced with custom Storage implementation, ClientOptions allows to inject custom object. If nothing is injected current behaviour using SharedPreferences is used so no need to do any changes to keep current functionality.

### DIFF
--- a/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
+++ b/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
@@ -1,0 +1,52 @@
+package io.ably.lib.push;
+
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+
+import java.lang.reflect.Field;
+
+public class SharedPreferenceStorage implements Storage{
+
+    private final ActivationContext activationContext;
+
+    public SharedPreferenceStorage(ActivationContext activationContext) {
+        this.activationContext = activationContext;
+    }
+
+    @Override
+    public void putString(String key, String value) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
+        prefs.edit().putString(key, value).apply();
+    }
+
+    @Override
+    public void putInt(String key, int value) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
+        prefs.edit().putInt(key, value).apply();
+    }
+
+    @Override
+    public String getString(String key, String defaultValue) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
+        return prefs.getString(key, defaultValue);
+    }
+
+    @Override
+    public int getInt(String key, int defValue) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
+        return prefs.getInt(key, defValue);
+    }
+
+    @Override
+    public void reset(Field[] fields) {
+        SharedPreferences.Editor editor = activationContext.getPreferences().edit();
+        for (Field f : fields) {
+            try {
+                editor.remove((String) f.get(null));
+            } catch (IllegalAccessException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        editor.commit();
+    }
+}
--- a/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
+++ b/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
@@ -13,28 +13,28 @@ public class SharedPreferenceStorage implements Storage{
         this.activationContext = activationContext;
     }
 
+    private SharedPreferences sharedPreferences() {
+        return PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
+    }
+
     @Override
     public void putString(String key, String value) {
-        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
-        prefs.edit().putString(key, value).apply();
+        sharedPreferences().edit().putString(key, value).apply();
     }
 
     @Override
     public void putInt(String key, int value) {
-        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
-        prefs.edit().putInt(key, value).apply();
+        sharedPreferences().edit().putInt(key, value).apply();
     }
 
     @Override
     public String getString(String key, String defaultValue) {
-        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
-        return prefs.getString(key, defaultValue);
+        return sharedPreferences().getString(key, defaultValue);
     }
 
     @Override
     public int getInt(String key, int defaultValue) {
-        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
-        return prefs.getInt(key, defaultValue);
+        return sharedPreferences().getInt(key, defaultValue);
     }
 
     @Override
--- a/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
+++ b/android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
@@ -18,22 +18,22 @@ public class SharedPreferenceStorage implements Storage{
     }
 
     @Override
-    public void putString(String key, String value) {
+    public void put(String key, String value) {
         sharedPreferences().edit().putString(key, value).apply();
     }
 
     @Override
-    public void putInt(String key, int value) {
+    public void put(String key, int value) {
         sharedPreferences().edit().putInt(key, value).apply();
     }
 
     @Override
-    public String getString(String key, String defaultValue) {
+    public String get(String key, String defaultValue) {
         return sharedPreferences().getString(key, defaultValue);
     }
 
     @Override
-    public int getInt(String key, int defaultValue) {
+    public int get(String key, int defaultValue) {
         return sharedPreferences().getInt(key, defaultValue);
     }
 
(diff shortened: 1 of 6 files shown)


lib/src/main/java/io/ably/lib/push/Storage.java
```diff
@@ -0,0 +1,16 @@
...
+
+    void putString(String key, String value);
+
+    void putInt(String key, int value);
+
+    String getString(String key, String defaultValue);
+
+    int getInt(String key, int defaultValue);
+
+    void reset(Field[] fields);
```
sacOO7, line 15: Can be renamed as `clear` , `flush`or `delete`

sacOO7: Reset feels like resetting to some default value

android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private final ActivationContext activationContext;
+
+    public SharedPreferenceStorage(ActivationContext activationContext) {
+        this.activationContext = activationContext;
+    }
+
+    @Override
+    public void putString(String key, String value) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(activationContext.getContext());
```
sacOO7, line 18: This code is being repeated in every interface method, can we take out somewhere?

lib/src/main/java/io/ably/lib/push/Storage.java
```diff
@@ -0,0 +1,16 @@
+package io.ably.lib.push;
+
+import java.lang.reflect.Field;
+
+public interface Storage {
+
+    void putString(String key, String value);
```
sacOO7, line 7: Also this is out of the context, but can we add small documentation at the top of the interface, as to what it does, so
developers using this interface will know what to do with it, like in case of `getString` it's mandatory to return `defaultValue` if value we are looking for a particular `key` is not found

lib/src/main/java/io/ably/lib/push/Storage.java
```diff
@@ -0,0 +1,16 @@
+package io.ably.lib.push;
+
+import java.lang.reflect.Field;
+
+public interface Storage {
```
sacOO7, line 5: Also, I was curious, if current `Storage` interface class is meant for both `ephermal` or `permanant` storage?

android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
martin-morek: Maybe it can be initialised just once, in constructor of `SharedPreferenceStorage`, however I'm not sure since it was also recreated every time when it was called in `LocalDevice` (original implementation). Or did you just mean to create method for it?

sacOO7: I am coming from `DRY` principle, we should take out repetition somewhere else.

sacOO7:
I am sorry, but I can see some of review comments being closed without reply comments? Do we have any reasons as to why there are being closed

lib/src/main/java/io/ably/lib/push/Storage.java
sacOO7: Not sure why this one is closed

sacOO7: Can we quickly add a small doc above each interface method, one liner should also be enough, same goes for interface name. When Javadoc API documentation is generated, it will help there too : )

lib/src/main/java/io/ably/lib/push/Storage.java
sacOO7: If it also works for epthermal storage, you can close this one too : ), I will approve right away

lib/src/main/java/io/ably/lib/push/Storage.java
martin-morek: Done

lib/src/main/java/io/ably/lib/push/Storage.java
martin-morek: Not sure if I 100% understand what do you mean.  I've tested it in this PR where I used hash map as a storage, which is ephemeral storage.

jamienewcomb:
@sacOO7 are you happy with the changes?

sacOO7 (approved):

android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
```diff
@@ -0,0 +1,52 @@
...
+        return sharedPreferences().getInt(key, defaultValue);
+    }
+
+    @Override
+    public void clear(Field[] fields) {
+        SharedPreferences.Editor editor = activationContext.getPreferences().edit();
+        for (Field f : fields) {
+            try {
+                editor.remove((String) f.get(null));
+            } catch (IllegalAccessException e) {
```
QuintinWillison, line 46: I was initially confused as to why it is necessary to catch this exception from the [remove](https://developer.android.com/reference/android/content/SharedPreferences.Editor#remove(java.lang.String)) method (it's unchecked and not documented as far as I can see).

But then I looked at the `SharedPreferences.Editor` API and spotted the [clear](https://developer.android.com/reference/android/content/SharedPreferences.Editor#clear()) method. Is there a reason you can't just use that rather than manually iterate fields?

lib/src/main/java/io/ably/lib/push/Storage.java
```diff
@@ -0,0 +1,45 @@
+package io.ably.lib.push;
+
+import java.lang.reflect.Field;
+
+/**
+ * Interface for an entity that supplies key value store
+ */
+public interface Storage {
```
QuintinWillison, line 8: I recognise that you're reflecting the Android APIs in your method names here, and in terms of using `get` and `put` as the verbs (instead of [my original](https://github.com/ably/ably-java/issues/593#issue-663543449) `read` and `write` suggestions) then I am happy with that.

What I don't think is needed, however, are the type suffixes as Java supports method overloads in this scenario and that leads to a cleaner API. i.e.:

- `putString` and `putInt` both rename to just `put` (or, maybe `putValue`)
- `getString` and `getInt` both rename to just `get` (or, maybe `getValue`)

lib/src/main/java/io/ably/lib/push/Storage.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Insert string value in to storage
```
QuintinWillison, line 11: `Insert` is the wrong verb to use in this context, as if there's an existing value then it'll be replaced.

lib/src/main/java/io/ably/lib/types/ClientOptions.java
```diff
@@ -195,4 +196,10 @@ public ClientOptions(String key) throws AblyException {
      * before responding.
      */
     public boolean pushFullWait = false;
+
+    /**
+     * Allows provide custom Local Device storage. In a case nothing is provided default implementation
```
QuintinWillison, line 201:
```suggestion
     * Custom Local Device storage. In the case that nothing is provided then a default implementation
```

lib/src/main/java/io/ably/lib/types/ClientOptions.java
```diff
@@ -195,4 +196,10 @@ public ClientOptions(String key) throws AblyException {
...
+     * using SharedPreferences is used.
+     */
+    public Storage storage = null;
```
QuintinWillison, line 204: I think we could perhaps rename this field to `localStorage` (the interface name can remain as `Storage`).

QuintinWillison (changes requested):
Some tweaks, please.

lib/src/main/java/io/ably/lib/push/Storage.java
martin-morek: of course, great catch!

android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
martin-morek: We surely can use `clear` method. The reason why I did it this way is because also origin implementation specified which properties should be removed (wasn't removing all of them automatically). I'm not sure if something else is stored there what we don't want to be removed.

lib/src/main/java/io/ably/lib/push/Storage.java
QuintinWillison: https://github.com/ably/ably-java/pull/681/commits/16bf5233f352b5817f5afae1af1e95ab7e283c7f

lib/src/main/java/io/ably/lib/push/Storage.java
QuintinWillison: https://github.com/ably/ably-java/pull/681/commits/16bf5233f352b5817f5afae1af1e95ab7e283c7f

lib/src/main/java/io/ably/lib/types/ClientOptions.java
QuintinWillison: https://github.com/ably/ably-java/pull/681/commits/16bf5233f352b5817f5afae1af1e95ab7e283c7f

lib/src/main/java/io/ably/lib/types/ClientOptions.java
QuintinWillison: https://github.com/ably/ably-java/pull/681/commits/16bf5233f352b5817f5afae1af1e95ab7e283c7f

android/src/main/java/io/ably/lib/push/SharedPreferenceStorage.java
QuintinWillison: Fair enough and thanks for explaining. I understand this much better now.

QuintinWillison (approved):

Issue #686, opened 2021-07-15
Remove use of forClass method in push activation state machine implementation


QuintinWillison:
See [here](https://github.com/ably/ably-java/blob/31d8ae442e36610158565a86069bff7e7b6782eb/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java#L681).

Why do we need to do this?

We have a customer who is experiencing a crash, which I am pretty sure is coming from an exception thrown here.

See [this internal thread](https://ably-real-time.slack.com/archives/CKAPDADC6/p1626334221082100).

We should probably work on https://github.com/ably/ably-java/issues/685 before trying to investigate this issue, as that code encapsulates this.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1041) by [Unito](https://www.unito.io)

QuintinWillison:
The `forClass` and `newInstance` code came from the fix applied for https://github.com/ably/ably-java/issues/598 back in September 2020.

It violates every protection offered to us by the compiler. This must be improved. We should not be relying on such fundamental runtime exceptions to fail such critical code paths.

`ActivationStateMachine.Event` is abstract, yet defines no methods. Perhaps a solution would be to delegate responsibility for persistence of events to this class, in such a way that the compiler can police it before it ever gets to runtime.

Clearly also related: https://github.com/ably/ably-java/issues/546

Issue #685, opened 2021-07-15
Push Activation State Machine exception handling needs improvement


QuintinWillison:
There are places where we catch `Exception` and then re-throw as uncaught `RuntimeException` (e.g. [here](https://github.com/ably/ably-java/blob/31d8ae442e36610158565a86069bff7e7b6782eb/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java#L695)), which would almost always be inappropriate and heavy handed.

Answers to [this question](https://stackoverflow.com/questions/2416316/why-is-the-catchexception-almost-always-a-bad-idea) perhaps illustrate the point better than I could describe myself, in respect as to why `catch (Exception)` is almost always a bad idea.

Plus, throwing a `RuntimeException` will cause the application process to crash. Thus, we effectively have engineered in a critical failure state when I suspect something lighter weight could have been implemented.

A customer has been seeing crashes on Android devices at runtime which pass through this exception handling path. Our manner of handling these exceptions is making it more difficult to debug that issue ([internal thread](https://ably-real-time.slack.com/archives/CKAPDADC6/p1626334221082100)).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1040) by [Unito](https://www.unito.io)

PR #689, opened 2021-07-23
Removed forName method 

martin-morek:
- fixes [686](https://github.com/ably/ably-java/issues/686) and [685](https://github.com/ably/ably-java/issues/685)
- Removed use of `forName` method in push activation state machine implementation and replaced by switch constructing Event object based on name

### DIFF
--- a/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
+++ b/android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
@@ -32,7 +32,7 @@ import java.util.ArrayDeque;
 
 public class ActivationStateMachine {
     public static class CalledActivate extends ActivationStateMachine.Event {
-        private static final String NAME = "CalledActivate";
+        public static final String NAME = "CalledActivate";
 
         public static ActivationStateMachine.CalledActivate useCustomRegistrar(boolean useCustomRegistrar, SharedPreferences prefs) {
             prefs.edit().putBoolean(ActivationStateMachine.PersistKeys.PUSH_CUSTOM_REGISTRAR, useCustomRegistrar).apply();
@@ -46,7 +46,7 @@ public class ActivationStateMachine {
     }
 
     public static class CalledDeactivate extends ActivationStateMachine.Event {
-        private static final String NAME = "CalledDeactivate";
+        public static final String NAME = "CalledDeactivate";
 
         static ActivationStateMachine.CalledDeactivate useCustomRegistrar(boolean useCustomRegistrar, SharedPreferences prefs) {
             prefs.edit().putBoolean(ActivationStateMachine.PersistKeys.PUSH_CUSTOM_REGISTRAR, useCustomRegistrar).apply();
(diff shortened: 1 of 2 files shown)


android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
```diff
@@ -678,21 +717,12 @@ private boolean persist() {
...
-                    continue;
-                }
+                ActivationStateMachine.Event event = Event.constructEventByName(className);
                 deque.add(event);
-            } catch(Exception e) {
-                throw new RuntimeException(e);
+            } catch (ClassNotFoundException e) {
+                Log.e(TAG, e.getLocalizedMessage());
+            } catch (InstantiationException e) {
+                continue;
```
QuintinWillison, line 725: If this needs to be checked then is there any reason we're not logging it out also?

QuintinWillison (changes requested):
Thanks, this is a great start. However, as discussed in our call earlier today, this only presents one side of the story... I think we need to see:

- Addition of functionality that does directly the opposite of your `constructEventByName` static method - probably an instance method on `Event`, perhaps simply something like `getEventName()`
- Some unit tests for this class hierarchy proving:
    - that `Event` instances can be represented as string and then rehydrated from the same
    - that attempts to rehydrate unsupported event types (those with non-nullary constructor) throw the right exception
    - that attempts to rehydrate unknown event types throw the right exception

We need to get rid of any code that uses runtime reflection. This has to be checked by the compiler.

martin-morek:
@QuintinWillison could you please elaborate on `getEventName()` method? Right now for getting event name we are using `getClass().getName()`, I'm not sure if this can be simplified. What we can do is define property called `name` for an event object and manually fill it with a value. That way when we change class name code stays functional but real name would differ.

QuintinWillison:
Hi @martin-morek - so, the use of `getClass().getName()` is the crux of the issue! 😄 

That method is dynamic at runtime. If the class name is different, because of obfuscation or something else outside of control as SDK developers, then that can mess things up - in particular from build to build of a application.

Each class mast vend its own serialisation name from a static string. These cannot be dynamically sourced like this.

android/src/main/java/io/ably/lib/push/ActivationStateMachine.java
```diff
@@ -26,169 +26,200 @@
...
-        public abstract String getName();
-
-        public static Event constructEventByName(String className) throws ClassNotFoundException, InstantiationException {
-            ActivationStateMachine.Event event;
+        /**
+         * The name to be used when persisting this class, or null if this class should not be persisted.
+         */
+        public String getPersistedName() {
+            return null;
+        }
```
martin-morek, line 167: My idea here was to force every event class to provide its own implementation of `getPersistedName()`, that why it was abstract. Main reason for that was avoid bug in the future if we add new type of event and forget to override name.

QuintinWillison: I understand that and I recognise your point. I also appreciate very much that you are thinking in terms of future maintainability. 😍 

In this particular case, however, I believe that the balance of reduced noise in subclass implementations for what is a well established area of the codebase tips the balance in favour of providing this default implementation.

QuintinWillison (approved):

Issue #382, opened 2018-04-06
Ensure correct feedback to developer when malformed key is supplied


funkyboy:
Now it's possible to write 

`ably = new AblyRealtime("");` 

The app starts and tries to connect. Then a timeout happens, because the key doesn't exists. ~~We should check if the key passed as a parameter is not empty and in the correct format.~~ This doesn't help user to solve the problem and doesn't show what the problem is.

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-418) by [Unito](https://www.unito.io)

SimonWoolf:
Note that per https://docs.ably.io/client-lib-development-guide/features/#RSC1a , since `""` doesn't contain a colon, it should be interpreted as a token, which means there's no clientside format validation.

(I mean sure we could validate that the string isn't empty if we want, a valid token's never going to be the empty string. But shrug, not a big deal if we don't, and there certainly isn't any token validation we can/should do clientside beyond that)

paddybyers:
It is interpreted as a token. I don't see what the error is.

What's the timeout?

kavalerov:
Based on the conversation we had this morning, the real problem here is that timeout is not really a right behaviour in this case. We need to: 
1) check what is the behaviour currently
2) If behaviour is same as described in this ticket, we need to make sure that developer gets correct feedback (error message?) if malformed key is provided.

PR #690, opened 2021-07-26
Throws exception when AuthOptions are initialized with an empty string

martin-morek:
- fixes [382](https://github.com/ably/ably-java/issues/382)
- AuthOptions cannot be initialised with an empty string. When empty string is provided `IllegalArgumentException` is thrown.

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
@@ -25,6 +25,16 @@ public class RealtimeAuthTest extends ParameterizedTest {
     public Timeout testTimeout = Timeout.seconds(30);
 
     /**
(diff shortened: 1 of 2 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -24,6 +24,16 @@
     @Rule
     public Timeout testTimeout = Timeout.seconds(30);
 
+    /**
+     * Verifies an Exception is thrown, when client is initialized with an empty key
+     *
+     * @throws IllegalArgumentException
+     */
+    @Test(expected = IllegalArgumentException.class)
```
QuintinWillison, line 32: Nice use of JUnit in the proper manner (a rare thing in this codebase, alas). ❤️

QuintinWillison (approved):

PR #696, opened 2021-08-05
Fix ably-android integration tests

KacperKluka:
After updating `compileSdkVersion` to 30 the tests extending `AndroidTestCase` are broken because that class is deprecated and removed. I fixed those tests by refactoring them in a newer way of writing integration tests.
I also had to add support for Java 8 desugaring to make the tests run.

Unfortunately, 2 tests from `AndroidPushTest` are failing.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -41,6 +41,11 @@ android {
         consumerProguardFiles 'proguard.txt'
     }
 
+    compileOptions {
+        sourceCompatibility 1.8
+        targetCompatibility 1.8
+    }
+
     buildTypes {
         release {
             minifyEnabled false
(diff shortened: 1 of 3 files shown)


martin-morek (commented):
From what I can see it mostly about how we setup the tests, which is great. I would propose to remove `test_` prefix from test method names since it's not necessary anymore.

QuintinWillison:
> remove test_ prefix from test method names

I support that suggestion - given they should all be using the `Test` annotation. Reduces unnecessary repetition.

KacperKluka:
Prefixes removed https://github.com/ably/ably-java/pull/696/commits/542032865f160dc18ae212c4d64d330e314b709f

Still those 2 tests from `AndroidPushTest` are failing:
- `restore_non_nullary_event` with
```
expected collection to have size 0, got 1: [GotPushDeviceDetails]
```

- `push_AfterRegistrationUpdateFailed_migrate_to_AfterRegistrationSyncFailed` with 
```
expected object of class io.ably.lib.push.ActivationStateMachine$NotActivated to be instance of io.ably.lib.push.ActivationStateMachine$AfterRegistrationSyncFailed
```

Issue #394, opened 2018-06-07
Update Stats fields with latest MessageTraffic types


SimonWoolf:
cf https://github.com/ably/docs/pull/443

PR #698, opened 2021-08-06
Updated Stats fileds with the latest MessageTraffic types

martin-morek:
- Fixes [394](https://github.com/ably/ably-java/issues/394)
- Added missing fields to Stats based on [documentation](https://docs.ably.com/general/statistics/)

### DIFF
--- a/lib/src/main/java/io/ably/lib/types/Stats.java
+++ b/lib/src/main/java/io/ably/lib/types/Stats.java
@@ -89,6 +89,12 @@ public class Stats {
         public Map<String, ProcessedCount> delta;
     }
 
+    public static class PushedMessages {
+        public int messages;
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/types/Stats.java
```diff
@@ -125,5 +133,5 @@ public static long fromIntervalId(String intervalId) {
     public ResourceCount channels;
     public RequestCount apiRequests;
     public RequestCount tokenRequests;
-    public ProcessedMessages processed;
```
QuintinWillison, line 128: As a result of this change do you need to remove the `ProcessedMessages` class?

QuintinWillison (changes requested):

Issue #430, opened 2018-11-06
WebsocketNotConnectedException on send


SimonWoolf:
Customer getting `WebsocketNotConnectedException` on send. Not sure what guarantees java gives, but if there's any possibility that `onClose` isn't called synchronously with the websocket closing, then that exception should be caught and treated as an abnormal close (it should call `onClose` itself and finish the transport with REASON_DISCONNECTED).

cf https://github.com/ably/ably-js/issues/397

PR #701, opened 2021-08-13
Separate handling WebsocketNotConnectedException

martin-morek:
- fixes [430](https://github.com/ably/ably-java/issues/430)
- Added separate handling for `WebsocketNotConnectedException` in `send()` method which terminates connection

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -17,6 +17,7 @@ import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
 
 import org.java_websocket.client.WebSocketClient;
+import org.java_websocket.exceptions.WebsocketNotConnectedException;
 import org.java_websocket.framing.CloseFrame;
 import org.java_websocket.framing.Framedata;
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -112,7 +113,12 @@ public void send(ProtocolMessage msg) throws AblyException {
                     Log.v(TAG, "send(): " + new String(ProtocolSerializer.writeJSON(msg)));
                 wsConnection.send(ProtocolSerializer.writeJSON(msg));
             }
-        } catch (Exception e) {
+        }
+        catch (WebsocketNotConnectedException e){
+            AblyException ablyException = AblyException.fromThrowable(e);
+            connectListener.onTransportUnavailable(this, ablyException.errorInfo);
```
QuintinWillison, line 119: For what it's worth, the `ablyException` local variable is pretty superfluous here - I would just use `AblyException.fromThrowable(e)` at the call site. 🤷‍♂️ 

Also, would it not be a good idea to check whether `connectListener` is `null` here?

Also, are there any threading considerations (i.e. contract with consumers) made around `ConnectListener` that might suggest calling it synchronously here might cause further issues?

QuintinWillison (changes requested):

PR #718, opened 2021-09-24
Add steps to build AAR locally and to use it in another project locally

ben-xD:


CONTRIBUTING.md
```diff
@@ -168,6 +168,42 @@ End-to-end tests for push notifications (ie where the Android client is the targ
...
+  - Export a secret key ring file: `gpg --export-secret-keys -o ably-java-secring.gpg`.
+  - Add the details of the GPG key pair inside `./gradle/gradle.properties`:
+```bash
+signing.keyId=XXXXXXXX
+signing.password=ably-debug-key
+signing.secretKeyRingFile=/Users/username/.ably/ably-java-secring.gpg
+```
+- Run `./gradlew android:assembleRelease` or `./gradlew android:assembleDebug`.
+
+## Using Ably Java / Ably Android locally in other projects
```
QuintinWillison, line 184: Given you're referring to them by artifact name, it's worth formatting these intentionally so:

```suggestion
## Using `ably-java` / `ably-android` locally in other projects
```

CONTRIBUTING.md
```diff
@@ -168,6 +168,42 @@ End-to-end tests for push notifications (ie where the Android client is the targ
...
+
+- Build the AAR: See [Building an AAR](#building-an-aar)
```
QuintinWillison, line 186: This link does not work as the heading appears to have changed since you generated it.

```suggestion
- Build the AAR: See [Building an AAR locally](#building-an-aar-locally)
```

QuintinWillison (changes requested):
This is really helpful - thank you for capturing this as you went along with your current work.

Beyond the comments I've made here, I would also suggest that it might be helpful to have an introductory paragraph in each of the two new sections you've added that explains _why_ you may need to do this. This could include an example use-case - perhaps even refer to `ably-flutter` in that.

QuintinWillison (approved):

PR #724, opened 2021-09-30
Fix indentation and typos in authCallback example

QuintinWillison:


README.md
```diff
@@ -240,14 +240,14 @@ Callback that provides either tokens (`TokenDetails`), or signed token requests
...
-                // process parms and return what is needed
-                return null;
-            }
-        };
+options.authCallback = new Auth.TokenCallback() {
+    @Override
+    public Object getTokenRequest(Auth.TokenParams params) {
+        System.out.println("Token Params: " + params);
+        // TODO: process params
+        return null; // TODO: return TokenDetails or TokenRequest
```
KacperKluka, line 248: Can't you also return a JWT as a `String`? :thinking:

KacperKluka (approved):
LGTM :+1: Just one small suggestion

README.md
sacOO7: Yes, you can return pure token as a `JWT` string, which later gets wrapped inside  empty `TokenDetails` object (which is not a JWT)

QuintinWillison: ok, so my focus when making this PR was indentation and spacing.

Perhaps @KacperKluka and @sacOO7, one of you could suggest a change that I should make to address the point around `String` support?

sacOO7: We will not need it. We already check internally if object is instance of `String`
https://github.com/ably/ably-java/blob/f2ec2844028b226adc7c80fcd39e2779bbcbde5d/lib/src/main/java/io/ably/lib/rest/Auth.java#L581-L587
Maybe we can modify todo to return string token.

KacperKluka: So maybe let's append that info to the TODO comment? Something like
```java
return null; // TODO: return TokenDetails or TokenRequest or JWT string
```

martin-morek (approved):

Issue #703, opened 2021-08-23
Remove references to GCM and simplify code


ben-xD:
We should remove references to GCM and simplify code.

For example, users currently have to specify 
```java
ActivationContext.getActivationContext(this).onNewRegistrationToken(RegistrationToken.Type.FCM, registrationToken);
```

Because there is only one option on Android, FCM, this should just be:
```java
ActivationContext.getActivationContext(this).onNewRegistrationToken(registrationToken);
```

All references to GCM should be removed, including e.g. `@Deprecated GCM`. GCM is not just deprecated, it has been killed/ shutdown.

QuintinWillison:
@ikbalkaya @KacperKluka @qsdigor [Another](https://github.com/ably/ably-java/issues/501#issuecomment-1189945074) one to consider for [our 2.0.0 milestone](https://github.com/ably/ably-java/milestone/7)? From #727 it seems we concluded this had to be a breaking change.

qsdigor:
@QuintinWillison Can you confirm that we are going with braking change and it is ok to just remove GCM references, tests, and every mention of it?

QuintinWillison:
Yes, that is my understanding, thanks. 😁

PR #727, opened 2021-10-21
Simplified code to use FCM token type as default

martin-morek:
- fixes [#703](https://github.com/ably/ably-java/issues/703)
- By default FCM token type will be used, deprecated GCM type is kept only for cases when it's loaded from storage

### DIFF
--- a/android/src/main/java/io/ably/lib/push/LocalDevice.java
+++ b/android/src/main/java/io/ably/lib/push/LocalDevice.java
@@ -74,7 +74,7 @@ public class LocalDevice extends DeviceDetails {
         }
 
         final String tokenType = recipient.get("transportType").getAsString();
-        if (!RegistrationToken.TOKEN_TYPE_FCM.equals(tokenType)) {
+        if (!RegistrationToken.TOKEN_TYPE_FCM.equalsIgnoreCase(tokenType)) {
             // FCM is only supported as we do not support other types.
             Log.e(TAG, "getRegistrationToken(): returning null because push.recipient.transportType is not FCM (it's \"" + tokenType + "\")");
             return null;


android/src/main/java/io/ably/lib/types/RegistrationToken.java
```diff
@@ -6,11 +6,20 @@
...
         this.type = type;
         this.token = token;
     }
 
+    public RegistrationToken(String token) {
+        this.type = Type.FCM;
+        this.token = token;
+    }
+
     public enum Type {
```
ben-xD, line 23: Could you deprecate this entire enum?

android/src/main/java/io/ably/lib/push/ActivationContext.java
```diff
@@ -95,21 +95,21 @@ public boolean setClientId(String clientId, boolean propagateGotPushDeviceDetail
...
-        Log.v(TAG, "onNewRegistrationToken(): type=" + type + ", token=" + token);
+    public void onNewRegistrationToken(String token) {
+        Log.v(TAG, "onNewRegistrationToken(): type=" + RegistrationToken.Type.FCM + ", token=" + token);
         LocalDevice localDevice = getLocalDevice();
         RegistrationToken previous = localDevice.getRegistrationToken();
         if (previous != null) {
-            if (previous.type != type) {
-                Log.e(TAG, "trying to register device with " + type + ", but it was already registered with " + previous.type);
+            if (previous.type != RegistrationToken.Type.FCM) {
+                Log.e(TAG, "trying to register device with " + RegistrationToken.Type.FCM + ", but it was already registered with " + previous.type);
```
ben-xD, line 104: If this happens, what can a user do to resolve it? Or perhaps instead of doing nothing (just returning), we could call `localDevice.setAndPersistRegistrationToken(new RegistrationToken(token));`.

ben-xD (commented):
Just a couple of comments, thanks :)

android/src/main/java/io/ably/lib/push/ActivationContext.java
martin-morek: yes sure, if that's appropriate handling

android/src/main/java/io/ably/lib/types/RegistrationToken.java
ben-xD: I'm not sure what you mean by:
> Token type GCM is deprecated, enum will be **replaced with constant**

We are deprecating it because users should just use the method with single parameter only: `public void onNewRegistrationToken(String token)`.

android/src/main/java/io/ably/lib/push/ActivationContext.java
ben-xD: This would have to be validated to work. Would realtime servers be updated in response to the new token? Looking at the code, I don't think so. Therefore, just removing the `return` statement will not be a good idea? What do you think?

android/src/main/java/io/ably/lib/types/RegistrationToken.java
martin-morek: That enum will be removed at all since it will be obsolete but we will still need string representation of `FCM` so something like `public static final String FCM = "FCM";` will replace it. However I can remove it from comment.

QuintinWillison:
Emulation failing. 🙄 ... I'm going to have a play in https://github.com/ably/ably-java/pull/738 to see if I can get it more stable, before coming back to this PR.

QuintinWillison:
Emulation run [failed](https://github.com/ably/ably-java/runs/4872716194?check_suite_focus=true):

```
io.ably.lib.test.android.AndroidPushTest > push_syncOnStartup[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: expected null, but was:<GotPushDeviceDetails>
	at org.junit.Assert.fail(Assert.java:88)

io.ably.lib.test.android.AndroidPushTest > NotActivated_on_CalledActivate_with_registrationToken[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: expected collection to have size 1, got 0: [] expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
```

QuintinWillison:
Emulation run [failed](https://github.com/ably/ably-java/runs/4874008576?check_suite_focus=true):

```
io.ably.lib.test.android.AndroidPushTest > push_syncOnStartup[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: expected null, but was:<GotPushDeviceDetails>
	at org.junit.Assert.fail(Assert.java:88)

io.ably.lib.test.android.AndroidPushTest > NotActivated_on_CalledActivate_with_registrationToken[test(AVD) - 7.0] FAILED 
	java.lang.AssertionError: expected collection to have size 1, got 0: [] expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
```

QuintinWillison:
Given the same test failed twice in a row I think it's back to the drawing board with this PR, alas. 😢

qsdigor:
I have removed all instances and mentions of GCM as it is long past deprecated and removed in 2019.

android/src/main/java/io/ably/lib/push/LocalDevice.java
```diff
@@ -54,38 +56,38 @@ private void loadPersisted() {
...
             return null;
         }
-        Log.v(TAG, "getRegistrationToken(): returning a new registration token because push.recipient is set");
-        return new RegistrationToken(
-            RegistrationToken.Type.fromName(recipient.get("transportType").getAsString()),
-            recipient.get("registrationToken").getAsString()
-        );
+
+        final String tokenType = recipient.get("transportType").getAsString();
+        if (!RegistrationToken.TOKEN_TYPE_FCM.equals(tokenType)) {
```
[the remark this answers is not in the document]
qsdigor: Do you want us to use equalsIgnoreCase instead of equals?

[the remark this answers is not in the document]
qsdigor: Here you go https://github.com/ably/ably-java/pull/727/commits/c377629c54f26a5e3a849703b5393115532fda34

KacperKluka (approved):
LGTM :+1:

Issue #501, opened 2019-10-18
JSONUtilsObject.add() silently discards data of unsupported type


paddybyers:
If `value` has a type that doesn't translate naturally to a JSON value, then it is silently discarded: https://github.com/ably/ably-java/blob/develop/lib/src/main/java/io/ably/lib/util/JsonUtils.java#L34

This should at least log that event, and possibly throw an exception.

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/SDK-434) by [Unito](https://www.unito.io)

amsurana:
Pushed a PR #502

QuintinWillison:
@ikbalkaya @KacperKluka @qsdigor I was just perusing our open, draft PRs and noticed #729. I expect it may well be the case that that particular PR is no longer viable, it probably needs rework / conflicts addressing - but what do we think to including this issue in [our 2.0.0 milestone](https://github.com/ably/ably-java/milestone/7)? It feels like a 'quick win' for correctness, right? 🤔

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-434

Issue #573, opened 2020-04-28
Should GSON be an API dependency?


QuintinWillison:
Reported by a customer ([internal conversation](https://ably-real-time.slack.com/archives/C012FG295QD/p1588006502037500)). In their words, given this tripped them up...

Initial Report:

> class `HttpPaginatedResponse`, returned by `AblyRest.request` method returns an array of `com.google.gson.JsonElement` from its `items()` method, yet the ably-java library (v1.1.10) declares only a runtime dependency on GSON library, leading my IDE to flag calls to the items() method as an error with message ‘Cannot access com.google.gson.JsonElement’. Interested to know why GSON is not a compile-time dependency given its classes are used in the public API of ably-java?

Follow-up opinion as to how this should be instead:

> `implementation` results in those dependencies being declared as ‘runtime’ dependencies in the published POM, hence they are transitively included in the consuming applications runtimeClasspath but not the compileClasspath. Personally, I would consider having GSON as an `api` dependency in the project (this would also require use of the Gradle ‘java-library’ plugin rather than the plain ‘java’ plugin), meaning it would end up as ‘compile’ dependency in the ably-java POM and thus be available in the compileClasspath of consuming applications. I suspect what may have happened is that when the original Gradle `compile` scope (which resulted in dependencies being published as ‘compile’ scope in the POM) was deprecated, someone simply replaced it with `implementation` (as I originally did in all my personal projects!) without fully appreciating the subtle difference between `implementation` and `api`.

> Like most things, no right or wrong (there are downsides to using `api`), but maybe worth thinking about. https://docs.gradle.org/current/userguide/java_library_plugin.html

QuintinWillison:
I've reproduced this in IntelliJ IDEA CE when working with a brand new Gradle project. I have the ably-java Java jar in a `libs` folder and introduced it to the project using:

```gradle
dependencies {
    implementation fileTree(include: ['*.jar'], dir: 'libs')
}
```

It builds fine:

```
hello-deltas-java % ./gradlew build

BUILD SUCCESSFUL in 638ms
5 actionable tasks: 4 executed, 1 up-to-date
```

but crashes at runtime:

```
hello-deltas-java % ./gradlew run

> Task :run FAILED
Ably Client Library: java v.1.2.0
Exception in thread "main" java.lang.NoClassDefFoundError: com/google/gson/JsonElement
        at io.ably.lib.rest.Auth.setTokenDetails(Auth.java:928)
[...]

PR #729, opened 2021-10-26
Throw exception on unsupported json type

martin-morek:
- fixes [#501](https://github.com/ably/ably-java/issues/501)
- when object type doesn't match any json supported value type `JsonParseException` is thrown


QuintinWillison (approved):
I wasn't expecting this to be so simple to fix as my worry was that there might be other parts of the codebase that relied upon this behaviour. I'm going to add @ikbalkaya to this review for a second pair of eyes.

lib/src/main/java/io/ably/lib/util/JsonUtils.java
```diff
@@ -30,7 +31,10 @@ public JsonUtilsObject add(String key, Object value) {
                 json.addProperty(key, (Number) value);
             } else if (value instanceof JsonUtilsObject) {
                 json.add(key, ((JsonUtilsObject) value).toJson());
+            } else {
+                throw new JsonParseException("Unsupported Object Type");
```
ikbalkaya, line 35: I have quickly gone through the reference calls to this method and apart from test classes this seems to be used in `DeviceDetails`s `pushRecipientJsonObject` method. But I don't know whether this might be triggered by a client mistake. If so I think it would be better to. have checked exception, but if this is only something that library code will cause, then unchecked exception is fine

ikbalkaya (changes requested):
I have requested some clarification on the impact of unchecked exception, a code change is not necessarily required

lib/src/main/java/io/ably/lib/util/JsonUtils.java
KacperKluka: This method is used not only in our internal code but also it is used by our clients. The `JsonUtils` is mentioned both in our [README](https://github.com/ably/ably-java#using-the-push-api) and [official docs](https://ably.com/documentation/general/push/publish#channel-broadcast-example). 

Given that, how do we proceed with this issue? Do we want to add a checked exception here which will basically be a change in the API (all clients that use Push API will have to make changes to their code and we probably will need to update the code examples in the docs, right)? Or do we want to get rid of the exception entirely and simply add a log statement informing that a value was not used? :thinking: WDYT @ikbalkaya @QuintinWillison

QuintinWillison: If it's a change in API then that's a 'breaking change', in which case we'll not be able to do that until this SDK makes a release where we can break stuff.

What we can do in the meantime, however, is introduce a new API that does the same thing and deprecate this API. This could be as simple as a new method on `JsonUtils`, or it could be a brand new structure (i.e. class or class+interface set). This would be my preferred approach to solving this.

qsdigor: @KacperKluka @ikbalkaya Where are we on this?

KacperKluka: I'd either change this PR's target branch to the `2.0.0` milestone and make that a breaking change (remembering about updating docs as well) or implement the solution from [this comment](https://github.com/ably/ably-java/pull/729#discussion_r835106748).

qsdigor: I am going with "breaking change" as we are moving to v2.0.0 and it seems as the only correct way to do this.

qsdigor: I have updated readme https://github.com/ably/ably-java/pull/729/commits/838487b1dfe2c7c2a0c5b855a050cf4983ada18b

README.md
```diff
@@ -453,6 +454,9 @@ rest.push.admin.publishAsync(recipient, payload, , new CompletionListener() {
  });
 ```
 
+Note that only JsonElement, String, Boolean, Character, Number, and JsonUtilsObject are supported by JsonUtils.
+If an unsupported type is added JsonUtils will throw JsonParseException!
```
KacperKluka, line 458: I'm not sure if that exclamation mark at the end is necessary :sweat_smile: 
```suggestion
If an unsupported type is added JsonUtils will throw JsonParseException.
```

core/src/main/java/io/ably/lib/util/JsonUtils.java
```diff
@@ -30,7 +31,10 @@ public JsonUtilsObject add(String key, Object value) {
                 json.addProperty(key, (Number) value);
             } else if (value instanceof JsonUtilsObject) {
                 json.add(key, ((JsonUtilsObject) value).toJson());
+            } else {
+                throw new JsonParseException("Unsupported Object Type");
```
KacperKluka, line 35: Keeping in mind [the issue about removing GSON from our public API](https://github.com/ably/ably-java/issues/573), maybe let's not use the `JsonParseException` that's coming from the `gson` library, but instead create our own exception? Or use some other exception that's not coming from a third-party library? :thinking:

KacperKluka (changes requested):
A few questions and suggestions :wink:

ikbalkaya (changes requested):
Just a small readability improvement request if possible please

README.md
```diff
@@ -409,16 +409,17 @@ To publish a message to a channel including a push payload:
 
 ```java
 Message message = new Message("example", "realtime data");
-message.extras = io.ably.lib.util.JsonUtils.object()
+JsonObject jsonObject = io.ably.lib.util.JsonUtils.object()
```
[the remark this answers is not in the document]
qsdigor: Here it is https://github.com/ably/ably-java/pull/729/commits/f2fcfac69acf9baf0da9088a1934a59e2394da9e

README.md
```diff
@@ -409,16 +409,17 @@ To publish a message to a channel including a push payload:
...
     .add("push", io.ably.lib.util.JsonUtils.object()
         .add("notification", io.ably.lib.util.JsonUtils.object()
             .add("title", "Hello from Ably!")
             .add("body", "Example push notification from Ably."))
         .add("data", io.ably.lib.util.JsonUtils.object()
             .add("foo", "bar")
-            .add("baz", "qux")));
+            .add("baz", "qux"))).toJson();
```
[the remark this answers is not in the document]
qsdigor: here it is https://github.com/ably/ably-java/pull/729/commits/1820e44c901e011075d91c7edd249a237f3ea0ab

README.md
qsdigor: here it is https://github.com/ably/ably-java/pull/729/commits/1820e44c901e011075d91c7edd249a237f3ea0ab

core/src/main/java/io/ably/lib/util/JsonUtils.java
qsdigor: here it is https://github.com/ably/ably-java/pull/729/commits/1820e44c901e011075d91c7edd249a237f3ea0ab

ikbalkaya (approved):

core/src/main/java/io/ably/lib/util/JsonUtils.java
```diff
@@ -30,7 +30,10 @@ public JsonUtilsObject add(String key, Object value) {
                 json.addProperty(key, (Number) value);
             } else if (value instanceof JsonUtilsObject) {
                 json.add(key, ((JsonUtilsObject) value).toJson());
+            } else {
+                throw new IllegalArgumentException("Unsupported Object Type");
```
KacperKluka, line 34: Please add a javadoc to this method that would say it throws this exception :wink: Something like
```java
/**
 * Adds the key-value pair to the JSON object. 
 * If the value type is unsupported an IllegalArgumentException is thrown.
 */
```

KacperKluka (changes requested):
One last thing :grin:

core/src/main/java/io/ably/lib/util/JsonUtils.java
qsdigor: done https://github.com/ably/ably-java/pull/729/commits/196314e6e2539b470618beaa0760426349112141

KacperKluka (approved):
LGTM :+1:

PR #751, opened 2022-03-09
Fix connection example in README

owenpearson:


QuintinWillison (approved):

QuintinWillison:
Merging despite only one review, due to the fact that the change is low risk and the fix obvious.

PR #763, opened 2022-05-02
Use only the clientId and data of the original presence message when automatically re-entering

KacperKluka:
Previously, we were copying the whole presence message and overriding the action to `ENTER`. However, [the spec](https://docs.ably.io/client-lib-development-guide/features/#RTP17d) requires to only copy the `clientId` and `data`. After this change we no longer receive the `Malformed message; mismatched connectionId` error.

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/Presence.java
+++ b/lib/src/main/java/io/ably/lib/realtime/Presence.java
@@ -254,7 +254,13 @@ public class Presence {
                     /* Message is new to presence map, send it */
                     final String clientId = item.clientId;
                     try {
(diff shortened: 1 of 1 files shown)


QuintinWillison (commented):
Is this testable?

KacperKluka:
@QuintinWillison it should be testable but with the current setup it's not easy, therefore, to not block the fix release, I've created an issue for adding the tests https://github.com/ably/ably-java/issues/765 :wink:

QuintinWillison (approved):
@KacperKluka is going to create an issue in the backlog to look at testing around this. We are sure of the fix and there is a customer waiting for it.

PR #785, opened 2022-05-18
Update Gradle version

KacperKluka:
I've updated the Gradle version and Android Gradle plugin version. After upgrading from `4.x` to `7.x` I had to use another BuildConfig plugin (as the old one was last time updated 5 years ago) and change the test sharing configuration (as the `testRuntime` was removed in Gradle 7).

While I was replacing the BuildConfig plugin I noticed that we only use it in the `lib` module, so I've removed it from the other ones. I also had to exclude it from the checkstyle as it was complaining about the generated code and that's something we can't change.

After the upgrade the Gradle lint tool was complaining about variables naming conventions, so I've changed the `SNAKE_CASE` names to `camelCase` ones.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -1,6 +1,6 @@
 package io.ably.lib.transport;
 
-import io.ably.BuildConfig;
+import io.ably.lib.BuildConfig;
 import io.ably.lib.types.ClientOptions;
 
 import java.text.DecimalFormat;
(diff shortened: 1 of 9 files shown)


lib/src/main/java/io/ably/lib/transport/Defaults.java
```diff
@@ -1,6 +1,6 @@
 package io.ably.lib.transport;
 
-import io.ably.BuildConfig;
+import io.ably.generated.BuildConfig;
```
KacperKluka, line 3: This change should be reflected in the API updating guide as mentioned in this [comment from another PR](https://github.com/ably/ably-java/pull/768#discussion_r881591978)

qsdigor (approved):

QuintinWillison (approved):
🥳

PR #788, opened 2022-05-18
Improve Gradle setup

KacperKluka:
I've implemented the same setup that we've used for the AAT project. Thanks to that I got rid of `common.gradle` and `dependencies.gradle` files.

It turns out that ably-java project doesn't share that much of a gradle configuration, so the benefit of having it in the root `build.gradle` is limited when compared to the AAT setup. However, it is still an improvement that (hopefully) makes the project better.


qsdigor (approved):

QuintinWillison (approved):
Nice. You know me... I love a bit of 'creative' Gradle. 😜

PR #792, opened 2022-05-26
Change visibility of SDK dependencies

KacperKluka:
- Classes from `lib` are now visible when someone uses `ably-java`
- Classes from `gson` are now visible (unfortunately, it is used in our API)
- Hidden classes from `msgpack` from the public API by changing utility methods visibility

### DIFF
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -11,7 +11,7 @@ sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
 dependencies {
-    implementation project(':lib')
+    api project(':lib')
     // Enable access to test classes from the :lib module for integration test cases sharing
     testImplementation project(path: ':lib', configuration: 'sharedTestClasses')
 }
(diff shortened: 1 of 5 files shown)


qsdigor (approved):

java/build.gradle
```diff
@@ -11,7 +11,7 @@ sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
 dependencies {
-    implementation project(':lib')
+    api project(':lib')
```
QuintinWillison, line 14: Sorry that I am noticing this so late on in the evolution of this work, but my knee-jerk level dislike for abbreviations like `lib` is kicking in. I also reflect on our use of `common` and `core` [in AAT](https://github.com/ably/ably-asset-tracking-android/blob/main/settings.gradle).

I think we should conform this project/module name to reflect closer what we feel it eventually means for customers, where that name is more reflective of the fact that it's common API. Perhaps `core` would be appropriate? 🤔

QuintinWillison (changes requested):

java/build.gradle
KacperKluka: Good point, I've created an issue for that as this feels outside of scope for this PR :wink: https://github.com/ably/ably-java/issues/811

QuintinWillison (approved):

PR #798, opened 2022-06-15
Update dependencies that contain known vulnerabilities

KacperKluka:
We've found out that the `gson` and `Java-WebSocket` dependency versions have known vulnerabilities. Therefore, we should update them to use the new, secure versions.


QuintinWillison (approved):

qsdigor:
Looks good!

qsdigor (approved):

PR #807, opened 2022-06-20
Add `finalize()` and `AutoCloseable` support to `AblyRest` instances

### LINKED ISSUE #801: Threads remain in parked (waiting) state indefinitely when `AblyRest` instance is freed
We've had a [report](https://ably-real-time.slack.com/archives/C02C0CNCPTK/p1655701678874329?thread_ts=1655536688.502499&cid=C02C0CNCPTK) from a customer:

> we have noticed that when an instance of the AblyRest is replaced by another instance (as part of regenerating the client), the threads held by the old instance is not getting released or destroyed. This causes the threads to keep increasing as new instances of AblyRest gets created.
>
> The threads are managed by the ThreadpoolExecutor in the AsyncHttpScheduler and a max of 64 (default) can get created. They are placed in parked (waiting) state indefinitely. These threads continue to remain in this state even after the AblyRest instance in which they belong to is removed.
>
> It seems the AsyncHttpScheduler.dispose method is not getting called in this case.
>
> This is causing the jvm memory usage to slowly increase over time.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2062) by [Unito](https://www.unito.io)

QuintinWillison:
A dual pronged approach to resolve #801.

Either:

- The addition of `finalize()` implementation in https://github.com/ably/ably-java/commit/f29e497760b5bfcbe95271d9c80e0303c391e787 will _passively_ clean up the thread pool executor once GC catches cleans up the `AsyncHttpScheduler` instance (once it's no longer referenced by the `AblyRest` instance that created it)
- The adoption of `AutoCloseable` in https://github.com/ably/ably-java/commit/56593e5822b2baf808bbcbe36e1d9b2940dacdbf will allow app developers to explicitly call `close()` on their `AblyRest` instances, or use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
@@ -13,17 +13,17 @@ import io.ably.lib.util.Log;
  */
 public class AsyncHttpScheduler extends HttpScheduler {
     public AsyncHttpScheduler(HttpCore httpCore, ClientOptions options) {
-        super(httpCore, new WrappedExecutor(options));
+        super(httpCore, new CloseableThreadPoolExecutor(options));
     }
 
     private static final long KEEP_ALIVE_TIME = 2000L;
 
     protected static final String TAG = AsyncHttpScheduler.class.getName();
 
-    private static class WrappedExecutor implements CloseableExecutor {
+    private static class CloseableThreadPoolExecutor implements CloseableExecutor {
         private final ThreadPoolExecutor executor;
 
-        WrappedExecutor(final ClientOptions options) {
+        CloseableThreadPoolExecutor(final ClientOptions options) {
             executor = new ThreadPoolExecutor(
                 options.asyncHttpThreadpoolSize,
                 options.asyncHttpThreadpoolSize,
--- a/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
+++ b/lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
@@ -14,18 +14,7 @@ public class AsyncHttpScheduler extends HttpScheduler<ThreadPoolExecutor> {
         super(httpCore, new ThreadPoolExecutor(options.asyncHttpThreadpoolSize, options.asyncHttpThreadpoolSize, KEEP_ALIVE_TIME, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>()));
     }
 
-    public void dispose() {
-        ThreadPoolExecutor threadPoolExecutor = executor;
-        threadPoolExecutor.shutdown();
-        try {
-            threadPoolExecutor.awaitTermination(SHUTDOWN_TIME, TimeUnit.MILLISECONDS);
-        } catch (InterruptedException e) {
-            threadPoolExecutor.shutdownNow();
-        }
-    }
-
     private static final long KEEP_ALIVE_TIME = 2000L;
-    private static final long SHUTDOWN_TIME = 5000L;
 
     protected static final String TAG = AsyncHttpScheduler.class.getName();
 }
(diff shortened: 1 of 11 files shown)


lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
```diff
@@ -6,27 +6,49 @@
...
-        }
+        super(httpCore, new WrappedExecutor(options));
     }
 
     private static final long KEEP_ALIVE_TIME = 2000L;
-    private static final long SHUTDOWN_TIME = 5000L;
 
     protected static final String TAG = AsyncHttpScheduler.class.getName();
+
+    private static class WrappedExecutor implements CloseableExecutor {
```
KacperKluka, line 23: I think the name of this class can be improved, now we only know that the executor is wrapped, but we don't know why. I'd like to include the reason behind the wrapper in this class name as well. Maybe something like `ClosableThreadPoolExecutor` or something similar? :thinking:

qsdigor (approved):

lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
QuintinWillison: That's a good point. I was struggling with naming. I shall adopt your suggestion. Thanks!

QuintinWillison: https://github.com/ably/ably-java/pull/807/commits/dce84ef20d89edbd663e6996371257a35eec6549

KacperKluka (approved):
LGTM :+1:

lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
```diff
@@ -6,27 +6,49 @@
...
+        }
+
+        @Override
+        public void execute(final Runnable command) {
+            executor.execute(command);
+        }
+
+        @Override
+        public void close() throws Exception {
+            final int drainedCount = executor.shutdownNow().size();
```
ikbalkaya, line 43: Will this not shut down all running threads in the thread pool? Isn't that we only want to shut down waiting threads 🤔

ikbalkaya (commented):
I made a comment on a line, otherwise looks good.

lib/src/main/java/io/ably/lib/http/AsyncHttpScheduler.java
QuintinWillison: Yes, that's the intention. I'm trying to keep the implementation as simple as possible. There was [some implementation previously (unused, as it happened)](https://github.com/ably/ably-java/pull/807/commits/6705060846c3a9f8b2102528476afa5bac157fb4) which appeared to want to try to 'gracefully' shut things down - but that is a risky strategy, especially when this API is synchronous and there's no clear structure as to who/what would hold a strong reference to things while they close down anyway. Therefore, I've taken the view that an explicit `close()` like this means that the user is not expecting any further responses back from this client instance.

ikbalkaya: Sorry if I am missing the point of usage of this class. So for example when there are some threads that are running (eg. sending requests) will there be a case that they would be shutdown before they got a chance to complete?

ikbalkaya: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdown()
From shutdownNow doc
> Attempts to stop all actively executing tasks, halts the processing of waiting tasks, and returns a list of the tasks that were awaiting execution.

My question is probably be better formed as - What would happen to those tasks that were awaiting execution?

QuintinWillison: > So for example when there are some threads that are running (eg. sending requests) will there be a case that they would be shutdown before they got a chance to complete?

Correct. That is my intention.

ikbalkaya (approved):

PR #808, opened 2022-06-21
Fix Java-WebSocket problem on Android below 24

KacperKluka:
There's a [known issue](https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm) in the Java-WebSocket library that has affected us after the recent upgrade. Luckily, there's a [workaround](https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm#workaround) that we have added to make it work.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -164,7 +164,7 @@ public class WebSocketTransport implements ITransport {
          * https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm#workaround
          */
         private boolean isHostnameVerified(String hostname) {
-            SSLSession session = getSSLSession();
+            final SSLSession session = getSSLSession();
             if (HttpsURLConnection.getDefaultHostnameVerifier().verify(hostname, session)) {
                 Log.i(TAG, "Successfully verified hostname");
                 return true;
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -13,7 +13,10 @@ import java.nio.ByteBuffer;
 import java.util.Timer;
 import java.util.TimerTask;
 
+import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSession;
 
 import org.java_websocket.client.WebSocketClient;
 import org.java_websocket.exceptions.WebsocketNotConnectedException;
@@ -147,8 +150,29 @@ public class WebSocketTransport implements ITransport {
         @Override
         public void onOpen(ServerHandshake handshakedata) {
             Log.d(TAG, "onOpen()");
-            connectListener.onTransportAvailable(WebSocketTransport.this);
-            flagActivity();
+            if (isHostnameVerified(params.host)) {
+                connectListener.onTransportAvailable(WebSocketTransport.this);
+                flagActivity();
+            } else {
+                connectListener.onTransportUnavailable(WebSocketTransport.this, ConnectionManager.REASON_REFUSED);
+                close();
+            }
+        }
+
+        /**
+         * Added because we had to override the onSetSSLParameters() that usually performs this verification.
+         * When the minSdkVersion will be updated to 24 we should remove this method and its usages.
+         * https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm#workaround
+         */
+        private boolean isHostnameVerified(String hostname) {
+            SSLSession session = getSSLSession();
+            if (HttpsURLConnection.getDefaultHostnameVerifier().verify(hostname, session)) {
+                Log.i(TAG, "Successfully verified hostname");
+                return true;
+            } else {
+                Log.e(TAG, "Hostname verification failed, expected " + hostname + ", found " + session.getPeerHost());
+                return false;
+            }
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -147,8 +150,29 @@ class WsClient extends WebSocketClient {
...
         public void onOpen(ServerHandshake handshakedata) {
             Log.d(TAG, "onOpen()");
-            connectListener.onTransportAvailable(WebSocketTransport.this);
-            flagActivity();
+            if (isHostnameVerified(params.host)) {
+                connectListener.onTransportAvailable(WebSocketTransport.this);
+                flagActivity();
+            } else {
+                connectListener.onTransportUnavailable(WebSocketTransport.this, ConnectionManager.REASON_REFUSED);
+                close();
```
KacperKluka, line 158: @QuintinWillison I'm not sure if we should call this `close()` here as it will result in calling the `connectListener.onTransportUnavailable()` once again from withing `onClose()` method. WDYT?

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -147,8 +150,29 @@ class WsClient extends WebSocketClient {
...
+
+        /**
+         * Added because we had to override the onSetSSLParameters() that usually performs this verification.
+         * When the minSdkVersion will be updated to 24 we should remove this method and its usages.
+         * https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm#workaround
+         */
+        private boolean isHostnameVerified(String hostname) {
+            SSLSession session = getSSLSession();
+            if (HttpsURLConnection.getDefaultHostnameVerifier().verify(hostname, session)) {
+                Log.i(TAG, "Successfully verified hostname");
```
QuintinWillison, line 170: I think `i` is too high - I would use `d` or even `v` here.

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -147,8 +150,29 @@ class WsClient extends WebSocketClient {
...
+            }
+        }
+
+        /**
+         * Added because we had to override the onSetSSLParameters() that usually performs this verification.
+         * When the minSdkVersion will be updated to 24 we should remove this method and its usages.
+         * https://github.com/TooTallNate/Java-WebSocket/wiki/No-such-method-error-setEndpointIdentificationAlgorithm#workaround
+         */
+        private boolean isHostnameVerified(String hostname) {
+            SSLSession session = getSSLSession();
```
QuintinWillison, line 168:
```suggestion
            final SSLSession session = getSSLSession();
```

please. 😉

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -234,6 +258,13 @@ public void onError(final Exception e) {
             connectListener.onTransportUnavailable(WebSocketTransport.this, new ErrorInfo(e.getMessage(), 503, 80000));
         }
 
+        @Override
+        protected void onSetSSLParameters(SSLParameters sslParameters) {
```
QuintinWillison, line 262: There's a niggling worry in the back of my head that we should perhaps consider making this more sophisticated, such that it will prefer the default implementation if we have no reason to expect it to fail... I recognise this is `lib/` code, but is there any way that we can detect if the `setEndpointIdentificationAlgorithm` method is available and call `super`'s implementation if it is? Presumably the subset of runtime devices without it is quite small. 🤔

QuintinWillison (changes requested):

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
QuintinWillison: I think I would drop the call to `onTransportUnavailable` and just let `close` do the right thing. I note that you already have a `Log.e` statement in the `isHostnameVerified` implementation, so there should be a clue left when this happens.

KacperKluka: https://github.com/ably/ably-java/pull/808/commits/2999d2a48e31b9fe2fe1175281c887419c5c7f66

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
KacperKluka: https://github.com/ably/ably-java/pull/808/commits/99927b0877ab57010520358da73625a6ba2b7a47

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
KacperKluka: https://github.com/ably/ably-java/pull/808/commits/8831ae3d391b000e48ae961701a8069a3fb6816f

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
KacperKluka: We can make it more general (as this is an old Java API version problem rather than an Android-specific problem) and call the `super.onSetSSLParameters()` in a `try / catch` block and try to catch the `NoSuchMethodError` if it is thrown :thinking: Then, if it is thrown we can send a warning (or info perhaps) to the log handler and continue, as the verification will happen in the `isHostnameVerified()` method. WDYT?
```java
try {
    super.onSetSSLParameters(sslParameters);
} catch (NoSuchMethodError exception) {
    Log.w(TAG, "Error when trying to set SSL parameters, most likely due to an old Java API version", exception);
}
```

qsdigor (approved):

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
QuintinWillison: I like that. Furthermore, could you enhance that to include some state that is set if the method call was successful, so that `isHostnameVerified()` can be skipped later on in the lifecycle?

KacperKluka: Good idea, here's the change :wink:  https://github.com/ably/ably-java/pull/808/commits/f35c5b6a11956637a4ed0a9eeebc554693ff8778

QuintinWillison (approved):

KacperKluka:
Even though the CI integration-test is failing we are going to land this PR because of [those reasons](https://github.com/ably/ably-java/pull/807#issuecomment-1163063730) :wink:

PR #812, opened 2022-06-24
Add multiple android emulation devices

qsdigor:
Added Android 19(KitKat) as SDK minimum supported and 29(Android 10) as the latest default target available in [Android emulator runner](https://github.com/ReactiveCircus/android-emulator-runner) to workflow.

### DIFF
--- a/.github/workflows/emulate.yml
+++ b/.github/workflows/emulate.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        android-api-level: [ 19, 24, 30 ]
+        android-api-level: [ 19, 24, 29 ]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/emulate.yml
+++ b/.github/workflows/emulate.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-      - feature/emulation_test
 
 jobs:
   check:
--- a/.github/workflows/emulate.yml
+++ b/.github/workflows/emulate.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/emulate.yml
+++ b/.github/workflows/emulate.yml
@@ -7,12 +7,17 @@ on:
 jobs:
   check:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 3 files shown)


qsdigor:
I have added API 19, 21, 24, and 29 as test devices in CI.
Some tests were fixed as well.

ikbalkaya (approved):
LGTM

.github/workflows/emulate.yml
```diff
@@ -6,14 +6,20 @@ on:
...
     steps:
       - uses: actions/checkout@v2
 
       - uses: reactivecircus/android-emulator-runner@v2
         with:
-          api-level: 24
-          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim
+          api-level: ${{ matrix.android-api-level }}
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
```
KacperKluka, line 22: Just out of interest, why you've added the `-camera-back none` and disabled the animations? Were they causing any issues? :thinking:

KacperKluka (approved):
LGTM :+1:

.github/workflows/emulate.yml
qsdigor: As emulators do not have real cameras and we do not use them, the best thing to do is disable camera emulation in the first place to improve performance. Disabling animations improves the performance of the emulator as well.

.github/workflows/emulate.yml
```diff
@@ -6,14 +6,20 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
QuintinWillison, line 9: Is there a reason for the change from Ubuntu to macOS?

This is ground we've covered before, where we intentionally changed from macOS to Ubuntu for a couple of reasons:

1. Test run stability
2. GitHub resource constraints

QuintinWillison (dismissed):
I have a question (sorry for the lag in me getting around to reviewing this one 😅)...

.github/workflows/emulate.yml
qsdigor: Yes there is, not all emulators are available in Ubuntu as macOS has HAXM installed so we are able to create a new AVD instance, launch an emulator with hardware acceleration, and run our Android tests directly on the VM.
This is described [here](https://github.com/ReactiveCircus/android-emulator-runner#github-action---android-emulator-runner).

qsdigor: sure, we will keep an eye on it.

Issue #814, opened 2022-07-01
Early return from  onAuthUpdated creates issues


ikbalkaya:
`AblyBase` and `ConnectionManager` have a method called `onAuthUpdated` that contains a `waitForResponse` and is used to decide whether to wait for async response before returning from method. 
This seems to be set to `false`  only when `Auth.renew()` is called. 

This feature doesn't seem to be on spec. Apart from this, this method is implicitly asynchronous. When the `waitForResponse` is set to `false` there is no opportunity for caller to know about the result. For our specific case we want to know when the token renewal is completed. As a result we either have to
* Remove this parameter and wait for response for all cases, so that it is safe to call it synchronously
or
* Provide a callback so that  callers are notified of completion. 

Please see [this thread](https://github.com/ably/ably-asset-tracking-android/pull/703/files#r907540279) to see where this discussion has started.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2385) by [Unito](https://www.unito.io)

PR #816, opened 2022-07-05
Add new renew async method

ikbalkaya:
On this PR,

I added a new `renewAuth` to `Auth` that accepts a token and a completion handler. I also deprecated the old `renew`. This is done to prevent callers blocked and provide a handler for completion. 

The essential change here is moving the "waiting for connection state change" part to a background thread and invoking a callback on completion

Related to https://github.com/ably/ably-asset-tracking-android/issues/695

Closes https://github.com/ably/ably-java/issues/814

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/AblyBase.java
+++ b/lib/src/main/java/io/ably/lib/rest/AblyBase.java
@@ -1,5 +1,6 @@
 package io.ably.lib.rest;
 
+import java.util.concurrent.Future;
 import io.ably.annotation.Experimental;
 import io.ably.lib.http.AsyncHttpScheduler;
 import io.ably.lib.http.Http;
@@ -314,6 +315,15 @@ public abstract class AblyBase implements AutoCloseable {
      */
     protected void onAuthUpdated(String token, boolean waitForResponse) throws AblyException {
         /* Default is to do nothing. Overridden by subclass. */
+    }
+
+    /**
+     * Override this method in AblyRealtime and pass updated token to ConnectionManager
+     * @param token new token
+     */
+    protected Future<AblyException> onAuthUpdatedAsync(String token)  {
+        //this must be overriden by subclass
+        return null;
     }
 
     /**
--- a/lib/src/main/java/io/ably/lib/rest/AblyBase.java
+++ b/lib/src/main/java/io/ably/lib/rest/AblyBase.java
@@ -321,7 +321,7 @@ public abstract class AblyBase implements AutoCloseable {
      * Override this method in AblyRealtime and pass updated token to ConnectionManager
      * @param token new token
      */
-    protected Future<AblyException> onAuthUpdatedAsync(String token)  {
+    protected Future<Void> onAuthUpdatedAsync(String token)  {
         //this must be overriden by subclass
         return null;
     }
--- a/lib/src/main/java/io/ably/lib/rest/AblyBase.java
+++ b/lib/src/main/java/io/ably/lib/rest/AblyBase.java
@@ -1,6 +1,5 @@
 package io.ably.lib.rest;
 
-import java.util.concurrent.Future;
 import io.ably.annotation.Experimental;
 import io.ably.lib.http.AsyncHttpScheduler;
 import io.ably.lib.http.Http;
@@ -320,10 +319,10 @@ public abstract class AblyBase implements AutoCloseable {
     /**
      * Override this method in AblyRealtime and pass updated token to ConnectionManager
      * @param token new token
+     * @param authUpdateResult Callback result
      */
-    protected Future<Void> onAuthUpdatedAsync(String token)  {
+    protected void onAuthUpdatedAsync(String token, Auth.AuthUpdateResult authUpdateResult)  {
         //this must be overriden by subclass
-        return null;
     }
 
     /**
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -827,7 +827,11 @@ public class Auth {
      * Renew auth credentials.
      * Will obtain a new token, even if we already have an apparently valid one.
      * Authorization will use the parameters supplied on construction.
+
+     * @deprecated this method is deprecated
+     * Please use  {@link Auth#renewAuth()} instead.
      */
+    @Deprecated
     public TokenDetails renew() throws AblyException {
         TokenDetails tokenDetails = assertValidToken(this.tokenParams, this.authOptions, true);
         ably.onAuthUpdated(tokenDetails.token, false);
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -3,11 +3,9 @@ package io.ably.lib.rest;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
-import java.util.AbstractMap;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.Future;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
@@ -37,7 +35,6 @@ import io.ably.lib.util.Serialisation;
  *
  */
 public class Auth {
-
     /**
      * Authentication methods
      */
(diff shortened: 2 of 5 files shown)


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -827,13 +829,30 @@ public AuthOptions getAuthOptions() {
      * Renew auth credentials.
      * Will obtain a new token, even if we already have an apparently valid one.
      * Authorization will use the parameters supplied on construction.
+     * @deprecated Use {@link Auth#renewAuth} instead
```
QuintinWillison, line 832: I think this entry should be expanded to explain the reasons behind the deprecation, being that this method's behaviour was insufficient for the needs of most developers, given it would return immediately - not providing any mechanism for the caller to be notified when the re-auth had actually completed.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -827,13 +829,30 @@ public AuthOptions getAuthOptions() {
...
+     * Renew auth credentials.
+     * Will obtain a new token, even if we already have an apparently valid one.
+     * Authorization will use the parameters supplied on construction.
+     *
+     * @return
+     * A single entry that contain a token detail and a future that represent an asynchronous result
+     * Clients must wait for the future result to finish before processing. If there is an exception happened during
+     * asynchronous operation the future will contain an AblyException
+     */
+    public Map.Entry<TokenDetails, Future<AblyException>> renewAuth() throws AblyException {
```
QuintinWillison, line 851: This interface definition looks wrong to me, in that there's an [in-built exception handling path](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completeExceptionally-java.lang.Throwable-) provided by `Future`s, therefore it's not idiomatic to return an exception as the result.

It's been a while since I've stared at `Future`-based APIs, or tried to define one myself, but I think the standard return type when you've got nothing to return except for the notion of 'success' is `CompletableFuture<Void>`... [this Stackoverflow answer](https://stackoverflow.com/a/39002692/392847) seems to capture that quite well.

QuintinWillison (changes requested):
I've not gone too deep on this first review pass, but a couple of observations that were immediately obvious to me. 😁

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -975,6 +981,79 @@ public void onAuthUpdated(final String token, final boolean waitForResponse) thr
...
+                    final ErrorInfo reason = waiter.waitForChange();
+                    final ConnectionState connectionState = currentState.state;
+                    switch (connectionState) {
+                        case connected:
+                            Log.v(TAG, "onAuthUpdated: got connected");
+                            waitingForConnected = false;
+                            break;
+
+                        case connecting:
+                        case disconnected:
```
KacperKluka, line 1040: Why for `disconnected` do we simply log and do nothing more? Wouldn't it mean that the SDK can be stuck in this state? :thinking:

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -861,6 +861,76 @@ public Object getTokenRequest(Auth.TokenParams params) {
         }
     }
 
+
+    /**
+     * Call renewAuth() whilst connecting; verify there's no crash (see https://github.com/ably/ably-java/issues/503)
+     */
+    @Test
+    public void auth_renewAuth_whilst_connecting() {
```
KacperKluka, line 869: This test is a bit hard to read and comprehend, could you please refactor it and extract its steps to separate methods? For example the first few lines could possibly be extracted to a method `createTokenDetails()` :wink:

KacperKluka (changes requested):
A few questions and suggestions :smile:

lib/src/main/java/io/ably/lib/rest/AblyBase.java
```diff
@@ -316,6 +317,15 @@ protected void onAuthUpdated(String token, boolean waitForResponse) throws AblyE
...
     }
 
+    /**
+     * Override this method in AblyRealtime and pass updated token to ConnectionManager
+     * @param token new token
+     */
+    protected Future<AblyException> onAuthUpdatedAsync(String token)  {
+        //this must be overriden by subclass
+        return null;
+    }
```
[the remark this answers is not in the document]
ikbalkaya: Fair question but this is due to current architecture of the app and I just created a similar method as the method in above. Because `AblyRest` is a concrete class and `AblyRealtime` is the one that's going to be using it, we will have to create an empty method on `AblyRest` with the same name if this is described an abstract. Apparently this method is not even suitable to be a class method, it would be better suited as an interface method and only the classes that implements interface would have that in, but here we are 🤷‍♂️

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: This code block is just the same code block that was blocking the method to return, so I have not made any change there. So I do not know the answer to your question

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
ikbalkaya: Again, this is exactly the same test that was done on `auth_renew_whilst_connecting` with the difference of new function call

lib/src/main/java/io/ably/lib/rest/Auth.java
ikbalkaya: 89d4acd806e72105801a3f778f5d6e166f782b57

lib/src/main/java/io/ably/lib/rest/Auth.java
ikbalkaya: Fair enough, I did not think it was ideal at all. I have actually initially used a Future<Void> than decided that it would be pragmatic to return an optional exception. I will see if I can use CompletableFuture and provide a more appropriate error path.

ikbalkaya: Unfortunately we cannot yet use [CompletableFuture](https://developer.android.com/reference/java/util/concurrent/CompletableFuture) as it's only available on Android SDK >=24 
However I have changed the API surface a little bit and threw exception directly from `call()` . According to this [SO answer ](https://stackoverflow.com/a/2248166) we are permitted to throw checked exception from call that is going to be propagated as an `ExecutionException` to the caller.  Change : 14359cea0406dddb895873c8d0a3e66ab6fb9357

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -827,13 +829,31 @@ public AuthOptions getAuthOptions() {
...
+    /**
+     * Renew auth credentials.
+     * Will obtain a new token, even if we already have an apparently valid one.
+     * Authorization will use the parameters supplied on construction.
+     *
+     * @return
+     * A single entry that contain a token detail and a future that represent an asynchronous result
+     * Clients must wait for the future result to finish before processing.
+     */
+    public Map.Entry<TokenDetails, Future<Void>> renewAuth() throws AblyException {
```
KacperKluka, line 852: Since this is our public API, do we really want to return this as a `Map.Entry`? Then the users would have to call `renewAuth().getKey()` to get the `TokenDetails` and `renewAuth().getValue()` to get the `Future`, which is not very intuitive IMO :thinking: Maybe let's create a new class that would act as a container for those two values? Something like `RenewAuthResult`? :thinking:

KacperKluka (changes requested):
A few suggestions :wink:

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -861,6 +861,76 @@ public Object getTokenRequest(Auth.TokenParams params) {
...
+                }
+            };
+
+            final AblyRealtime ably = new AblyRealtime(opts);
+            synchronized (opts) {
+                ably.connect();
+                try {
+                    opts.wait();
+                } catch(InterruptedException ie) {}
+                ably.auth.renewAuth();
```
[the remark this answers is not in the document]
ikbalkaya: As we discussed, above particular test does not examine the result. This one also does the same. In the new version I have added a comment for that

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -975,6 +982,79 @@ public void onAuthUpdated(final String token, final boolean waitForResponse) thr
...
+                            Log.v(TAG, "onAuthUpdated: " + connectionState);
+                            break;
+
+                        default:
+                            /* suspended/closed/error: throw the error. */
+                            Log.v(TAG, "onAuthUpdated: throwing exception");
+                            authUpdateResult.onUpdate(false, reason);
+                            waitingForConnected = false;
```
KacperKluka, line 1048: I know that I'm the one who advised to put this instruction here but as I look at it now I see that's not enough :sweat_smile: This will break the `while` loop and jump to the instruction after the loop which is `authUpdateResult.onUpdate(true, null);`, so the `onUpdate()` callback will be called twice, once with an error and once with a success. We have 2 options here:
1. Put a `return` statement here and exit the function (but I'm not sure if the `finally` block would be called in that case)
2. Move the `authUpdateResult.onUpdate(true, null);` from line 1051 to the `case connected:` as we want to notify about the success once we are connected

I'd personally go with 2 :wink:

KacperKluka (changes requested):
Almost there :grin:

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: Great spot. I made a change here : 1bbbd8ba1617fedc5fbd0cb0afb07e68fa676c04
Second one sounds more reasonable to me. But the first one would also work, because `finally` is executed regardless of you return from a function or not

KacperKluka (approved):
LGTM :+1:

Issue #800, opened 2022-06-17
Document which thread is whole SDK or callbacks using


qsdigor:
As we all know there are different threads in java, they usually differentiate into one main thread and many background threads. 

It looks like SDK is using background threads to do network work and returns results on background threads. This is good as it won't block apps and can run async. Network stuff is done on `WebSocketConnectReadThread` thread.

The issue is usually on android apps as they tend to change UI after receiving presence updates or get messages. Our background thread does not support UI updates like changing some texts or updating recycler view adapters.

We have multiple options:
- Change the thread to `main` when returning the results. This will increase complexity and introduce bugs.
- Implement an option to allow users to select a thread of their choice.
- Document usage of background threads and point out that all work is done on background tread, for UI changes please use the main thread. 

I would recommend the third option as a simpler and better option. This won't break any apps or cause implementation issues as the code won't change.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2059) by [Unito](https://www.unito.io)

qsdigor:
@QuintinWillison @KacperKluka I would value your opinion on this. Thank you

KacperKluka:
I think we should at least document it :+1: I had [an issue](https://github.com/ably/ably-java/issues/694#issuecomment-945632291) some time ago and it turned out to be related to background threads and they seem to be poorly documented in the SDK.

QuintinWillison:
In the short term, under this issue, I think we should go with your third option and document better the threading behaviours exhibited.

Longer term, marshalling to the main thread - at least on Android - is probably the simplest option for customers to understand. As you point out, this could cause all kinds of new issues for customer apps. And, of course, we need to work out how this would fit into the brave (new) world of Kotlin coroutines. 🤔

ikbalkaya:
I would prefer callbacks/listeners are invoked in the same thread as the thread that methods are called on. In my experience of using some good SDKs (Retrofit / Google SDks) , async results are provided on the same thread. I also remember using some other SDKs (including Ably) that callbacks are happening on background thread. I think this might be confusing for developers and not a great developer experience for the reasons below:

* Callback/listener methods aren't intended to be blocking, they are just there to notify about a result from an async operation. So I do not think there is a strong reason for background threads. 
* SDK users  might have the wrong assumptions about the behaviour as commonly used SDKs do not use them. 
[...]

PR #818, opened 2022-07-07
Document thread policy for callbacks and add missing documentation for callbacks

qsdigor:
As all of our callbacks are invoked on background threads. There was no documentation about which thread is used in Ably Java SDK and this can be a problem for android developers as UI updating needs to be done on the main thread.

I have updated JavaDoc and README of threading in Ably callbacks and their usage. 
I have also added JavaDoc to ChannelStateListener and ConnectionStateListener which were missing.

Please feel free to suggest improvements. 

Closes https://github.com/ably/ably-java/issues/800

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ Log.setHandler(null);
 #### Threads
 
 AblyRealtime will invoke all callbacks on background thread. 
-If you are using Ably in Android application it is advised to switch to main thread to update UI. 
+If you are using Ably in Android application you must switch to main thread to update UI. 
 
 ```java
 channel.presence.enter("john.doe", new CompletionListener() {
--- a/README.md
+++ b/README.md
@@ -399,7 +399,7 @@ Log.setHandler(null);
 
 #### Threads
 
-AblyRealtime will return all callbacks on background thread. 
+AblyRealtime will invoke all callbacks on background thread. 
 If you are using Ably in Android application it is advised to switch to main thread to update UI. 
 
 ```java
--- a/README.md
+++ b/README.md
@@ -397,6 +397,35 @@ import io.ably.lib.util.Log;
 Log.setHandler(null);
 ```
 
+#### Threads
+
+AblyRealtime will return all callbacks on background thread. 
+If you are using Ably in Android application it is advised to switch to main thread to update UI. 
+
+```java
+channel.presence.enter("john.doe", new CompletionListener() {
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -397,6 +397,35 @@ import io.ably.lib.util.Log;
 Log.setHandler(null);
 ```
 
+#### Threads
+
+AblyRealtime will return all callbacks on background thread.
```
ikbalkaya, line 402:
```suggestion
AblyRealtime will invoke all callbacks on background thread. 
```

lib/src/main/java/io/ably/lib/realtime/CompletionListener.java
```diff
@@ -10,12 +10,16 @@
  */
 public interface CompletionListener {
     /**
-     * Called when the associated operation completes successfully,
+     * Called when the associated operation completes successfully.
+     * <p>
+     * This callback is triggered on background thread.
```
ikbalkaya, line 15: And this

lib/src/main/java/io/ably/lib/realtime/ConnectionStateListener.java
```diff
@@ -2,8 +2,18 @@
...
 
+/**
+ * An interface whereby a client may be notified of state changes for a connection.
+ */
 public interface ConnectionStateListener {
 
+    /**
+     * Called when connection state changes.
+     * <p>
+     * This callback is triggered on background thread.
```
ikbalkaya, line 13: This.

ikbalkaya (changes requested):
Good starting and motivation.

I do not think interface methods are the right place to document thread policy. They should instead be documented on public methods that contain callbacks.  for example `publish` `enter` etc.

lib/src/main/java/io/ably/lib/realtime/ChannelStateListener.java
```diff
@@ -7,6 +7,10 @@
  */
 public interface ChannelStateListener {
 
+    /**
+     * Called when channel state changes.
+     * @param stateChange information about the new state. Check {@link ChannelState ChannelState} - for all states available.
+     */
```
ikbalkaya, line 13: This documentation is slightly out of scope of this PR. Can you update the PR / issue description for this?

ikbalkaya (approved):

PR #822, opened 2022-07-15
Increase minSdkVersion to 21 (lollipop)

qsdigor:
Increasing the minimum API level to 21 will support 98.6% or more devices while keeping the complexity of the old APIs at a minimum.


KacperKluka (changes requested):
I think we have information about supported versions in the README file, we need to update it as well to let people know that now we support Android 21 and above :wink:

KacperKluka (approved):
LGTM :+1:

QuintinWillison (approved):

PR #824, opened 2022-07-15
Rename module lib to core

qsdigor:
Rename module lib to core as it is more meaningful.


KacperKluka (changes requested):
Good job with this PR but even though we are renaming the `lib` module to `core` we should not change its package name :wink: With v2.0.0 we want to make as few changes to the API and the way of using the SDK as possible (more details in [this internal conversation](https://ably-real-time.slack.com/archives/C019LDWM9UH/p1652434930046949)). If we rename the package then users will have to fix most of their imports which is something we don't want to do. :sweat_smile: 

Please rename the `lib` module to `core` but do not change its package `io.ably.lib` :bow:

qsdigor:
@KacperKluka you are correct, a lot of code from lib module is open and it is used openly. I will revert this and push the code again for review. Thank you

README.md
```diff
@@ -364,10 +364,10 @@ You can get log output from the library by modifying the log level:
...
 import io.ably.lib.util.Log;
 
-ClientOptions opts = new ClientOptions(key);
-opts.logLevel = Log.VERBOSE;
-AblyRest ably = new AblyRest(opts);
-...
+ClientOptions opts=new ClientOptions(key);
+  opts.logLevel=Log.VERBOSE;
+  AblyRest ably=new AblyRest(opts);
+  ...
```
KacperKluka, line 370: This seems out of scope of this PR (and the formatting applied is a bit odd :sweat_smile:), please remove it :wink:

README.md
```diff
@@ -377,14 +377,14 @@ You can redirect the log output to a logger of your own by specifying a custom l
...
-AblyRest ably = new AblyRest(opts);
-...
+ClientOptions opts=new ClientOptions(key);
+  opts.logHandler=new LogHandler(){
+public void println(int severity,String tag,String msg,Throwable tr){
+  /* handle log output here ... */
+  }
+  };
+  AblyRest ably=new AblyRest(opts);
+  ...
```
KacperKluka, line 387: This seems out of scope of this PR (and the formatting applied is a bit odd :sweat_smile:), please remove it :wink:

java/build.gradle
```diff
@@ -11,10 +11,9 @@ sourceCompatibility = 1.7
 targetCompatibility = 1.7
 
 dependencies {
-    implementation project(':lib')
-
+    implementation project(':core')
     // Enable access to test classes from the :lib module for integration test cases sharing
-    testImplementation project(path: ':lib', configuration: 'sharedTestClasses')
+    testImplementation project(configuration: 'sharedTestClasses', path: ':core')
```
KacperKluka, line 16: The new line between the dependency and the comment was added there intentionally and should not be removed :wink: 
```
implementation project(':core')

// Enable access to test classes from the :lib module for integration test cases sharing
testImplementation project(configuration: 'sharedTestClasses', path: ':core')
```

KacperKluka (changes requested):
Almost there :wink: Please also update the `UPDATING.md` file as there we mention the `lib` module which is no longer correct :bow:

README.md
qsdigor: done

KacperKluka (changes requested):
Almost there :wink:

android/build.gradle
```diff
@@ -74,11 +74,11 @@ tasks.withType(com.android.build.gradle.internal.tasks.AndroidTestTask) { task -
 }
 
 dependencies {
-    api project(':lib')
+    api project(':core')
     implementation 'com.google.firebase:firebase-messaging:23.0.4'
 
     // Enable access to test classes from the :lib module for integration test cases sharing
-    testImplementation project(path: ':lib', configuration: 'sharedTestClasses')
+    testImplementation project(configuration: 'sharedTestClasses', path: ':core')
```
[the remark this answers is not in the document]
qsdigor: You are correct, I have revert the order in [java/build.gradle](https://github.com/ably/ably-java/pull/824/commits/fa593eaefef2cb630d548b3a7ea619971b5c5eec)

KacperKluka (approved):
LGTM :+1:

QuintinWillison (approved):

PR #834, opened 2022-09-13
Fail Ably connection if auth callback throws specific errors

KacperKluka:
The [RSA4d](https://docs.ably.com/client-lib-development-guide/features/#RSA4d) was not implemented properly as it only applied when you called `ably.auth.authorize()`. Now it applies whenever the token auth is being used.

When an exception is thrown from the token auth callback that has the `statusCode` set to 403 it should fail the connection https://github.com/ably/ably-java/pull/834/commits/aa1415904032391f3fa50aa908ede91bc8fedbf1.
Additionally, after [a discussion](https://github.com/ably/ably-asset-tracking-android/pull/745#issuecomment-1228170544) we've decided that we should not force SDK users to throw an `AblyException`. They should be able to throw any exception and it should be propagated and handled correctly https://github.com/ably/ably-java/pull/834/commits/92c260f3d7c0c8ee503a24b7cca51ad1e88db7d4. 
Furthermore, a new exception interface is added and when an exception implements it, we treat it as if it was an Ably exception with status code 403 https://github.com/ably/ably-java/pull/834/commits/5a23d44937600499d271d4166da0106492b9d869.

According to [RSA4d1](https://docs.ably.com/client-lib-development-guide/features/#RSA4d1) this behaviour could not happen in the `requestToken()` method but luckily the method is only used in the `assertValidToken()` method. Therefore, I've put the logic responsible for failing the connection there.

I wonder if we should also update some docs that would let users know about the new behaviour, WDYT?

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -616,7 +616,7 @@ public class Auth {
                     signedTokenRequest = (TokenRequest)authCallbackResponse;
                 else
                     throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));
-            } catch(AblyException e) {
+            } catch (Exception e) {
                 throw AblyException.fromErrorInfo(e, new ErrorInfo("authCallback failed with an exception", 401, 80019));
             }
         } else if(tokenOptions.authUrl != null) {
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -617,11 +617,11 @@ public class Auth {
                     signedTokenRequest = (TokenRequest)authCallbackResponse;
                 else
                     throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));
-            } catch (Exception e) {
-                boolean isTokenExceptionNonRetriable = e instanceof NonRetriableTokenException;
-                boolean isAblyExceptionNonRetriable = e instanceof AblyException && ((AblyException) e).errorInfo.statusCode == 403;
-                boolean shouldNotRetryAuthOperation = isTokenExceptionNonRetriable || isAblyExceptionNonRetriable;
-                int statusCode = shouldNotRetryAuthOperation ? 403 : 401; // RSA4c & RSA4d
+            } catch (final Exception e) {
+                final boolean isTokenExceptionNonRetriable = e instanceof NonRetriableTokenException;
+                final boolean isAblyExceptionNonRetriable = e instanceof AblyException && ((AblyException) e).errorInfo.statusCode == 403;
+                final boolean shouldNotRetryAuthOperation = isTokenExceptionNonRetriable || isAblyExceptionNonRetriable;
+                final int statusCode = shouldNotRetryAuthOperation ? 403 : 401; // RSA4c & RSA4d
                 throw AblyException.fromErrorInfo(e, new ErrorInfo("authCallback failed with an exception", statusCode, 80019));
             }
         } else if(tokenOptions.authUrl != null) {
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -618,8 +618,10 @@ public class Auth {
                 else
                     throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));
             } catch (Exception e) {
+                boolean isTokenExceptionNonRetriable = e instanceof NonRetriableTokenException;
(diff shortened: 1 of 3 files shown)


lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -616,8 +617,12 @@ public TokenDetails requestToken(TokenParams params, AuthOptions tokenOptions) t
                     signedTokenRequest = (TokenRequest)authCallbackResponse;
                 else
                     throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));
-            } catch(AblyException e) {
-                throw AblyException.fromErrorInfo(e, new ErrorInfo("authCallback failed with an exception", 401, 80019));
+            } catch (Exception e) {
+                boolean isTokenExceptionNonRetriable = e instanceof NonRetriableTokenException;
+                boolean isAblyExceptionNonRetriable = e instanceof AblyException && ((AblyException) e).errorInfo.statusCode == 403;
+                boolean shouldNotRetryAuthOperation = isTokenExceptionNonRetriable || isAblyExceptionNonRetriable;
+                int statusCode = shouldNotRetryAuthOperation ? 403 : 401; // RSA4c & RSA4d
```
qsdigor, line 624: maybe we should name this resolvedStatusCode as we determine which code it should be. 
Also, should we extract these common HTTP error codes to HttpConstants and use them from there rather than have it hard coded?

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -616,8 +617,12 @@ public TokenDetails requestToken(TokenParams params, AuthOptions tokenOptions) t
...
+                throw AblyException.fromErrorInfo(e, new ErrorInfo("authCallback failed with an exception", statusCode, 80019));
```
qsdigor, line 625: We should extract Ably error codes to AblyConstants as well and link them to a specific API doc page.

qsdigor (approved):

ikbalkaya (approved):
LGTM

paddybyers (approved):
LGTM

Should we also have tests that verify the correct behaviour:
- on a token renewal that happens after connection establishment (ie not just when failing during initial connection)? This could include client-initiated renewal via `authorize()`, plus server-initiated renewal on token expiry;
- for REST requests?

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -616,8 +617,12 @@ public TokenDetails requestToken(TokenParams params, AuthOptions tokenOptions) t
                     signedTokenRequest = (TokenRequest)authCallbackResponse;
                 else
                     throw AblyException.fromErrorInfo(new ErrorInfo("Invalid authCallback response", 400, 40000));
-            } catch(AblyException e) {
-                throw AblyException.fromErrorInfo(e, new ErrorInfo("authCallback failed with an exception", 401, 80019));
+            } catch (Exception e) {
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-java/pull/834/commits/8a99bcfc3681c442a0fb2f767345a9d09129cf91

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -616,8 +617,12 @@ public TokenDetails requestToken(TokenParams params, AuthOptions tokenOptions) t
...
+                boolean isTokenExceptionNonRetriable = e instanceof NonRetriableTokenException;
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-java/pull/834/commits/8a99bcfc3681c442a0fb2f767345a9d09129cf91

lib/src/main/java/io/ably/lib/types/NonRetriableTokenException.java
```diff
@@ -0,0 +1,8 @@
+package io.ably.lib.types;
+
+/**
+ * Implement this interface in your exception class if the token auth operation should not be retried.
```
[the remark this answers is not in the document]
KacperKluka: https://github.com/ably/ably-java/pull/834/commits/1b82e11e0ffddf9e6d202d9fae59f193786bf0d1

lib/src/main/java/io/ably/lib/rest/Auth.java
qsdigor: We agreed to move this to a new task which will be done later. https://github.com/ably/ably-java/issues/837

lib/src/main/java/io/ably/lib/rest/Auth.java
qsdigor: The same thing as HTTP constants, we will do this in the near future. https://github.com/ably/ably-java/issues/838

QuintinWillison (approved):

KacperKluka:
I'm going to land this PR as this is an important fix for the customers.

@paddybyers I've created an issue for adding those tests you've mentioned :wink: https://github.com/ably/ably-java/issues/839

PR #835, opened 2022-09-14
Add or update doc comment

qsdigor:
As part of task EDX-278, my work required me to add new docs from [canonical descriptions](https://github.com/ably/canonical-api-reference-prototyping/blob/main/descriptions.md) table.

While all things from the canonical table are implemented, some java objects or methods will have the same docs as before.


lib/src/main/java/io/ably/lib/types/ConnectionDetails.java
```diff
@@ -8,15 +8,70 @@
...
     public Long maxMessageSize;
+    /**
+     * The maximum allowable number of requests per second from a client or Ably.
+     * In the case of a realtime connection, this restriction applies to the number of messages sent,
+     * whereas in the case of REST, it is the total number of REST requests per second.
+     * <p>
+     * Spec: CD2e
+     */
     public Long maxInboundRate;
     public Long maxOutboundRate;
```
ikbalkaya, line 54: This is not documented

lib/src/main/java/io/ably/lib/types/Message.java
```diff
@@ -151,11 +166,16 @@ Message readMsgpack(MessageUnpacker unpacker) throws IOException {
...
+        /**
+         * An array of channel names to publish messages to.
+         */
         public String[] channels;
+        /**
+         * An array of {@link Message} objects to publish.
+         */
         public Message[] messages;
 
         public Batch(String channel, Message[] messages) {
```
ikbalkaya, line 181: No documentation for this constructor

lib/src/main/java/io/ably/lib/types/Stats.java
```diff
@@ -6,32 +6,51 @@
...
+         */
         public double count;
+        /**
+         * The total number of bytes transferred for all messages.
+         * <p>
+         * Spec: TS5b
+         */
         public double data;
+
         public double uncompressedData;
```
ikbalkaya, line 54: This is not documented

lib/src/main/java/io/ably/lib/types/Stats.java
```diff
@@ -121,18 +251,79 @@ public static long fromIntervalId(String intervalId) {
...
+     */
     public String intervalId;
+    /**
+     * The length of the interval the stats span. Values will be a {@link Granularity}.
+     * <p>
+     * Spec: TS12c
+     */
     public String unit;
+
     public int count;
```
ikbalkaya, line 269: This is not documented

lib/src/main/java/io/ably/lib/util/Crypto.java
```diff
@@ -44,17 +30,20 @@ public class Crypto {
...
+         * <p>
+         * Spec: TZ2a
+         */
         private final String algorithm;
+        /**
+         * The length of the key in bits; for example 128 or 256.
+         * <p>
+         * Spec: TZ2b
+         */
         private final int keyLength;
```
ikbalkaya, line 47: Is there a reason why only this and above are documented?

ikbalkaya (changes requested):
Caution : I have only scanned through changes to see any patterns and my review isn't about content of comments. Some observations
* Some params has been documented as @param X but no additional documentation is provided
* There are @throws XExceptions but generally no explanation given on in which cases those exceptions will be thrown

lib/src/main/java/io/ably/lib/util/Crypto.java
qsdigor: I have only documented docs that exist in the [canonical descriptions](https://github.com/ably/canonical-api-reference-prototyping/blob/main/descriptions.md) table. There is a lot of thing missing which I documented in a separate file. Some of them were added to canonical but a lot of them did not as they were flagged as java specific so we would need to come up with a doc.

ikbalkaya (approved):
Approving as Igor has said that there is a document for absent documentation.

QuintinWillison:
This PR has code issues, [for example](https://github.com/ably/ably-java/actions/runs/3051477364/jobs/4919754916#step:3:57):

> Error: eckstyle] [ERROR] /home/runner/work/ably-java/ably-java/lib/src/main/java/io/ably/lib/rest/ChannelBase.java:8:8: Unused import - io.ably.lib.push.Push. [UnusedImports]

So, I'm going to move it back to _Draft_ stage while @qsdigor resolves these.

PR #842, opened 2022-09-26
Protocol 2.0

qsdigor:


core/src/main/java/io/ably/lib/transport/ITransport.java
```diff
@@ -73,13 +72,11 @@ public Param[] getConnectParams(Param[] baseParams) {
                 paramList.add(new Param("resume", connectionKey));
                 if(connectionSerial != null)
```
andydunstall, line 73: Can `connectionSerial` be removed now?

core/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -872,7 +881,7 @@ synchronized Collection<PresenceMessage> get(Param[] params) throws AblyExceptio
          * false if the message is already superseded
          */
         synchronized boolean put(PresenceMessage item) {
-            String key = item.memberKey();
+            String key = item.clientId; //RTP17h
```
andydunstall, line 884: RTP17h says only the presence map containing members that match the current connectionId should be keyed by client ID - though this looks like both maps are being keyed by client ID? (both `PresenceMap presence` and `PresenceMap internalPresence`?)

andydunstall:
Just to check - looking at the presence tests in `RealtimePresenceTest.java` - the majority of the tests have an `@Ignore` annotation? Does this mean they could be failing?

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
```diff
@@ -92,13 +92,21 @@ private void setState(ChannelState newState, ErrorInfo reason, boolean resumed)
...
             stateChange = new ChannelStateListener.ChannelStateChange(newState, this.state, reason, resumed);
             this.state = stateChange.current;
             this.reason = stateChange.reason;
         }
 
-        if(notifyStateChange) {
+        if (newState == ChannelState.detached || newState == ChannelState.suspended || newState == ChannelState.failed)
+            properties.channelSerial = null; //RTP5a1
+
+        if (newState == ChannelState.attached && state == ChannelState.attached) {
```
andydunstall, line 104: I'm a bit confused what the difference between `newState` and `channel.state` is here? Are these expected to be the same?

core/src/main/java/io/ably/lib/transport/ITransport.java
qsdigor: > Just to check - looking at the presence tests in `RealtimePresenceTest.java` - the majority of the tests have an `@Ignore` annotation? Does this mean they could be failing?
> 
> edit: the recovery tests also have `@Ignore` annotations?

I have raised an issue about the ignored tests in the java channel. We will need to see what to do with those.

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: newState is received from the server and channel.state is the current state of the channel which is old in this context.

andydunstall: ok so is this only doing a re-enter if it receives `ATTACHED` if its already in the `ATTACHED` state? (which seems to contradict RTP17f?)

core/src/main/java/io/ably/lib/transport/ITransport.java
andydunstall: Ok - though its not clear how the presence changes are being tested?

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: I think I tried to meet expectation of the case in braces `(It does not need to do so for @RTL12@ @ATTACHED@ events received on already-@ATTACHED@ channels).`
As I can see newState and state will be the same as newState is assigned to state in code line 97 so latter check can be removed (state == ChannelState.attached) or moved before state is changed.

core/src/main/java/io/ably/lib/realtime/Presence.java
qsdigor: It seems spec are changed as I cannot find RTP17h in https://github.com/ably/docs/pull/1526/files

andydunstall: Its moved to https://github.com/ably/specification/pull/88/files

qsdigor: It looks like there is only one presence map in java as I cannot find another one.
Member key seems deprecated by RTP17h from this [internal slack message](https://ably-real-time.slack.com/archives/C02P7F81DPS/p1664904124309399) or I did not get this correctly?

andydunstall: I mean there are maps `presence` and `internalPresence` right?
```
    private final PresenceMap presence = new PresenceMap();
    private final PresenceMap internalPresence = new PresenceMap();
```

My understanding from the spec was:
* The `presence` map is the main mapping from `memberKey`s to presence message - from spec RTP2
* The `internalPresence` only contains the members that match the current connectionId - mapping `clientId` to presence message - from spec RTP17h

So `PresenceMap` needs to be indexed differently depending on its type? eg from ably-js https://github.com/ably/ably-js/pull/1045/commits/a036b0d5ff892ab2a577afd73ce4b1ca260f75c9

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1083,53 +1084,40 @@ public void onMessage(ITransport transport, ProtocolMessage message) throws Ably
...
-            /* The connection id has changed. Reset the message serial and the
-             * pending message queue (which fails the messages currently in
-             * there). */
-            pendingMessages.reset(msgSerial,
-                    new ErrorInfo("Connection resume failed", 500, 50000));
+        if (message.action == ProtocolMessage.Action.connected && message.connectionId.equals(connection.id) && error == null) {
+            //RTN15c6
+            Log.d(TAG, "connection has reconnected and resumed successfully");
+            connection.reason = null;
+            channels.reAttach();
```
andydunstall, line 1097: My understanding is we will always need to reattach after we get a new connection - what do you think about moving the `channels.reAttach()` after the `if/else` to make that explicit?

(I think it what you've will still work given the spec - though just to be safe incase we get an unexpected response from the backend maybe worth making sure `reAttach` will always run?)

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1083,53 +1084,40 @@ public void onMessage(ITransport transport, ProtocolMessage message) throws Ably
...
+
+            //RTN19a
+            if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
```
andydunstall, line 1100: Should the messages be processed in both branches given in both cases we have a connection? `RTN15c7` also says:
> The connection should also process any messages queued per @RTL6c2@.

In which case should this block be moved outside the `if/else` too?

qsdigor: So we are good here?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: As I see now this function does not need `message.action == ProtocolMessage.Action.connected` as this is already filtered to connected events. 
I will refactor this to not check for connected state but I think we need to keep reAttach() in if/else cases as they are specific about connectionId and error received.

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: I have added check to better specify `RTP17f` which will reEnter() only if new state is attached and old is not attached as described in braces text.

Old state is initiated with value of current state, before it is resigned with new state.
```
if (newState == ChannelState.attached && oldState != ChannelState.attached) {
    presence.reEnter(newState);  //RTP17f
}
```

qsdigor: Actually, I do not need to check states as I already check in presence.reEnter method. 
However, I needed to change case where channel state is not attached.
```
if (newState == ChannelState.attached && channel.state != ChannelState.attached) {
    internalPresence.reEnter();
}
```

qsdigor: @andydunstall can you confirm my findings, please?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
andydunstall: If `State.enact` already implements this then shouldn't this block (sending `pendingMessages`) be removed?

Though I'm not really familiar with the sdk so may be misunderstanding - if its not already implemented this should be sending `pendingMessages` in both branches

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
andydunstall: Though every time we get a new connection we must `reAttach` right? Regardless of whether we get a new connection ID and/or an error? Maybe it doesn't matter - though to me the safest option seems to be move out of the `if/else` so we can be sure we already reattach on a new connection?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: I moved this spec to the outside of if/else but I am not sure if I should remove a message from the pending queue once it is sent as this implementation for sending is not clear.

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: I am not sure if we should reattach every connect case as spec RTN15c6 and RTN15c7 state when we should do it.

`RTN15c6 - A @CONNECTED@ @ProtocolMessage@ with the same @connectionId@ as the current client (and no @error@ property).`
`RTN15c7 - CONNECTED@ @ProtocolMessage@ with a new @connectionId@ and an @ErrorInfo@ in the @error@ field.`

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: @andydunstall This was my last decision on the matter https://github.com/ably/ably-java/pull/842/commits/0287b17b8e72fc88332dac8d5b7cb33b6e3ffb25

core/src/main/java/io/ably/lib/realtime/Presence.java
qsdigor: @andydunstall I have refactored it to match JS SDK, commit is here https://github.com/ably/ably-java/pull/842/commits/86625b85409b7a78960b9680e299733e09dec90b. Thank you for your input.

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
SimonWoolf: @qsdigor I'm not sure what you have now makes any sense:
- You're only doing a re-enter if `newState is attached && channel.state != attached`. But right above, [on line 97](https://github.com/ably/ably-java/pull/842/files#diff-6b3bce3cee6e98aa7a5f584556ac3848bd9a25c3ecc1955f171648d3978c11b6R97), the method is setting `this.state = stateChange.current`. So at that point, won't the channel state always equal the newstate because you've just set it to be?
- Having logic in setState to avoid the RTL12 'attached event when already attached' case will never fire anyway, because  [`setAttached` doesn't call setState if it's already attached](https://github.com/ably/ably-java/blob/b4eaadd3bc613535f3862a78850e5ae8f7844c30/core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java#L315-L324) (because it doesn't need to, there's no state change)

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
SimonWoolf: I agree with Andy here, IMO it reads weirdly here that you have the same channels.reattach() line in both branches of a conditional that's intended to be exhaustive

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,34 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
...
+     * Send all pending messages which are queue.
+     * Remove them from the queue once they are send successfully
+     * Spec: RTN19a
+     */
+    public void sendPendingQueueMessages() {
+        //RTN19a
+        if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
+            for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+                try {
+                    send(queuedMessage.msg, false, new CompletionListener() {
```
SimonWoolf, line 1442: Just calling send() is a breach of RTN19a2, isn't it? The send is going to assign it a new msgSerial, so it won't be idempotent

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,34 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
...
+                        @Override
+                        public void onSuccess() {
+                            pendingMessages.queue.remove(queuedMessage);
```
SimonWoolf, line 1445: I don't understand what you're doing here here. Why do we have a completion listener here? send() will add the message into the pending message list, so it'll now have both instances in there? `pendingMessages#ack()` and `nack()` will handle acks and nacks, so now they'll be handled twice?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
SimonWoolf: (though I've just realised we do want it to assign a new msgSerial if the resume failed, bleh)

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: @SimonWoolf I am not sure if that is the case but there is a way where I can make sure to compare real channel states and makes it more readable. Here is the change https://github.com/ably/ably-java/pull/842/commits/e91d7704f919fbebe2a371bea229ffbc93c796e0

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: I thought we should remove a message from the pending list once it is successfully sent. This way we won't have that same message in a queue state as I do not see anywhere that items are removed from the queue.

core/src/main/java/io/ably/lib/transport/ITransport.java
qsdigor: This part is removed as it is not used at all.

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
SimonWoolf: (Discussed in standup)

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: Yours suggestion is applied here https://github.com/ably/ably-java/pull/842/commits/a6eeb001531505e02bd5149c2c7864d951d3c203

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: Here is the solution we agreed on the meeting https://github.com/ably/ably-java/pull/842/commits/5fae005bdb9ace00c18099a147e315be582ab00c

SimonWoolf (approved):

core/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -793,6 +801,14 @@ void setSuspended(ErrorInfo reason) {
      */
     private class PresenceMap {
 
+        PresenceMap() {
+            this.keyByClientID = false;
+        }
+
+        PresenceMap(boolean keyByClientID) {
```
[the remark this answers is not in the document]
sacOO7: Check https://github.com/ably/ably-dotnet/blob/48ffe9f5e628b57dbdbc01f38f5184eae6a647a1/src/IO.Ably.Shared/Realtime/PresenceMap.cs#L233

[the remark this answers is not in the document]
qsdigor: @sacOO7 Agreed, I was told to refer to JS SDK implementation https://github.com/ably/ably-js/commit/a036b0d5ff892ab2a577afd73ce4b1ca260f75c9
Refactored here https://github.com/ably/ably-java/pull/842/commits/a6fd26db29b23e8b0900304210a8719451c0186c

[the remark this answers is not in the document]
sacOO7: Thanks

core/src/main/java/io/ably/lib/realtime/AblyRealtimeBase.java
```diff
@@ -183,6 +208,21 @@ public void suspendAll(ErrorInfo error, boolean notifyStateChange) {
...
         }
 
+        /**
+         * By spec RTN15c6, RTN15c7
+         */
+        @Override
+        public void reAttach() {
+            for (Map.Entry<String, RealtimeChannelBase> entry : map.entrySet()) {
+                RealtimeChannelBase channel = entry.getValue();
+                if (channel.state == ChannelState.attaching || channel.state == ChannelState.attached || channel.state == ChannelState.suspended) {
```
ikbalkaya, line 218: Can you check if there is a method like `transition` exist in ably-java? I had done the same thing in ably-cocoa and Marat pointed out that this should be happening on its own transition method 🤔

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.ably.lib.util.Serialisation;
+
+public class ConnectionRecoveryKey {
+
+    public String connectionKey;
+    public long msgSerial;
```
ikbalkaya, line 13: Can you make those private and create getters for them if needed?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,24 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
         }
     }
 
+    /**
+     * Send all pending messages which are queue.
+     * Remove them from the queue once they are send successfully
+     * Spec: RTN19a
+     */
+    public void sendPendingQueueMessages() {
```
ikbalkaya, line 1437: Why is this a public method?

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,24 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
...
+        //RTN19a
+        if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
+            for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+                try {
+                    send(queuedMessage.msg, false, null);
```
ikbalkaya, line 1442: Didn't this pending message sending exist before this change?

ikbalkaya (changes requested):
I have not done in depth review but generally looks good. I have some questions and recommendations

android/src/main/java/io/ably/lib/push/LocalDevice.java
```diff
@@ -15,9 +15,10 @@
 import io.ably.lib.types.RegistrationToken;
 import io.ably.lib.util.Base64Coder;
 import io.ably.lib.util.Log;
+import io.ably.lib.util.Serialisation;
 
 public class LocalDevice extends DeviceDetails {
-    public String deviceSecret;
```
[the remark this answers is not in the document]
qsdigor: This field is moved to DeviceDetails as protocol 2.0 requires it now.

core/src/main/java/io/ably/lib/realtime/AblyRealtimeBase.java
```diff
@@ -183,6 +208,21 @@ public void suspendAll(ErrorInfo error, boolean notifyStateChange) {
             }
         }
 
+        /**
+         * By spec RTN15c6, RTN15c7
+         */
+        @Override
+        public void reAttach() {
+            for (Map.Entry<String, RealtimeChannelBase> entry : map.entrySet()) {
```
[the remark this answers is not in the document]
qsdigor: As map is internal object in class InternalMap I will not change its name as it is out of the scope of the task but I will change the name of the entry to channelEntry.

core/src/main/java/io/ably/lib/realtime/AblyRealtimeBase.java
qsdigor: There is no such method, only validateTransition but it is abstract method that does all kinds of stuff across SDK. I cannot find anything similar to transition in java SDK.

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: I will make this private.

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: `send()` is used in many ways but never before to iterate through all pending messages and send them.

core/src/main/java/io/ably/lib/realtime/AblyRealtimeBase.java
```diff
@@ -72,6 +72,31 @@ public void onConnectionStateChanged(ConnectionStateListener.ConnectionStateChan
...
 
+        if (options.recover != null) {
+            ConnectionRecoveryKey recoveryKey = ConnectionRecoveryKey.fromJson(options.recover);
+            if (recoveryKey == null) {
+                Log.d(TAG, "Recovery key initialization failed!");
+                connection.connectionManager.msgSerial = 0; //RTN16f
+            } else {
+                connection.connectionManager.msgSerial = recoveryKey.msgSerial; //RTN16f
+
+                for (Map.Entry<String, String> entry : recoveryKey.serials.entrySet()) {
```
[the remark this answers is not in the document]
qsdigor: Done. You can check it out here https://github.com/ably/ably-java/pull/842/commits/28f438ae58be9b214956eb426c027004ecff79ac

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
ikbalkaya: Yes, but the new protocol doesn't change the IDL for this. I recommend you use `LocalDevice` instead of `DeviceDetails` in transport. That's what I did in ably-cocoa

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
ikbalkaya: I don't think we should repeat the patterns before and we should try to use the best practices as much as we can in new classes. As this is still a Java library I think we should follow best practices for it. So object state must always be exposed through a method - and they must be immutable if possible. 
I don't understand what you mean by code clutter. Exposing object states as public goes against the principles of encapsulation

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
qsdigor: LocalDevice is only part of the Android build so it is not present in core tests. That is why we cannot just use LocalDevice. One solution was to make LocalDeviceBase class in core and override it in Android and Java builds (java not needed) or to just move one field in parent class and implement it in LocalDevice which is android only.

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
qsdigor: As this is a simple data object it is not necessary to preserve encapsulation. 
Best practices change over time, kotlin best practices differ from java ones and even though we are not in kotlin yet we will move towards it which will make all those changes obsolete.
Generating all getters, setters, and map manipulation functions will make clutter in this class.
I can implement your suggested changes if you strongly agree with them.

ikbalkaya: > As this is a simple data object it is not necessary to preserve encapsulation.

Can this object be accessed from outside and state be manipulated by multiple callers and/or threads? What's the difference between this being a data object or another object from the callers' perspective? You still expose state that needs to be encapsulated and can be modified by different actors putting it in an inconsistent state. I don't think we can make an exception for this.

> Best practices change over time, kotlin best practices differ from java ones and even though we are not in kotlin yet we will move towards it which will make all those changes obsolete.

I do not think Kotlin best practices differ that much in terms of object encapsulation and immutability. In fact Kotlin best practices recommends more immutability and state protection. `val` is exactly for this. The shorthand accessing object states is not equivalent to defining public mutable states like in here.

>  Generating all getters, setters, and map manipulation functions will make clutter in this class.

Yes, but you cannot really exchange code safety with uncluttered code. Plus, do you really need setters  here? Could you have defined one or more of these as `final` to make state immutable and only assign state when creating the object?

> I can implement your suggested changes if you strongly agree with them.

As above, yes please

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
ikbalkaya: I personally do not like that a `LocalDevice` that extends a class called `DeviceDetails`, it would be preferrable to have an interface that represents LocalDevice but I suppose we have to follow specs. 

> LocalDevice is only part of the Android build so it is not present in core tests

I wonder what sort of tests exist to test DeviceDetails 🤔 Isn't it supposed to be used for push notifications only.

>  That is why we cannot just use LocalDevice. One solution was to make LocalDeviceBase class in core and override it in Android and Java builds

Instead of doing this, why don't you move `LocalDevice` to core library, provided that it has relevant tests existing in core class, it should be part of it too.

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
qsdigor: I do not think this can be accessed outside and manipulated by different threads at the same time as this is used on initializing ably realtime and when we want to save external recovery key and that is a getter.
I have implemented suggested changes here https://github.com/ably/ably-java/pull/842/commits/18415fc1b0351392c71e6b41f3d23d055dc3ff63

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
qsdigor: Not sure if it is ok to move it to core. @KacperKluka will know better what to do here.

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
ikbalkaya: > I do not think this can be accessed outside and manipulated by different threads 

Of course it can. That's what `public` keyword [is for](https://en.wikibooks.org/wiki/Java_Programming/Keywords/public). 

> as this is used on initializing ably realtime and when we want to save external recovery key and that is a getter.

That's the design intention behind it. The way it was designed make the use of this class much more beyond this use case.

ikbalkaya: Thanks looks much better now!

qsdigor: I do not think end user can use our class in the way it would compromise SDK. End user can use it to create new instance of it but no way to pass it in our sdk.
Any way, it is good this way as well so I will close this conversation.

core/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -1031,13 +1038,45 @@ synchronized void clear() {
...
+         * Spec: RTP17g
+         */
+        synchronized void reEnter() {
+            for (Map.Entry<String, PresenceMessage> entry: members.entrySet()) {
+                PresenceMessage member = entry.getValue();
+                member.action = PresenceMessage.Action.enter;
+                try {
+                    updatePresence(member, null);
+                } catch (AblyException e) {
+                    e.printStackTrace();
```
sacOO7, line 1051: Please use proper logging for this, instead of `e.printStackTrace()`. Currently, `java-sdk` is being used on different platforms, so it won't really work there as expected. e.g. android

sacOO7: This is true for all `e.printStackTrace();` statements.

sacOO7: You can check with @SimonWoolf on this, how to log errors. I think most likely errors are supposed to be passed back to the user who is calling the code so they can be handled gracefully, otherwise logged with proper logging level.

SimonWoolf: RTP17e: `If the publish attempt fails for an automatic presence @ENTER@ (for example, by Ably rejecting it with a @NACK@), an @UPDATE@ event should be emitted on the channel with @resumed@ set to true and @reason@ set to an @ErrorInfo@ object with @code@ @91004@, a @message@ indicating that an automatic re-enter has failed and indicating the @clientId@, and @cause@ set to the reason for the enter failure. The error should also be logged at @warn@ level or higher.`

As usual you don't need to reinvent the wheel here, look at what the library already does in equivalent situations. Look a little further up this same file to endSyncAndEmitLeaves():
```
updatePresence(itemToSend, new CompletionListener() {
    @Override
    public void onSuccess() {
    }

    @Override
    public void onError(ErrorInfo reason) {
            /*
             * (RTP5c3)  If any of the automatic ENTER presence messages published
             * in RTP5c2 fail, then an UPDATE event should be emitted on the channel
             * with resumed set to true and reason set to an ErrorInfo object with error
             * code value 91004 and the error message string containing the message
             * received from Ably (if applicable), the code received from Ably
             * (if applicable) and the explicit or implicit client_id of the PresenceMessage
             */
        String errorString = String.format(Locale.ROOT, "Cannot automatically re-enter %s on channel %s (%s)",
                clientId, channel.name, reason.message);
        Log.e(TAG, errorString);
        channel.emitUpdate(new ErrorInfo(errorString, 91004), true);
    }
});
```

SimonWoolf: Speaking of endSyncAndEmitLeaves, it... is still there, still implements the old auto-re-entry logic (RTP17c)? The conditions on when to auto-re-enter changed, but instead of changing the existing implementation you've just added the new one right next to the old, so now it's doing both?

SimonWoolf: > there is not spec code anywhere. 

Yes, there is. In the same PR as the rest of the spec changes: https://github.com/ably/specification/pull/88/files#diff-d0e5a641ba4816766d38d6c0e7139df7502bed462dc676227057b4509cafad18R705 .

As you can see, `RTP17d` ("Automatic re-entry consists of, for each member of the RTP17 internal PresenceMap whose memberKey is not also a member of the normal PresenceMap...") was removed, and replaced with RTP17g, which says that reentry is unconditionally.

When this sort of requirement changes, you do need to use some common sense. Yes, it's _possible_ the library never implemented the old spec, but chances are it did, so look around and find the relevant implemention of the old spec, which you can modify to instead conform to the new spec.

> As I am still not sure what this thing does

You can see what it does by reading the code. It's not obfuscated. It implements the previous auto-re-entry spec.

qsdigor: Ok I will take a look at it. When I meant no spec code anywhere I meant in the SDK code so finding things is harder. It will be changed soon as most of the code was documented with spec code.

sacOO7: Thanks, I will take a look !

sacOO7: Added more comments

core/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -729,8 +731,19 @@ public void onChannelStateChanged(ChannelStateChange stateChange) {
...
+            //wait for callbacks and lists to populate
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+            }
+            assertEquals("Attached channel histories do not match", expectedAttachedChannelHistory, attachedChannelHistory);
+            assertEquals("Suspended channel histories do not match", expectedSuspendedChannelHistory, suspendedChannelHistory);
+        } catch (AblyException e) {
+            e.printStackTrace();
+            fail("channels_are_reattached_after_reconnecting_when_statettl_plus_idleinterval_has_passed: Unexpected exception");
```
sacOO7, line 743: Can you fail the test by passing the error message inside it ?
There must be some assertion that takes accepts exception as a parameter

core/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -1031,13 +1038,47 @@ synchronized void clear() {
...
+        synchronized void reEnter() {
+            for (Map.Entry<String, PresenceMessage> entry: members.entrySet()) {
+                PresenceMessage member = entry.getValue();
+                member.action = PresenceMessage.Action.enter;
+                try {
+                    updatePresence(member, null);
+                } catch (AblyException e) {
+                    String errorString = String.format(Locale.ROOT, "Cannot automatically re-enter %s on channel %s (%s)",
+                        member.clientId, channel.name, e.errorInfo.message);
+                    Log.e(TAG, errorString);
```
sacOO7, line 1053: As per example provided here ->
 https://github.com/ably/ably-java/pull/842#discussion_r1014099918
We might wanna emit error as a update
`channel.emitUpdate(new ErrorInfo(errorString, 91004), true);`
Just check if right errorCode used to emit the error.
I think all error codes are mentioned here ->
https://github.com/ably/ably-common/blob/main/protocol/errors.json
if error is not channel specific, rather connection specific, you can find there must be a way to emit error on connection
`connection.emitUpdate(new ErrorInfo(errorString, 91004), true);`

core/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
sacOO7: The thing is we should have `error` messages visible on the local/CI once test fails, so it's easy to debug failing test.

core/src/main/java/io/ably/lib/realtime/Presence.java
qsdigor: @sacOO7 I think this is it https://github.com/ably/ably-java/pull/842/commits/d022c3d5013d993f5c0bd49a0ec7a2c53e38fc46

core/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
qsdigor: I think this should do it https://github.com/ably/ably-java/pull/842/commits/d7ed41a600e1c15ce435c7ceb947c2aa4efaf7f0

core/src/main/java/io/ably/lib/realtime/Presence.java
qsdigor: @sacOO7 I think this is it https://github.com/ably/ably-java/commit/d022c3d5013d993f5c0bd49a0ec7a2c53e38fc46

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,26 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
...
+     * Spec: RTN19a
+     */
+    private void sendPendingQueueMessages() {
+        //RTN19a
+        if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
+            for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+                try {
+                    send(queuedMessage.msg, false, null);
+                } catch (AblyException e) {
+                    String errorString = String.format(Locale.ROOT, "Unable to send pending message %s (%s)",
```
sacOO7, line 1444: You can also emit errorString on `connection` as mentioned here => 
https://github.com/ably/ably-java/pull/842/files#r1014250157

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
ikbalkaya: > I do not think end user can use our class in the way it would compromise SDK. End user can use it to create new instance of it but no way to pass it in our sdk.

This is not only a matter of exposing things to the end user. Exposing it to the SDK is always a risk and design flaw

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: Can you just check ... I saw it somewhere sometimes bacl, it should be possible

core/src/main/java/io/ably/lib/realtime/Presence.java
qsdigor: @SimonWoolf I have removed old implementation of re-enter as I already implemented the new logic in separate function reEnter() by spec RTP17g. 
You can check it here https://github.com/ably/ably-java/pull/842/commits/f0e09fb7d59c3581840db1f6fe891b316a33bfda

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: There is no way to emitUpdate() the same way as I did in the Presence.reEnter().
I am not sure if we should as every send event ends up being just logged.

sacOO7: Please check this one -> https://github.com/ably/ably-java/blob/2775178075de882d6184b3ad46a8fa2707ec48f5/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1086

qsdigor: @sacOO7 is https://github.com/ably/ably-java/pull/842/commits/865bca81e42e4f8665abcb847e21349d3f07fa42 ok?

sacOO7: Looks okay. Just make sure, it's being sent for other places too

core/src/main/java/io/ably/lib/realtime/Presence.java
sacOO7: Looks good

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
KacperKluka: According to the [spec about LocalDevice](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8):
> In platforms that **support receiving push notifications**, the `device` method on the `RestClient` or `RealtimeClient` interfaces returns an instance of `LocalDevice` that represents the current state of the device in respect of it being a target for push notifications.

Due to this, I'm not sure if the `LocalDevice` should be present in both `ably-java` and `ably-android` 🤔 The other problem is that the `LocalDevice` uses Android-specific classes (such as `ActivationContext`) which should not be moved to the core library, as they are Android only 😉

core/src/main/java/io/ably/lib/realtime/ConnectionRecoveryKey.java
```diff
@@ -0,0 +1,62 @@
...
+    public Map<String, String> getSerials() {
+        return serials;
+    }
+
+    public void setSerials(Map<String, String> serials) {
+        this.serials.clear();
+        this.serials.putAll(serials);
+    }
+
+    public void addSerials(String channelName, String channelSerial) {
```
KacperKluka, line 45: Since you are only adding a single serial at a time, shouldn`t this method be called `addSerial`? 🤔 
```suggestion
    public void addSerial(String channelName, String channelSerial) {
```

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
```diff
@@ -137,7 +141,7 @@ public void attach(CompletionListener listener) throws  AblyException {
         this.attach(false, listener);
     }
 
-    private void attach(boolean forceReattach, CompletionListener listener) {
+    void attach(boolean forceReattach, CompletionListener listener) {
```
KacperKluka, line 144: Why is this method public now? If it's a part of the API then we should add a proper javadoc to it (just like the other `attach` methods have)

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1447,6 +1429,27 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
         }
     }
 
+    /**
+     * Send all pending messages which are queue.
+     * Remove them from the queue once they are send successfully
```
KacperKluka, line 1434: A small improvement to the docs
```suggestion
     * Send all pending messages which are in the queue.
     * Remove them from the queue once they are sent successfully
```

KacperKluka (changes requested):
A few suggestions 😉

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
qsdigor: @KacperKluka I have opened the issue about deviceSecret and you can read more here https://github.com/ably/ably-java/issues/845

core/src/main/java/io/ably/lib/realtime/RealtimeChannelBase.java
qsdigor: This method is needed internally in AblyRealtimeBase.reAttach(). We removed the visibility modifier so it is only visible in our package.

KacperKluka: True, my bad, this method is not public but visible only in the same package 😅

KacperKluka (commented):
LGTM 👍 But I don't have enough knowledge about the protocol 2.0 implementation details so this is just a "Comment" instead of an "Approve" 😉

KacperKluka (commented):
LGTM 👍

KacperKluka (approved):
Even though [I said](https://github.com/ably/ably-java/pull/842#pullrequestreview-1171630531) that I don't feel confident in leaving an approval I have to do it because Github ignores my "Comment" and shows my review as "Request changes" 😅 Therefore, in order to not block this PR I'm leaving an approval 👍

android/src/main/java/io/ably/lib/push/LocalDevice.java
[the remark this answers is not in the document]
ikbalkaya: Ok. I also created an issue on the documentation repo. https://github.com/ably/specification/issues/110. I'll resolve this conversation in order to unblock this PR.

ikbalkaya (approved):

core/src/main/java/io/ably/lib/realtime/Presence.java
```diff
@@ -1031,13 +982,48 @@ synchronized void clear() {
...
 
+        /**
+         * Spec: RTP17g
+         */
+        synchronized void reEnter() {
+            for (Map.Entry<String, PresenceMessage> entry: members.entrySet()) {
+                PresenceMessage member = entry.getValue();
+                member.action = PresenceMessage.Action.enter;
+                try {
+                    updatePresence(member, null);
```
SimonWoolf, line 993: 1. Could you elaborate on why you replaced
```
PresenceMessage itemToSend = new PresenceMessage();
itemToSend.clientId = item.clientId;
itemToSend.data = item.data;
itemToSend.action = PresenceMessage.Action.enter;
```
with
```
PresenceMessage member = entry.getValue();
member.action = PresenceMessage.Action.enter;
```
It's not obvious that that change is correct (won't it have a possibly-stale connectionId?)

2. Why did you remove the completionListener that was previously being passed to updatePresence? Is there some magic that will turn an async failure into an exception if a completionListener isn't passed that I'm missing?

qsdigor: I guess I misinterpreted RTP17g.
Is this a good implementation https://github.com/ably/ably-java/pull/842/commits/28531da66223c2beed1753ea7df9e1729eb87d4e?

sacOO7 (approved):

QuintinWillison:
I'm moving this pull request back to Draft state because it's currently based on top of the wrong branch (`integration/version-2`) due to the fact that we are very likely to need the protocol version `2` change to be rolled out sooner than our Java/Android SDK version `2` work can be completed, therefore we need this pull request to target the `main` (default) branch of this repository so that it can be released in a Java/Android SDK version `1.2` patch (per [guidance](https://github.com/ably/engineering/blob/main/sdk/versioning.md#semantic-versioning)).

Additionally we can't upgrade this SDK to utilise protocol version `2` until https://github.com/ably/ably-java/issues/859 has also been completed (because this SDK has been recently reverted to protocol version `1.0` rather than `1.2` because of a regression).

SimonWoolf:
> Additionally we can't upgrade this SDK to utilise protocol version 2 until https://github.com/ably/ably-java/issues/859 has also been completed

Given that protocol v2 completely changes the presence-auto-re-enter behaviour anyway, and also that we're going to be releasing protocol v2 as a patch to the current version, I'd argue that it doesn't make sense to do https://github.com/ably/ably-java/issues/859 first, we should just focus on protocol v2

core/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1525,6 +1507,27 @@ public QueuedMessage(ProtocolMessage msg, CompletionListener listener) {
...
+     * Send all pending messages which are in the queue.
+     * Remove them from the queue once they are sent successfully
+     * Spec: RTN19a
+     */
+    private void sendPendingQueueMessages() {
+        //RTN19a
+        if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
+            for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+                try {
+                    send(queuedMessage.msg, false, null);
```
ikbalkaya, line 1520: Completion listener is ignored here

ikbalkaya (changes requested):

PR #852, opened 2022-11-15
Implement backoff and jitter timeout by spec RTB1

qsdigor:


lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -546,4 +555,198 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
...
+    /**
+     * Connect and check if timer time is jittered
+     * Spec: RTB1
+     */
+    @Test
+    public void disconnect_retry_channel_timeout_jitter() {
+        long oldRealtimeTimeout = Defaults.realtimeRequestTimeout;
+        int channelRetryTimeout = 150;
+        /* Reduce timeout for test to run faster */
+        Defaults.realtimeRequestTimeout = channelRetryTimeout;
```
owenpearson, line 642: This looks strange to me - you're setting the `realtimeRequestTimeout` to a value which you're calling `channelRetryTimeout`, which should be two separate timeouts. And is there a reason not to use the `channelRetryTimeout` client option here?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -546,4 +555,198 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
...
+
+                assertTrue("retry time higher range for count " + i + " is not in valid: " + retryTime + " expected: " + higherRange,
+                    retryTime < higherRange);
+                assertTrue("retry time lower range for count " + i + " is not in valid: " + retryTime + " expected: " + lowerRange,
+                    retryTime > lowerRange);
+            }
+            System.out.println("------------------------------------------------------------");
+        } catch (AblyException e) {
+            e.printStackTrace();
+            fail("init0: Unexpected exception instantiating library");
```
owenpearson, line 743: The `try` block includes a lot more than just instantiating the library so maybe this fail message should be more generic?

owenpearson (changes requested):
The implementation looks correct to me, I've left a couple of minor comments. I would also strongly consider using `Random` with a seed for testing if possible, that will give you a much cleaner and more predictable way of testing this.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
qsdigor: Setting channelRetryTimeout did not work alone and from [other test](https://github.com/ably/ably-java/blob/main/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java#L1726-L1801) I saw that setting the default value does the trick.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
qsdigor: I thing including exception in text is best way to go https://github.com/ably/ably-java/pull/852/commits/688f4b5fe52f4e07b3ed2cc78514a35b62d305d7

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
qsdigor: I am not sure if we should use Random seed as it will give us a static value when we expect a random range.
Random range helped me find the issue in the core implementation as well.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1805,6 +1818,7 @@ private boolean isFatalError(ErrorInfo err) {
     private CMConnectivityListener connectivityListener;
     private long connectionStateTtl = Defaults.connectionStateTtl;
     long maxIdleInterval = Defaults.maxIdleInterval;
+    private int disconnectedRetryCount = 0;
```
QuintinWillison, line 1821: I'm curious as to why this field has a `disconnected` prefix on its name when the (equivalent?) field you've added in `ChannelBase` is just `retryCount`. Does it carry extra meaning in this class?

QuintinWillison (changes requested):

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -81,6 +82,8 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
      */
     public ChannelProperties properties = new ChannelProperties();
 
+    public int retryCount = 0;
```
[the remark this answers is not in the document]
qsdigor: You are absolutely correct, it is fixed here https://github.com/ably/ably-java/pull/852/commits/c2f52bc4a92b4009c5355c4a786ab4e1ac6db2b3

KacperKluka (approved):
LGTM 👍

ikbalkaya (approved):
I have not carefully checked spec against this change and I just generally went through code which looks good to me

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
qsdigor: @QuintinWillison what should we do here?

owenpearson: The reason for the difference here is that a connection will retry while in the `DISCONNECTED` state and also while in the `SUSPENDED` state, so the `disconnectedRetryTimeout` is named as such to make it clear that this variable only counts the number of retries while in the `DISCONNECTED` state. Realtime channels only retry while they are `SUSPENDED` so there is no distinction to be made in that case. That being said, I have no objection to calling it `ChannelBase.suspendedRetryCount` if that makes it clearer.

QuintinWillison: Thanks, both. If it's matching the JS implementation then let's keep it as is. I'm also working on a new release this afternoon so this could nicely slip into that also. 😁

QuintinWillison (approved):

lib/src/test/java/io/ably/lib/test/realtime/RealtimeConnectFailTest.java
```diff
@@ -546,4 +555,196 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
...
+            System.out.println("wait done in: " + (waitCount / 10) + " seconds");
+
+            assertTrue("Disconnect retry was not finished, count was: " + retryCount.get(), retryCount.get() >= 4);
+
+            //check for all received retry times in onConnectionStateChanged callback
+            System.out.println("------------------------------------------------------------");
+            for (int i = 0; i < retryValues.size(); i++) {
+                long retryTime = retryValues.get(i);
+                long higherRange = disconnectedRetryTimeout + Math.min(i, 3) * 50L;
+                double lowerRange = 0.6 * disconnectedRetryTimeout + Math.min(i, 3) * 50L;
```
sacOO7, line 613: change 
```
double lowerRange = 0.6 * disconnectedRetryTimeout + Math.min(i, 3) * 50L;
```
to
```
double lowerRange = 0.8 * (disconnectedRetryTimeout + Math.min(i, 3) * 50L);
```

lib/src/test/java/io/ably/lib/util/TimerUtilsTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class TimerUtilsTest {
+
+    @Test
+    public void timer_retry_time_is_incremental() {
+        for (int i = 1; i <= 5; i++) {
+            int defaultTimerMs = 150;
+            int timerMs = TimerUtil.getRetryTime(defaultTimerMs, i);
+            long higherRange = defaultTimerMs + Math.min(i, 3) * 50L;
+            double lowerRange = 0.3 * defaultTimerMs + Math.min(i, 3) * 50L;
```
sacOO7, line 15: change to
```
double lowerRange = 0.8 * (disconnectedRetryTimeout + Math.min(i, 3) * 50L);
```

PR #856, opened 2022-11-18
Provide a path for user when transport is null

ikbalkaya:
This change provides a path where `sendImpl` function returns without a callback or through throwing an exception. This PR addresses that without changing synchronization policy of the `ConnectionManager`

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1544,6 +1544,7 @@ public class ConnectionManager implements ConnectListener {
     private void sendImpl(ProtocolMessage message, CompletionListener listener) throws AblyException {
         if(transport == null) {
             Log.v(TAG, "sendImpl(): Discarding message; transport unavailable");
+            listener.onError(new ErrorInfo("Unable to send message; transport unavailable", 500, 50000));
             return;
         }
         if(ProtocolMessage.ackRequired(message)) {
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1543,13 +1543,9 @@ public class ConnectionManager implements ConnectListener {
 
     private void sendImpl(ProtocolMessage message, CompletionListener listener) throws AblyException {
         if(transport == null) {
-            Log.v(TAG, "sendImpl(): Discarding message; transport unavailable");
-            if (listener != null){
-                listener.onError(new ErrorInfo("Unable to send message; transport unavailable", 80009, 80009));
-            }
-
-            return;
+            throw AblyException.fromErrorInfo(new ErrorInfo("Unable to send message; transport unavailable", 80009, 80009));
         }
+
         if(ProtocolMessage.ackRequired(message)) {
             message.msgSerial = msgSerial++;
             pendingMessages.push(new QueuedMessage(message, listener));
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1544,7 +1544,10 @@ public class ConnectionManager implements ConnectListener {
     private void sendImpl(ProtocolMessage message, CompletionListener listener) throws AblyException {
         if(transport == null) {
             Log.v(TAG, "sendImpl(): Discarding message; transport unavailable");
-            listener.onError(new ErrorInfo("Unable to send message; transport unavailable", 500, 50000));
+            if (listener != null){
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1544,6 +1544,10 @@ public void send(ProtocolMessage msg, boolean queueEvents, CompletionListener li
     private void sendImpl(ProtocolMessage message, CompletionListener listener) throws AblyException {
         if(transport == null) {
```
QuintinWillison, line 1545: It looks like you're checking against a field (reference) here and then later on using that same reference to call `send` on. Without looking at the threading model in more depth, I would worry that it might become `null` between here and that `send` call. I wonder if it might be better to cache the reference to a local `final` variable at the start of this method's implementation and then use that local throughout.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1544,6 +1544,10 @@ public void send(ProtocolMessage msg, boolean queueEvents, CompletionListener li
...
             Log.v(TAG, "sendImpl(): Discarding message; transport unavailable");
+            if (listener != null){
+                listener.onError(new ErrorInfo("Unable to send message; transport unavailable", 500, 50000));
```
QuintinWillison, line 1548: From what I know of our [error code ranges](https://github.com/ably/ably-common/tree/main/protocol#ranges) then the `5xx`/`5xxxx` range is intended to be used when we've had an HTTP `5xx` response back from an endpoint. Therefore, I wonder if `1xx`/`1xxxxx` was intended for use in this kind of scenario.

QuintinWillison (commented):
I'm a bit worried that this is quite a big change of behaviour, albeit for an edge case, that appears not to be defined in [the spec](https://sdk.ably.com/builds/ably/specification/main/features/) (that I can find, or am aware of). 🤔

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: I have not written this part 🤷

KacperKluka (commented):
Overall it looks good but I wonder if it would be possible to add a test for it? Can we mock no internet connection easily in our integration tests? 🤔

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: I think 80009 makes sense so I updated it here b72945f1529331d3beafa8f1aff4d8a08793e96f . However I am not sure whether status and error codes be the same -> I haven't seen status codes for error codes above 50000

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
QuintinWillison: Indeed, but at the point this was written then this method implementation only read the `transport` field once. You've changed that.

ikbalkaya: > then this method implementation only read the transport field once. You've changed that.

Sorry can you clarify this please? This field was already reading this and returning if the transport was null. Do you mean that this field isn't synchronized  and the state could change after callback and return?

QuintinWillison: Yes, so I am worried about the field value changing from being a valid reference to `null` between your null-check and the subsequent use of the field to call `send`.

ikbalkaya: Of course. But I cannot see how this differs from the method returning after checking the `transport` which can change in the point of return which is also a synchronization issue. What I do here is to provide a path (in this case a completion listener) so that they can have opportunity to respond as this method was originally designed, in a thread confined manner

ikbalkaya:
> Overall it looks good but I wonder if it would be possible to add a test for it? Can we mock no internet connection easily in our integration tests? 🤔

I will try to write a test for this - but I don't think this would be an easy case to write test for

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: Until we understand the circumstances in which this happens, how can we decide the error reason to return?

ikbalkaya: This case was checking for `transport` and would return if it wasn't available. So I thought that was a suitable message/code. I didn't think about adding additional message / reason about the circumstances on why this is happening (in this case a concurrency issue). Would you have a recommendation on the error message code?

PR #858, opened 2022-11-21
Revert to protocol 1.1

KacperKluka:
We will be reverting to protocol 1.1 in order to quick-fix an urgent issue with automatic presence re-enter. The protocol will be bumped to 1.2 again once the presence issue is fixed on the client side.

This PR aims to revert the work done in https://github.com/ably/ably-java/pull/591

PR #864, opened 2022-11-24
Revert to protocol 1.0

### LINKED ISSUE #857: Automatic presence re-enter after network connection is back does not work
We've discovered that a channel's presence is not automatically re-entered once you lose an internet connection and then it comes back on. After some further investigation, it turns out that this feature was working on older versions of `ably-java` SDK but it stopped working since version `1.2.2`. Looking at the [release notes](https://github.com/ably/ably-java/releases/tag/v1.2.2) of version `1.2.2` the most likely cause of this issue is the "Bump protocol version to 1.2".

**Steps to reproduce:**
- Create an `AblyReatlime` instance
- Join a channel's presence
- Verify that the client has joined the channel's presence (e.g. with the Ably dev dashboard)
- Turn the internet connection off
- Wait for the automatic "leave" event on the channel's presence
- Turn the internet connection on
- Wait for the automatic "enter" event

What should happen: The client should automatically re-enter the channel's presence after a few seconds
What happens: The client never re-enters the channel's presence

QuintinWillison:
This is our second attempt to quick-fix an urgent issue relating to presence re-entry.
The first attempt was in #858.

Closes #857.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -8,7 +8,7 @@ import java.text.DecimalFormatSymbols;
 import java.util.Locale;
 
 public class Defaults {
-    public static final float ABLY_VERSION_NUMBER   = 1.1f;
+    public static final float ABLY_VERSION_NUMBER   = 1.0f;
 
     /**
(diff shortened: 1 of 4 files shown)


KacperKluka (approved):
LGTM 👍

SimonWoolf (approved):

davyskiba (approved):

ikbalkaya (approved):

PR #876, opened 2022-12-06
Extract Ably error codes to enum

qsdigor:
Extracted all Ably error codes into AblyErrors enum class with code and message for each one of them presented [here](https://github.com/ably/ably-common/blob/main/protocol/errors.json).

### DIFF
--- a/lib/src/main/java/io/ably/lib/util/AblyErrors.java
+++ b/lib/src/main/java/io/ably/lib/util/AblyErrors.java
@@ -34,7 +34,7 @@ public enum AblyErrors {
     INVALID_PUBLISH_REQUEST_IMPERMISSIBLE_EXTRAS(40032, "Invalid publish request (impermissible extras field)"),
     RESERVED_FOR_ARTIFICIAL_ERROR(40099, "Reserved for artificial errors for testing"),
     UNAUTHORIZED(40100, "unauthorized"),
-    INVALID_CREDENTIALS_2(40101, "invalid credentials"),
+    INVALID_CREDENTIALS_AUTH(40101, "invalid credentials"),
     INCOMPATIBLE_CREDENTIALS(40102, "incompatible credentials"),
     INVALID_USE_OF_BASIC_AUTH(40103, "invalid use of Basic auth over non-TLS transport"),
     TIMESTAMP_NOT_CURRENT(40104, "timestamp not current"),
--- a/lib/src/main/java/io/ably/lib/util/AblyErrors.java
+++ b/lib/src/main/java/io/ably/lib/util/AblyErrors.java
@@ -1,0 +1,189 @@
+package io.ably.lib.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum AblyErrors {
+
+    NO_ERROR(10000, "no error"),
+
+    BAD_REQUEST(40000, "bad request"),
+    INVALID_REQUEST_BODY(40001, "invalid request body"),
+    INVALID_PARAMETER_NAME(40002, "invalid parameter name"),
+    INVALID_PARAMETER_VALUE(40003, "invalid parameter value"),
+    INVALID_HEADER(40004, "invalid header"),
+    INVALID_CREDENTIALS(40005, "invalid credential"),
+    INVALID_CONNECTION_ID(40006, "invalid connection id"),
+    INVALID_MESSAGE_ID(40007, "invalid message id"),
+    INVALID_CONTENT_LENGTH(40008, "invalid content length"),
+    MAXIMUM_LENGHT_EXCEEDED(40009, "maximum message length exceeded"),
+    INVALID_CHANNEL_NAME(40010, "invalid channel name"),
+    STALE_RING_STATE(40011, "stale ring state"),
+    INVALID_CLIENT_ID(40012, "invalid client id"),
+    INVALID_MESSAGE_DATA_OR_ENCODING(40013, "Invalid message data or encoding"),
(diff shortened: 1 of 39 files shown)


lib/src/main/java/io/ably/lib/util/AblyErrors.java
```diff
@@ -0,0 +1,189 @@
...
+    CHANNEL_PRESENCE_LEAVE_ERROR(91002, "unable to leave presence channel that is not entered"),
+    CHANNEL_PRESENCE_ENTER_LIMIT_EXCEED(91003, "unable to enter presence channel (maximum member limit exceeded)"),
+    CHANEL_PRESENCE_RE_ENTER_ERROR(91004, "unable to automatically re-enter presence channel"),
+    PRESENCE_STATE_BAD_SYNC(91005, "presence state is out of sync"),
+    MEMBER_LEFT_IMPLICITLY(91100, "member implicitly left presence channel (connection closed)");
+
+    public final int code;
+    public final String message;
+    private static final Map<Integer, AblyErrors> BY_CODE = new HashMap<>();
+    private static final Map<String, AblyErrors> BY_MESSAGE = new HashMap<>();
```
ikbalkaya, line 168: I can't see any of these (valueOf methods that use`BY_CODE`, BY_MESSAGE) are used in codebase - Can you remove them?

ikbalkaya (changes requested):
I have not compared the replacements in detail, just a request for newly created enum

lib/src/main/java/io/ably/lib/util/AblyErrors.java
qsdigor: I have removed it https://github.com/ably/ably-java/pull/876/commits/15aff9046985575fc8642c138a86395b41c71eea

QuintinWillison (changes requested):
While I recognise that this change is well intentioned, I have concerns around scope. I think there is opportunity to have a more significant impact on the quality and maintainability of the codebase by going further. In current state of this pull request:

- Only the `code` for `ErrorInfo` instances is populated from the new data, meaning that `statusCode` and `message` values are still floating at call sites (to `ErrorInfo` constructor, generally)
- The `message` values are now repeated in two places in this codebase, which wasn't the case previously

I can't see any reason why in many, if not all, cases the `ErrorInfo` instance should not be fully instantiated from the values encapsulated by the new `enum` values.

Also, of course, there could be the potential for us to auto-generate the `enum` values from our [canonical definitions](https://github.com/ably/ably-common/blob/main/protocol/errors.json).

I think this will need discussion with other team members but we're heading into Christmas holidays season, so this is a discussion we should pick up in the new year.

lib/src/main/java/io/ably/lib/util/AblyErrors.java
```diff
@@ -0,0 +1,169 @@
+package io.ably.lib.util;
+
+public enum AblyErrors {
```
sacOO7, line 3: This can be modified same as https://github.com/ably/ably-java/pull/881#discussion_r1059132748. 
Create a separate static class 
```
    // Also, more helpful in IDE code indexing/lookups
    public static class AblyErrorCode {
        public static final int NO_ERROR = 10000;
        public static final int BAD_REQUEST= 40000;
        ......
    }  
```

This is also more readable.
```
System.out.println(AblyErrorCode.BAD_REQUEST);
```

I think the message is almost the same for most of the string constants, so there's no point in keeping it. 
If you want to add a description to some of those messages, you can write an extra static method

```
 public static class AblyErrorCode {
        public static final int NO_ERROR = 10000;
        public static final int BAD_REQUEST= 40000;
        ......
        
      static string describe(int errorCode) {
       switch(errorCode) {
         case NO_ERROR:
              return "no error";
           }
        }
    }
```
Usage

```
System.out.println(AblyErrorCode.describe(AblyErrorCode.BAD_REQUEST));
```

The main aim is to simplify the readability of constants. Enum just makes it difficult to use at places and doesn't fulfill the purpose.

qsdigor: Agreed, I will implement your suggestions.

lib/src/main/java/io/ably/lib/util/AblyError.java
```diff
@@ -0,0 +1,160 @@
+package io.ably.lib.util;
+
+public class AblyError {
+    public static final int NO_ERROR = 10000; //No error
+
+    public static final int BAD_REQUEST = 40000; // "bad request"
```
sacOO7, line 6: Is this a doc comment? 
If it is, then we are good, otherwise you can add doc comment at the top 👍

lib/src/main/java/io/ably/lib/util/AblyErrors.java
sacOO7: Thanks!

sacOO7 (approved):
LGTM : )

lib/src/main/java/io/ably/lib/util/AblyError.java
```diff
@@ -0,0 +1,160 @@
+package io.ably.lib.util;
+
+public class AblyError {
```
sacOO7, line 3: Can we use className as `AblyErrorCode` instead of `AblyError` ? 
Because, this class only contains integer error codes.

sacOO7 (changes requested):

lib/src/main/java/io/ably/lib/util/AblyError.java
qsdigor: These are regular comments, if I make doc comments the whole class will lose readability as these codes won't look like list.

PR #881, opened 2022-12-12
Extract HTTP status codes in HttpConstants enumeration

qsdigor:
Extracted all HTTP status codes into HttpCodes enum class with code and message for each one of them presented [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status).


CHANGELOG.md
```diff
@@ -1,4 +1,15 @@
 # Change Log
+## [1.2.21](https://github.com/ably/ably-java/tree/v1.2.21)
```
QuintinWillison, line 2: Something is wrong with the state of this branch if it's showing changes made in a release PR. Looks like a merge operation has gone wrong at some point.

QuintinWillison (changes requested):
This pull request is not revieable in its current state.

CHANGELOG.md
qsdigor: It is fixed now.

sacOO7:
@qsdigor I strongly feel, it's better to use static class instead of complex enum. So, replacement will be 
```
    public static class HttpCode {
        public static final int  CONTINUE = 100;
        public static final int  SWITCHING_PROTOCOLS= 101;
        ......
    }  
```
So, while accessing, you just need to do 
```
System.out.println(HttpCode.OK) // This is more readable
```
instead of 
```
System.out.println(OK.code) // Not readable
```

I don't feel the message field specified in enum class `HttpCodes`  provides any special meaning, rather adds more duplication for the same string constant.

lib/src/main/java/io/ably/lib/util/HttpCodes.java
```diff
@@ -0,0 +1,86 @@
+package io.ably.lib.util;
+
+public enum HttpCodes {
```
sacOO7, line 3: @qsdigor I strongly feel, it's better to use static class instead of complex enum. So, replacement will be 
```
    public static class HttpCode {
        public static final int  CONTINUE = 100;
        public static final int  SWITCHING_PROTOCOLS= 101;
        ......
    }  
```
So, while accessing, you just need to do 
```
System.out.println(HttpCode.OK) // This is more readable
```
instead of 
```
System.out.println(OK.code) // Not readable
```

I don't feel the message field specified in enum class `HttpCodes`  provides any special meaning, rather adds duplication for the same string constant.

sacOO7 (changes requested):

lib/src/main/java/io/ably/lib/util/HttpCodes.java
sacOO7: Enums are not much flexible in Java. So, most of the time, they do not fit in the expected use case.

qsdigor: Agreed, I will implement your suggestions.

sacOO7 (approved):
LGTM : )

lib/src/main/java/io/ably/lib/util/HttpCodes.java
sacOO7: Looks good and readable now : )

lib/src/main/java/io/ably/lib/util/HttpCode.java
```diff
@@ -0,0 +1,199 @@
...
+    public static final int BAD_GATEWAY = 502;
+    public static final int SERVICE_UNAVAILABLE = 503;
+    public static final int GATEWAY_TIMEOUT = 504;
+    public static final int HTTP_VERSION_NOT_SUPPORTED = 505;
+    public static final int VARIANTS_ALSO_NEGOTIATES = 506;
+    public static final int INSUFFICIENT_STORAGE = 507;
+    public static final int NOT_EXTENDED = 510;
+    public static final int NETWORK_AUTHENTICATION_REQUIRED = 511;
+
+    public static String describe(int errorCode) {
```
sacOO7, line 74: Can you rename the argument as `httpCode` instead of `errorCode`. Because, not all of the HttpCodes are errorCodes. Only codes with value >= 400 are error codes. 
Sorry for suggesting wrong name in the example.

sacOO7 (changes requested):

lib/src/main/java/io/ably/lib/util/HttpCode.java
qsdigor: Renamed here https://github.com/ably/ably-java/pull/881/commits

sacOO7 (approved):
LGTM : )

PR #883, opened 2022-12-29
Skip checking WS hostname when not using SSL

cruickshankpg:
otherwise an exception is raised as there is no SSL session to check


KacperKluka (approved):
I don't have much domain knowledge about this problem but it looks sensible 👍

paddybyers (approved):
I think the logic in the existing code is a bit of a mess and could be improved, but this change LGTM

QuintinWillison (approved):

cruickshankpg:
I am not authorised to merge PRs to SDKs so would someone else kindly do the honours? :)

Issue #885, opened 2023-01-04
Re-attach fails due to previous detach request


ikbalkaya:
Context : https://github.com/ably/ably-asset-tracking-android/issues/863

According to issue above, there can be a case where a DETACH is acknowledged after another ATTACH request is sent which leads the ATTACH to fail and channel go into suspended.

Todos 
* Check if there are existing tests for the spec points @paddybyers posted here https://github.com/ably/ably-asset-tracking-android/issues/863#issuecomment-1370191834 . Modify and create new tests to confirm the issue.

ikbalkaya:
@paddybyers I am a bit confused about timers used during attach/detach or reattach operations.
For example `attachTimer` is scheduled here
https://github.com/ably/ably-java/blob/f9e491999b194a6ad250c003ce1202e56eede080/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L419
And also `attachTimer` is scheduled here on detach
https://github.com/ably/ably-java/blob/f9e491999b194a6ad250c003ce1202e56eede080/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L517

Can you provide some context on why this timers are used please? Thanks

paddybyers:
See https://sdk.ably.com/builds/ably/specification/main/features/#RTL4f

> Once an `ATTACH` ProtocolMessage is sent, if an `ATTACHED` ProtocolMessage is not received within the [default realtime request timeout](https://sdk.ably.com/builds/ably/specification/main/features/#defaults), the attach request should be treated as though it has failed and the channel should transition to the `SUSPENDED` state. The channel will then be subsequently automatically re-attached as described in [RTL13](https://sdk.ably.com/builds/ably/specification/main/features/#RTL13)

The timer is used to implement this requirement

ikbalkaya:
This is fixed : https://github.com/ably/ably-java/pull/887

PR #887, opened 2023-01-05
Fix attach/detach race condition

ikbalkaya:
This PR improves spec conformance for [RTL4h](https://sdk.ably.com/builds/ably/specification/main/features/#RTL4h) and [RTL5i](https://sdk.ably.com/builds/ably/specification/main/features/#RTL5i)
However I think the spec might be a bit misleading (Please correct me if I'm wrong)
In RTL4h, it writes

> If the channel is in a pending state DETACHING or ATTACHING, do the attach operation after the completion of the pending request

I'm not sure if the attach operation should be done after an attach already happened (no need for adding a pending attach). Currently it just register a completion listener if provided and then returns which I think is the correct behaviour. I changed the behaviour when the channel was in DETACHING state so it wouldn't send an attach right away and add a pending attach request to be sent after completion of detach.

The same applies for RTL5i, but in reverse order 

Closes https://github.com/ably/ably-java/issues/885

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
@@ -1071,6 +1071,51 @@ public class RealtimeChannelTest extends ParameterizedTest {
     }
 
     /**
+     * When client detaches from a channel in attaching state, verify that detach call will be done after attach
+     * response is received
+     * verify attach {@code CompletionListener#onSuccess()} gets called.
+     */
+    // Spec: RTL5i
+    // https://github.com/ably/ably-java/issues/885
+    @Test
+    public void detach_when_channel_in_attaching_state() throws AblyException {
+        AblyRealtime ably = null;
+        try {
+            ClientOptions opts = createOptions(testVars.keys[0].keyStr);
+            opts.logLevel = Log.VERBOSE;
+            ably = new AblyRealtime(opts);
+
+            /* wait until connected */
+            (new ConnectionWaiter(ably.connection)).waitFor(ConnectionState.connected);
+            assertEquals("Verify connected state reached", ably.connection.state, ConnectionState.connected);
+
+            /* create a channel and attach */
+            final String channelName = "attach_channel";
+            final Channel channel = ably.channels.get(channelName);
+            final Helpers.CompletionWaiter attachCompletionWaiter = new Helpers.CompletionWaiter();
+            channel.attach(attachCompletionWaiter);
+            assertEquals("Verify detaching state reached", channel.state, ChannelState.attaching);
+            //immediately start detach operation
+            final Helpers.CompletionWaiter detachCompletionWaiter = new Helpers.CompletionWaiter();
+            channel.detach(detachCompletionWaiter);
+
+            new ChannelWaiter(channel).waitFor(ChannelState.attached);
+            assertEquals("Verify attached state reached", channel.state, ChannelState.attached);
+
+            //now wait for detach to complete
+            new ChannelWaiter(channel).waitFor(ChannelState.detached);
+            /* detach */
+            detachCompletionWaiter.waitFor();
+
+            assertThat(detachCompletionWaiter.success,is(true));
+            assertThat(channel.state, is(ChannelState.detached));
+        } finally {
+            if(ably != null)
+                ably.close();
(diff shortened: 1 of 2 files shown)


davyskiba (changes requested):

davyskiba (approved):

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1019,6 +1021,100 @@ public void detach_success_callback_detaching() throws AblyException {
...
+    public void detach_when_channel_in_attaching_state() throws AblyException {
+        AblyRealtime ably = null;
+        try {
+            ClientOptions opts = createOptions(testVars.keys[0].keyStr);
+            opts.logLevel = Log.VERBOSE;
+            ably = new AblyRealtime(opts);
+
+            /* wait until connected */
+            (new ConnectionWaiter(ably.connection)).waitFor(ConnectionState.connected);
+            assertEquals("Verify connected state reached", ably.connection.state, ConnectionState.connected);
```
QuintinWillison, line 1090: Minor observation, and I recognise that lots of other sites in this codebase do this wrong, but the JUnit API wants `expected` first, followed by `actual`, as this ordering is used in some of the log messages it outputs. Therefore, for this line:

```suggestion
            assertEquals("Verify connected state reached", ConnectionState.connected, ably.connection.state);
```

This also applies to a few other asserts in this test method.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1019,6 +1021,100 @@ public void detach_success_callback_detaching() throws AblyException {
...
+            //immediately start detach operation
+            final Helpers.CompletionWaiter detachCompletionWaiter = new Helpers.CompletionWaiter();
+            channel.detach(detachCompletionWaiter);
+
+            new ChannelWaiter(channel).waitFor(ChannelState.attached);
+            assertEquals("Verify attached state reached", channel.state, ChannelState.attached);
+
+            //now wait for detach to complete
+            new ChannelWaiter(channel).waitFor(ChannelState.detached);
+            /* detach */
```
QuintinWillison, line 1107: This comment seems to be floating a bit, also not in the same style as some of your other single line comments. Perhaps worth removing or conforming?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -1019,6 +1021,100 @@ public void detach_success_callback_detaching() throws AblyException {
...
+            assertEquals("Verify detaching state reached", channel.state, ChannelState.attaching);
+            //immediately start detach operation
+            final Helpers.CompletionWaiter detachCompletionWaiter = new Helpers.CompletionWaiter();
+            channel.detach(detachCompletionWaiter);
+
+            new ChannelWaiter(channel).waitFor(ChannelState.attached);
+            assertEquals("Verify attached state reached", channel.state, ChannelState.attached);
+
+            //now wait for detach to complete
+            new ChannelWaiter(channel).waitFor(ChannelState.detached);
```
QuintinWillison, line 1106: For readability/clarity and consistency with the earlier construction of a `ConnectionWaiter` instance, this could be:

```suggestion
            (new ChannelWaiter(channel)).waitFor(ChannelState.detached);
```

QuintinWillison (approved):
Minor comments that don't change the execution, but might improve the maintainability.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
ikbalkaya: I had also realized this elsewhere and I seem to apply the same mistakes. I have change orders d0d8665517582da836d7e41ac2ac8f95851eb0f4 and created an issue to fix others: https://github.com/ably/ably-java/issues/892

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
ikbalkaya: 21c85a9daabdc704d044ba4ed08837ae7d046623

Issue #863, opened 2022-11-23
`auth_renewAuth_callback_invoked` test failing with `AssertionError` "Callback not invoked"


QuintinWillison:
From [this test run](https://github.com/ably/ably-java/actions/runs/3532982505/jobs/5929397279#step:3:885):

```
io.ably.lib.test.realtime.RealtimeAuthTest > auth_renewAuth_callback_invoked[binary_protocol] FAILED
    java.lang.AssertionError: Callback not invoked
```

As discussed in [this Slack thread (internal)](https://ably-real-time.slack.com/archives/C019LDWM9UH/p1669218490931509) this test was added in https://github.com/ably/ably-java/pull/825 and a working theory is that this might be a regression as a result of the `1.2` to `1.1` protocol downgrade in https://github.com/ably/ably-java/pull/858.

QuintinWillison:
Seen to fail [here](https://github.com/ably/ably-java/actions/runs/3854323429/jobs/6574733842#step:3:528):

```
io.ably.lib.test.realtime.RealtimeSuite > io.ably.lib.test.realtime.RealtimeAuthTest.auth_renewAuth_callback_invoked[binary_protocol] FAILED
    java.lang.AssertionError: Callback not invoked
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at io.ably.lib.test.realtime.RealtimeAuthTest.auth_renewAuth_callback_invoked(RealtimeAuthTest.java:1121)
```

PR #890, opened 2023-01-10
Ignore consistently failing test : auth_renewAuth_callback_invoked

ikbalkaya:
This PR ignores a test that is mostly failing on CI : auth_renewAuth_callback_invoked

There is already an issue for this : https://github.com/ably/ably-java/issues/863 and we should be addressing fixing from that issue

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
@@ -1052,6 +1052,7 @@ public class RealtimeAuthTest extends ParameterizedTest {
         }
     }
 
+    @Ignore("Fix flakey test")
     @Test
     public void auth_renewAuth_callback_invoked() throws InterruptedException {
         try {
--- a/lib/src/main/java/io/ably/lib/transport/Defaults.java
+++ b/lib/src/main/java/io/ably/lib/transport/Defaults.java
@@ -8,7 +8,7 @@ import java.text.DecimalFormatSymbols;
 import java.util.Locale;
 
 public class Defaults {
-    public static final float ABLY_VERSION_NUMBER   = 1.0f;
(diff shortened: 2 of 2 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
```diff
@@ -1099,17 +1099,12 @@ public void onRawMessageRecv(ProtocolMessage message) {
...
-                } catch (InterruptedException ie) {
-                    fail("auth_expired_token_expire_renew protocolListener.wait(): interrupted -" + ie.getMessage());
-                }
-            }
+            ably.connect();
 
             final Helpers.ConnectionWaiter connectionWaiter = new Helpers.ConnectionWaiter(ably.connection);
             boolean isConnected = connectionWaiter.waitFor(ConnectionState.connected, 1, 4000L);
+            //wait for the token to expire
+            wait(1000);
```
QuintinWillison, line 1107: A couple of observations about this time value:

1. It is floating here, disassociated from the token params' TTL defined earlier (maintainability)
2. It matches exactly the token params' TTL value which might be a risk of not waiting _quite enough_ at times, so could remain 'flakey' in nature 

So... I would, earlier on, create a `final` local variable containing the TTL value programmed into the token details. Then use that here, with a multiplier (perhaps x2 to be sure) - e.g. `tokenTtl * 2`.

QuintinWillison (changes requested):

davyskiba (approved):

lib/src/test/java/io/ably/lib/test/realtime/RealtimeAuthTest.java
ikbalkaya: Resolving in @QuintinWillison's behalf as this is no longer the case

PR #893, opened 2023-01-11

AndyTWF:
RTE3 requires that adding the same listener to the EventEmitter twice should
have the effect of that listener being invoked twice for a given event. This is currently not the case.

This change makes EventEmitter compliant with RTE3 by allowing listeners
to be added multiple times.


AndyTWF:
Going to close this one and rebase off of the integration branch

PR #894, opened 2023-01-11

### LINKED ISSUE #888: EventEmitter.on() does not permit the same listener to be added multiple times
https://github.com/ably/ably-java/blob/9786de65bd44c2f139a53ee235535cb7617cbc87/lib/src/main/java/io/ably/lib/util/EventEmitter.java#L41

The single-argument implementation of EventEmitter.on() prevents the same listener from being registered multiple times (and thus being invoked multiple times per event). It should allow this to happen as per https://sdk.ably.com/builds/ably/specification/main/features/#RTE4.

The multi-argument version of the method is unaffected.

AndyTWF:
RTE3 requires that adding the same listener to the EventEmitter twice should have the effect of that listener being invoked twice for a given event. This is currently not the case.

This change makes EventEmitter compliant with RTE3 by allowing listeners to be added multiple times.

Fixes #888


paddybyers:
The fix LGTM but are we comfortable that this change in behaviour is something we can make in a minor or patch release?

AndyTWF:
@paddybyers my apologies I must have gotten the wrong branch - it should have been targeting ably-java 2.0, I'll recreate the PR for the correct branch :)

PR #895, opened 2023-01-12

### LINKED ISSUE #888: EventEmitter.on() does not permit the same listener to be added multiple times

AndyTWF:
RTE3 requires that adding the same listener to the EventEmitter twice should have the effect of that listener being invoked twice for a given event. This is currently not the case.

This change makes EventEmitter compliant with RTE3 by allowing listeners to be added multiple times.

This is a BREAKING CHANGE as it changes the behaviour of EventEmitter that customers may be relying on.

Fixes #888

### DIFF
--- a/core/src/main/java/io/ably/lib/util/EventEmitter.java
+++ b/core/src/main/java/io/ably/lib/util/EventEmitter.java
@@ -1,10 +1,7 @@
 package io.ably.lib.util;
 
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 
 /**
  * A generic interface for event registration and delivery used in a number of the types in the Realtime client library.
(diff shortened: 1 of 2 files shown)


jamienewcomb:
@AndyTWF are we planning to do anything with this or shall we close it?
cc @paddybyers

AndyTWF:
> @AndyTWF are we planning to do anything with this or shall we close it? cc @paddybyers

As it's a breaking change, it would need `ably-java 2.0`. As a migration towards Kotlin is a possibility in the near future, I suggest we close this and defer it back to @ttypic and the team as part of that plan.

I'll close this :)

PR #900, opened 2023-01-13
Connection resumption improvements

ikbalkaya:
This PR introduces some fixes to connection resumption issues by implementing spec RTN15c1, RTN15c2 and RTN15c3. To summarize the rule of implementation.

* If connection resume is successful, current serial is set to the first serial of pending queue, queue is cleared and previously sent pending messages are added in front of queued messages. 
* If connection resume is failed the message serial is reset to 0 and all pending messages are added the queue to be sent with the newly assigned serials

Previously when there was a new connection established, channels were suspended and pending messages were reset. This should not be happening anymore

Also I have made some changes to general interface of `WebSocketTransport` to be able to mock blocking receives 
* There is a new `receive` method that uses a single method interface to delegate handling of sending messages to connectionManager. There are also some changes with MockWebSocketTransport that allows blocking incoming acks/nacks. This class is also now exposed so that we can read sent published messages directly. Please note that changes done here for these classes only serve the particular case of publishing. I think it can be improved to contain other protocol actions too.

Some changes are inspired and taken from https://github.com/ably/ably-java/pull/842

Closes https://github.com/ably/ably-java/issues/474

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -81,6 +81,8 @@ public class ConnectionManager implements ConnectListener {
         void onMessage(ProtocolMessage msg);
         void suspendAll(ErrorInfo error, boolean notifyStateChange);
         Iterable<Channel> values();
+
+        void reattachOnResumeFailure();
     }
 
     /***********************************
@@ -1183,44 +1185,38 @@ public class ConnectionManager implements ConnectListener {
     }
 
     private synchronized void onConnected(ProtocolMessage message) {
-        /* if the returned connection id differs from
-         * the existing connection id, then this means
-         * we need to suspend all existing attachments to
-         * the old connection.
-         * If realtime did not reply with an error, it
-         * signifies that this was a result of an earlier
-         * connection being invalidated due to being stale.
-         *
-         * Suspend all channels attached to the previous id;
-         * this will be reattached in setConnection() */
         ErrorInfo error = message.error;
-        if(connection.id != null && !message.connectionId.equals(connection.id)) {
-            /* we need to suspend the original connection */
-            if(error == null) {
-                error = REASON_SUSPENDED;
+
+        if (message.connectionId.equals(connection.id)) {
+            //RTN15c1
+            if (error == null) {
+                Log.d(TAG, "connection has reconnected and resumed successfully");
+                connection.reason = null;
+            }else { //RTN15c2
+                Log.d(TAG, "connection resume success with non-fatal error: " + error.message);
+                connection.reason = error;
             }
-            channels.suspendAll(error, false);
+        //make sure it's not a fresh connection
+        } else if (connection.id != null) { //RTN15c3
+            if (error != null){
+                Log.d(TAG, "connection resume failed with error: " + error.message);
+                connection.reason = error;
+            }
+            channels.reattachOnResumeFailure();
+            connection.connectionManager.msgSerial = 0;
         }
 
-        /* set the new connection id */
-        ConnectionDetails connectionDetails = message.connectionDetails;
-        connection.key = connectionDetails.connectionKey;
-        if (!message.connectionId.equals(connection.id)) {
-            /* The connection id has changed. Reset the message serial and the
-             * pending message queue (which fails the messages currently in
-             * there). */
-            pendingMessages.reset(msgSerial,
-                    new ErrorInfo("Connection resume failed", 500, 50000));
-            msgSerial = 0;
-        }
         connection.id = message.connectionId;
+
         if(message.connectionSerial != null) {
-            connection.serial = message.connectionSerial.longValue();
+            connection.serial = message.connectionSerial;
             if (connection.key != null)
                 connection.recoveryKey = connection.key + ":" + message.connectionSerial;
         }
 
+        ConnectionDetails connectionDetails = message.connectionDetails;
         /* Get any parameters from connectionDetails. */
+        connection.key = connectionDetails.connectionKey; //RTN16d
         maxIdleInterval = connectionDetails.maxIdleInterval;
         connectionStateTtl = connectionDetails.connectionStateTtl;
 
@@ -1233,10 +1229,33 @@ public class ConnectionManager implements ConnectListener {
             return;
         }
 
+        //RTN19a
+        sendPendingQueueMessages();
+
         /* indicated connected currentState */
         setSuspendTime();
         requestState(new StateIndication(ConnectionState.connected, error));
     }
+
+    /**
+     * Send all pending messages which are in the queue.
+     * Remove them from the queue once they are sent successfully
+     * Spec: RTN19a
+     */
+    private void sendPendingQueueMessages() {
+        //RTN19a
+        for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+            try {
+                send(queuedMessage.msg, false, null);
+            } catch (AblyException e) {
+                String errorString = String.format(Locale.ROOT, "Unable to send pending message %s (%s)",
+                    queuedMessage.msg.id, e.errorInfo.message);
+                Log.e(TAG, errorString);
+                connection.emitUpdate(e.errorInfo);
+            }
+        }
+    }
+
 
     private synchronized void onDisconnected(ProtocolMessage message) {
         ErrorInfo reason = message.error;
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1249,28 +1249,20 @@ public class ConnectionManager implements ConnectListener {
      * on pending queue, for example when a connection resume failed
      */
     private void addPendingMessagesToQueuedMessages(boolean resetMessageSerial) {
+        // Add messages from pending messages to front of queuedMessages in order to retry them
+        queuedMessages.addAll(0, pendingMessages.queue);
+        //rewind start serial back to the first serial since we are clearing the queue
+        if (!pendingMessages.queue.isEmpty()){
+            //Reset current serial to the first pending message on previous queue as we are going to clear the queue now
+            msgSerial =  pendingMessages.queue.get(0).msg.msgSerial;
+            pendingMessages.resetStartSerial((int) (msgSerial));
+            pendingMessages.clearQueue();
+        }
+
         //RTN19a
         if (resetMessageSerial){
             pendingMessages.resetStartSerial(0);
             msgSerial = 0; //msgSerial will increase in sendImpl when messages are sent
-        }
-        //put messages from pending messages to front of queuedMessages, the ones with the message serials will already
-        //have been assigned new message serial to them at this point
-        final int pendingMessageCount = pendingMessages.queue.size();
-        queuedMessages.addAll(0, pendingMessages.queue);
-        //Clear the pendingQueue now, because we do not want the retried messages to accumulate on it.
-        pendingMessages.clearQueue();
-
-
-        // reassign new serials for remaining queued messages if reset was required
-        if (resetMessageSerial) {
-            int startIndex = pendingMessageCount != 0 ? pendingMessageCount - 1 : 0;
-            for (int i = startIndex; i < queuedMessages.size(); i++) {
-                //if index is 0, it means there wasn't any previous pending messages so we use newly reset msgSerial as
-                //starting serial
-                final long previousMessageSerial = i == 0 ? msgSerial : queuedMessages.get(i - 1).msg.msgSerial;
-                queuedMessages.get(i).msg.msgSerial = previousMessageSerial + 1;
-            }
         }
     }
 
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1197,8 +1197,8 @@ public class ConnectionManager implements ConnectListener {
                     // RTN15c2: no action required wrt channel state
                     Log.d(TAG, "connection resume success with non-fatal error: " + error.message);
                 }
-                // send any messages still pending from the previous transport (RTN19a)
-                sendPendingQueueMessages();
+                // Add pending messages to the front of queued messages to be sent later
+                addPendingMessagesToQueuedMessages(false);
             } else {
                 // RTN15c3: resume failed
                 if (error != null){
@@ -1208,11 +1208,10 @@ public class ConnectionManager implements ConnectListener {
                 }
 
                 channels.reattachOnResumeFailure();
-                connection.connectionManager.msgSerial = 0;
-                // send any messages still pending from the previous transport (RTN19a)
-                // however, this time the re-sent pending messages have to have newly assigned `
+                // Add any messages still pending from the previous transport (RTN19a) to the front of queued messages
+                // however, this time the pending messages have to have newly assigned `
                 // msgSerial`s. They can't simply be replayed, as they are in the successful resume case
-                sendPendingQueueMessages();
+                addPendingMessagesToQueuedMessages(true);
             }
         }
 
@@ -1238,30 +1237,39 @@ public class ConnectionManager implements ConnectListener {
             requestState(transport, new StateIndication(ConnectionState.failed, e.errorInfo));
             return;
         }
-
-        //RTN19a
-        sendPendingQueueMessages();
-
         /* indicated connected currentState */
         setSuspendTime();
         requestState(new StateIndication(ConnectionState.connected, error));
     }
 
     /**
-     * Send all pending messages which are in the queue.
-     * Remove them from the queue once they are sent successfully
+     * Add all pending queued messages to the front of QueuedMessages for them to be sent later
      * Spec: RTN19a
+     * @param resetMessageSerial whether to reset message serial, this will determine whether to reset message serials
+     * on pending queue, for example when a connection resume failed
      */
-    private void sendPendingQueueMessages() {
+    private void addPendingMessagesToQueuedMessages(boolean resetMessageSerial) {
         //RTN19a
-        for (final QueuedMessage queuedMessage : pendingMessages.queue) {
-            try {
-                send(queuedMessage.msg, false, null);
-            } catch (AblyException e) {
-                String errorString = String.format(Locale.ROOT, "Unable to send pending message %s (%s)",
-                    queuedMessage.msg.id, e.errorInfo.message);
-                Log.e(TAG, errorString);
-                connection.emitUpdate(e.errorInfo);
+        if (resetMessageSerial){
+            pendingMessages.resetStartSerial(0);
+            msgSerial = 0; //msgSerial will increase in sendImpl when messages are sent
+        }
+        //put messages from pending messages to front of queuedMessages, the ones with the message serials will already
+        //have been assigned new message serial to them at this point
+        final int pendingMessageCount = pendingMessages.queue.size();
+        queuedMessages.addAll(0, pendingMessages.queue);
+        //Clear the pendingQueue now, because we do not want the retried messages to accumulate on it.
+        pendingMessages.clearQueue();
+
+
+        // reassign new serials for remaining queued messages if reset was required
+        if (resetMessageSerial) {
+            int startIndex = pendingMessageCount != 0 ? pendingMessageCount - 1 : 0;
+            for (int i = startIndex; i < queuedMessages.size(); i++) {
+                //if index is 0, it means there wasn't any previous pending messages so we use newly reset msgSerial as
+                //starting serial
+                final long previousMessageSerial = i == 0 ? msgSerial : queuedMessages.get(i - 1).msg.msgSerial;
+                queuedMessages.get(i).msg.msgSerial = previousMessageSerial + 1;
             }
         }
     }
@@ -1747,6 +1755,13 @@ public class ConnectionManager implements ConnectListener {
             startSerial = 0;
         }
 
+        public void resetStartSerial(int from) {
+             startSerial = from;
+        }
+
+        synchronized void clearQueue() {
+            queue.clear();
+        }
     }
 
     /***********************
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1185,25 +1185,35 @@ public class ConnectionManager implements ConnectListener {
     }
 
     private synchronized void onConnected(ProtocolMessage message) {
-        ErrorInfo error = message.error;
+        final ErrorInfo error = message.error;
+        connection.reason = error;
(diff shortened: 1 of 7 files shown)


davyskiba (approved):
Those changes make sense, but keep in my mind my relatively small ably-java knowledge

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1233,11 +1222,36 @@ private synchronized void onConnected(ProtocolMessage message) {
...
     }
 
+    /**
+     * Send all pending messages which are in the queue.
+     * Remove them from the queue once they are sent successfully
+     * Spec: RTN19a
+     */
+    private void sendPendingQueueMessages() {
+        //RTN19a
+        if (pendingMessages.queue != null && !pendingMessages.queue.isEmpty()) {
```
davyskiba, line 1240: Could `isEmpty` check be skipped without any impact? 🤔

ikbalkaya: Indeed both checks are not necessary. So I removed them both

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1183,44 +1185,31 @@ private void onChannelMessage(ProtocolMessage message) {
...
-         *
-         * Suspend all channels attached to the previous id;
-         * this will be reattached in setConnection() */
         ErrorInfo error = message.error;
-        if(connection.id != null && !message.connectionId.equals(connection.id)) {
-            /* we need to suspend the original connection */
-            if(error == null) {
-                error = REASON_SUSPENDED;
+
+        if (message.connectionId.equals(connection.id)) {
```
paddybyers, line 1190: Personally I think it's clearer to separate the resume logic explicitly:

```
connection.reason = message.error;
if (connection.id != null) { // there was a previous connection, so this is a resume and RTN15c applies
  if(message.connectionId.equals(connection.id)) {
    // resume succeeded
    if(message.error == null) {
      // RTN15c1: no action required wrt channel state
      Log.d(TAG, "connection has reconnected and resumed successfully");
    } else {
      // RTN15c2: no action required wrt channel state
      Log.d(TAG, "connection resume success with non-fatal error: " + error.message);
    }
    // send any messages still pending from the previous transport (RTN19a)
    sendPendingQueueMessages();
  } else {
    // RTN15c3: resume failed
    Log.d(TAG, "connection resume failed with error: " + error.message);
    channels.reattachOnResumeFailure();
    connection.connectionManager.msgSerial = 0;
    // send any messages still pending from the previous transport (RTN19a)
    // however, this time the re-sent pending messages have to have newly assigned `msgSerial`s. They can't simply be replayed, as they are in the successful resume case
    sendPendingQueueMessages();
  }
}
```

So perhaps `sendPendingQueueMessages()` needs a new boolean argument that tells it whether to reassign `msgSerial`s on the messages, starting from 0.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1233,11 +1222,34 @@ private synchronized void onConnected(ProtocolMessage message) {
...
         setSuspendTime();
         requestState(new StateIndication(ConnectionState.connected, error));
     }
 
+    /**
+     * Send all pending messages which are in the queue.
+     * Remove them from the queue once they are sent successfully
+     * Spec: RTN19a
+     */
+    private void sendPendingQueueMessages() {
```
paddybyers, line 1238: I think there are two issues here:
1) it needs to be possible, in the non-resume case, for messages to be reassigned a new `msgSerial` for the new connection

2) What happens to the messages where the `send()` failed with an exception? Are they left in the queue or should they be removed? And does the `CompletionListener` ever get called for them?

paddybyers (changes requested):
See comments inline.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
paddybyers: Also - have you thought about adding some tests?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: This is where the if branch makes resume case explicit. 
https://github.com/ably/ably-java/pull/900/commits/15797488ef39486a780182ba5db97a13ca504860

https://github.com/ably/ably-java/pull/900/commits/15797488ef39486a780182ba5db97a13ca504860 adds new boolean parameter to reset start serial for pending messages

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: > it needs to be possible, in the non-resume case, for messages to be reassigned a new msgSerial for the new connection

This should be happening https://github.com/ably/ably-java/pull/900/commits/62882a585949a428a9d6bf4f604f13fd4214df4c now

> What happens to the messages where the send() failed with an exception? Are they left in the queue or should they be removed? And does the CompletionListener ever get called for them?

I had not checked this as I used that portion of code directly from the other branch. But I think `QueuedMessage` itself should hold a reference to listener which seems to have been ignored here. 
I have passed the listener to send call 41ca4054be6b9a947cde6ffcbab8e2c35c07e6c9 but will check further whether SDK properly discards messages or keeps the messages in the queue. 

> Also - have you thought about adding some tests?
Yes, sorry I was going to but I manage to miss that. I will add any relevant tests

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1232,12 +1231,35 @@ private synchronized void onConnected(ProtocolMessage message) {
...
+     * @param resetMessageSerial
+     */
+    private void sendPendingQueueMessages(boolean resetMessageSerial) {
+        //RTN19a
+        if (resetMessageSerial){
+            pendingMessages.resetStartSerial(0);
+        }
+        for (final QueuedMessage queuedMessage : pendingMessages.queue) {
+            try {
+                send(queuedMessage.msg, false, queuedMessage.listener);
```
paddybyers, line 1252: This is incorrect. We don't call the `CompletionListener` just because the `send()` was successful; we call it when we get a response from the server with an `ACK` or `NACK`.
So here, if `send()` succeeds, we will keep the `QueuedMessage` in the queue, and when the server responds we can call the listener. But in the exception case when send() fails we have to decide what to do. Why did it fail? Is it because the connection dropped again? Or is there some other problem?

If we decide that the failure to `send()` means that we're going to give up on sending the message, then here we would remove the `QueuedMessage` from the queue, and call the listener with an error. However, I'm not aware of any circumstances where the `send()` would fail, other than connectivity problems (in which case we'd expect there to be a subsequent resume that triggered another retry attempt).

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1232,12 +1231,35 @@ private synchronized void onConnected(ProtocolMessage message) {
...
         requestState(new StateIndication(ConnectionState.connected, error));
     }
 
+    /**
+     * Send all pending messages which are in the queue.
+     * Remove them from the queue once they are sent successfully
+     * Spec: RTN19a
+     * @param resetMessageSerial
+     */
+    private void sendPendingQueueMessages(boolean resetMessageSerial) {
```
paddybyers, line 1245: Personally I would change `resetStartSerial()` so that it returns the next unassigned `msgSerial`.

Then I would remove this boolean `resetMessageSerial` argument from here.

The calling code above (`RTN15c3` case) should call `resetStartSerial()` directly, and use the result to set `this.msgSerial`.

paddybyers (changes requested):
Please see inline comments.

I suggest please that we tidy this and remove the whitespace-related commits, because the commit history is now a mess.

And we still need tests pls.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: > This is incorrect. We don't call the CompletionListener just because the send() was successful; we call it when we get a response from the server with an ACK or NACK.

Could you please clarify whether this is incorrect because of the following paragraph or because it's not doing what it's supposed to do as described here? As far as I can see, this completion listener will be passed to `PendingMessageQueue` in `sendImpl` if ack required, so ultimately this is going to get its callback from there.  However I realized that message queue inside `PendingMessageQueue` never clears. So I added fa0a4bdec281e5b0245809a4477f2135ad92c533 that will at least clear the queue when fresh messages are added. But the queue remains in memory otherwise, potentially causing memory issues. Shouldn't we remove relevant items from the queue after we received ack or nack from server about them?

> So here, if send() succeeds, we will keep the QueuedMessage in the queue, and when the server responds we can call the listener. But in the exception case when send() fails we have to decide what to do. Why did it fail? Is it because the connection dropped again? Or is there some other problem?

The current behaviour for errors for queued messages for completion listener seem to be inside `PendingQueuedMessages` and connection emits an update with an error if there is an exception with the send. I think we should be removing that relevant message from pending queue if that happens but currently there is no such a mechanism on `PendingMessageQueue` and I cannot see any spec points that describes this case. What would you think should happen in that case?

> If we decide that the failure to send() means that we're going to give up on sending the message, then here we would remove the QueuedMessage from the queue, and call the listener with an error. However, I'm not aware of any circumstances where the send() would fail, other than connectivity problems (in which case we'd expect there to be a subsequent resume that triggered another retry attempt).

As above this should trigger a connection event back to the user, but I'm really not sure what to do in that case. But going through codebase where the message is added into the pending queue, this seems to be happening before transport has sent the message (not even waiting for success of message) I think this will cause another issue where a failed sent message would also be added to the pending queue. Happy to look into this further

ikbalkaya:
> I suggest please that we tidy this and remove the whitespace-related commits, because the commit history is now a mess.

Sorry about this, this should be clearer now - I squashed commits that were causing that

> And we still need tests pls.

I have not finished addressing everything yet sorry. I should request for another review when I'm done

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1232,12 +1230,43 @@ private synchronized void onConnected(ProtocolMessage message) {
...
+        }
+        final List<QueuedMessage> messagesFromPendingQueue = CollectionUtils.deepCopyOfQueuedMessages(pendingMessages.queue);
+        //Clear the pendingQueue now, because we do not want the retried messages to accumulate on it.
+        pendingMessages.clearQueue();
+
+        //put messages from pending messages to front of queuedMessages, the ones with the message serials will already
+        //have been assigned new message serial to them at this point
+        queuedMessages.addAll(0,messagesFromPendingQueue);
+        // reassign new serials for remaining queued messages if reset was required
+        if (resetMessageSerial) {
```
paddybyers, line 1258: Why reset the `msgSerial` as two separate operations (ie once for the contents of `pendingMessages`, and once for the original contents of `queuedMessages`. Is it simpler just to add the messages from `pendingMessages`, and then set the new `msgSerial` for all of them as a single operation?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1183,44 +1185,41 @@ private void onChannelMessage(ProtocolMessage message) {
...
-             * there). */
-            pendingMessages.reset(msgSerial,
-                    new ErrorInfo("Connection resume failed", 500, 50000));
-            msgSerial = 0;
-        }
         connection.id = message.connectionId;
-        if(message.connectionSerial != null) {
-            connection.serial = message.connectionSerial.longValue();
-            if (connection.key != null)
-                connection.recoveryKey = connection.key + ":" + message.connectionSerial;
```
paddybyers, line 1220: Why was this removed? Where else is `recoveryKey` set?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1232,12 +1231,40 @@ private synchronized void onConnected(ProtocolMessage message) {
...
+        queuedMessages.addAll(0, pendingMessages.queue);
+        //Clear the pendingMessages queue now, because we do not want the retried messages to accumulate on it.
+        pendingMessages.clearQueue();
+
+        //RTN19a
+        if (resetMessageSerial){
+            pendingMessages.resetStartSerial(0);
+            // reassign new serials for queued messages - we can set them to their indexes since they
+            // will also start with 0
+            for (int i = 0; i < queuedMessages.size(); i++) {
```
paddybyers, line 1256: since the `msgSerial` is set later in `sendImpl()`, I don't think it's necessary to set it here.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1718,6 +1745,13 @@ public synchronized void reset(long oldMsgSerial, ErrorInfo err) {
             startSerial = 0;
         }
 
+        public void resetStartSerial(int from) {
```
paddybyers, line 1748: Is this called now?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
         }
     }
 
+    /**
+     * In case of resume success, verify that pending messages are resent. By blocking ack/nacks before sending the
+     * message while connected and then disconnect, add some mroe messages
```
paddybyers, line 699: mroe typo

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+            received.remove(0);
+            received.remove(1);
+            received.remove(2);
+            assertEquals(
+                "Did not receive the entire second round of messages (queued)",
+                received.size(), messageCount
+            );
+
+            //make sure that pending queue is cleared
+            assertEquals("Thre are still pending messages in the queue",
```
paddybyers, line 788: Thre typo

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+            assertEquals(
+                "The receiver's channel should be attached",
+                receiverChannel.state, ChannelState.attached
+            );
+            /* subscribe */
+            MessageWaiter messageWaiter =  new MessageWaiter(receiverChannel);
+            //block acks nacks
+            mockWebsocketFactory.blockReceive(message -> message.action == ProtocolMessage.Action.ack ||
+                message.action == ProtocolMessage.Action.nack);
+            final int messageCount = 6; // make 3 pending 3 queued
```
paddybyers, line 737: Personally I think I would have had `messageCountPerPhase = 3`

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+                received.size(), messageCount
+            );
+
+            //make sure that pending queue is cleared
+            assertEquals("Thre are still pending messages in the queue",
+                sender.connection.connectionManager.getPendingMessages().size(),
+                0);
+
+        } catch (AblyException e) {
+            fail("init0: Unexpected exception instantiating library");
```
paddybyers, line 793: I know that this particular error message is in lots of other tests, but it's actually a copy/paste error. It could be something more meaningful for this test.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+            MessageWaiter messageWaiter =  new MessageWaiter(receiverChannel);
+            //block acks nacks
+            mockWebsocketFactory.blockReceive(message -> message.action == ProtocolMessage.Action.ack ||
+                message.action == ProtocolMessage.Action.nack);
+            final int messageCount = 6; // make 3 pending 3 queued
+            CompletionSet senderCompletion = new CompletionSet();
+            for (int i = 0; i < messageCount/2; i++) {
+                senderChannel.publish("pending_queued_message_" + i, "Test pending queued messages " + i,
+                    senderCompletion.add());
+                try { Thread.sleep(delay); } catch(InterruptedException e){}
```
paddybyers, line 742: Do we have to wait here? A wait like this is just going to make the test flaky. Can't we synchronise on something?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+            assertEquals(
+                "Did not receive the entire second round of messages (queued)",
+                received.size(), messageCount
+            );
+
+            //make sure that pending queue is cleared
+            assertEquals("Thre are still pending messages in the queue",
+                sender.connection.connectionManager.getPendingMessages().size(),
+                0);
+
```
paddybyers, line 791: Is there a way to verify that the messages were sent with the expected `msgSerial`s ?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -684,6 +694,237 @@ public void resume_publish_queue() {
...
+            List<Message> received = messageWaiter.receivedMessages;
+            //we need to discard the first 3 - that were sent before disconnection
+            received.remove(0);
+            received.remove(1);
+            received.remove(2);
+            assertEquals(
+                "Received messages",
+                messageCount, received.size()
+            );
```
paddybyers, line 924: Again, can we have an assertion that the `msgSerial`s are set correctly?

paddybyers (changes requested):
Please see inline comments

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: Yes - this should be called from `addPendingMessagesToQueuedMessages`

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: I have added ability to note the sent messages on mock transport and improved tests e44c840990e98f860220c5b2c89cc5f394dc4a59

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: This was removed by mistake - as part of changes from protocol-2.0. Introduced it back bc2cfef966abec1abaf02d8037e637fd51527484

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: d8c21269f39b89b6ec294d39752b4d5b61765ab7

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: 0b1a358357f232a12f730dd10cbb0ee7e9df25b9

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: In most recent changes, I removed this and just explicitly set the constant 3 for each send as I thought this was a bit confusing. I hope that's ok.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: 54eae132f07d5db417ed3760c08db39f819cefb0

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: Removed all waits here 988cce9ab3c41e93f0d4c7f7243022b290e5c808

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: That should be the case in newly improved tests as above

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
ikbalkaya: This should be fixed as part of e44c840990e98f860220c5b2c89cc5f394dc4a59

paddybyers (commented):
This is looking good.

Can you please tidy the commit history before merging this please?

paddybyers (approved):
LGTM, thanks

jaley (approved):
I've not looked at the code but see you already had two others do that. Instead, I checked out this branch, built it and integrated it with AAT to rerun the `NetworkingConnectivityTest`. I can confirm that this fixes the following fault simulations:

* `DisconnectWithFailedResume`
* `EnterUnresponsive`

For those two, it'd be good to re-enable those tests when you integrate this release to AAT, to confirm they're green on `main` too.

It also fixes the problems initially causing `ReenterOnResumeFailed` to fail, leaving us now with just a non-fatal `NACK` causing an exception to be thrown - a separate known issue, so we can't re-enable that test until the fix comes it.

Nice work!

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1232,12 +1230,43 @@ private synchronized void onConnected(ProtocolMessage message) {
...
+     * @param resetMessageSerial whether to reset message serial, this will determine whether to reset message serials
+     * on pending queue, for example when a connection resume failed
+     */
+    private void addPendingMessagesToQueuedMessages(boolean resetMessageSerial) {
+        //RTN19a
+        if (resetMessageSerial){
+            pendingMessages.resetStartSerial(0);
+            msgSerial = 0; //msgSerial will increase in sendImpl when messages are sent
+        }
+        final List<QueuedMessage> messagesFromPendingQueue = CollectionUtils.deepCopyOfQueuedMessages(pendingMessages.queue);
```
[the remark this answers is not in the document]
ikbalkaya: Resolving this on behalf of Paddy as addressed

Issue #905, opened 2023-01-26
A failed resume incorrectly retries queued messages prior to reattachment


paddybyers:
When handling a failed resume in accordance with RTN15c3 (https://github.com/ably/specification/blob/ee511f8a62e3e09cdd2ce50fadddb2d66f2ebaa3/textile/features.textile#RTN15c3), the `ConnectionManager` processes outstanding queued messages via `ConnectionManager.queuedMessages (https://github.com/ably/ably-java/blob/07ad898dcfc733fb0cc2a0fc6c9ca56aec93eed3/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1215).

However, these messages instead need to be requeued at the Channel level, so that they are correctly processed relative to the re-attach that happens. (Currently, replayed presence messages are reattempted before the channel attach has completed.)

Normal Messages need to be removed from `ConnectionManager.queuedMessages` and `ConnectionManager.pendingMessages` and put back into `ChannelBase.queuedMessages` (https://github.com/ably/ably-java/blob/e57183f8801615a2a9ca916f745928f49f192cba/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L1270)

Presence messages need to be removed from `ConnectionManager.queuedMessages` and `ConnectionManager.pendingMessages` and put back into `Presence.pendingPresence` (https://github.com/ably/ably-java/blob/e57183f8801615a2a9ca916f745928f49f192cba/lib/src/main/java/io/ably/lib/realtime/Presence.java#L850)

Issue #904, opened 2023-01-26
Pending messages are not failed when transitioning to suspended


paddybyers:
A state transition in `ConnectionManager` that is to `suspended` (or any other state with `queueMessages == false`) will cause unpublished queued messages to fail (see https://github.com/ably/ably-java/blob/07ad898dcfc733fb0cc2a0fc6c9ca56aec93eed3/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L170).

However, this should also fail `pendingMessages`.

PR #906, opened 2023-01-26
Improve after resume failure logic

ikbalkaya:
This PR fixes two issues

* When `queueEvents` is false, pending messages will also be cleared which was missing before. 
* Fixes the issue of reattachment when a connection resume failed.

Previously as soon as connected message was received by connection manager, channels would directly attempt to attach. I have changed the behaviour so that after a resume failure (connected event with a new connection id) I transfer all queued messages (if any) from the connection manager to associated channel. This sets a flag for later to be processed from `setConnected`. Previously when attempted to call directly to attach from reattach, that would also fail and retry would happen quite a bit later. Attach would time out and then a reattach with some delay would be invoked. I also removed a sync() call from `setConnected()` as recommended by @paddybyers . 

There was also an issue with previously enters members wouldn't automatically reenter. There should be a fix for this and a test that confirms the behaviour.

There are also some test and test utilities improvements.

Fixes https://github.com/ably/ably-java/issues/905
Fixes https://github.com/ably/ably-java/issues/904

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -256,9 +256,9 @@ public class AblyRealtime extends AblyRest {
                     Log.d(TAG, "reAttach(); channel = " + channel.name);
 
                     if (channelQueueMap.containsKey(channel.name)){
-                        channel.transferQueuedMessages(channelQueueMap.get(channel.name));
+                        channel.transferQueuedPresenceMessages(channelQueueMap.get(channel.name));
                     }else {
-                        channel.transferQueuedMessages(null);
+                        channel.transferQueuedPresenceMessages(null);
                     }
                 }
             }
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -257,6 +257,8 @@ public class AblyRealtime extends AblyRest {
 
                     if (channelQueueMap.containsKey(channel.name)){
                         channel.reattach(channelQueueMap.get(channel.name));
+                    }else {
+                        channel.reattach(null);
                     }
                 }
             }
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -1,6 +1,9 @@
 package io.ably.lib.realtime;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 import io.ably.lib.rest.AblyRest;
@@ -233,14 +236,28 @@ public class AblyRealtime extends AblyRest {
 
         /**
          * By spec RTN15c3
+         * Move queued messages from connection manager to their respective channel and reattach
+         * @param queuedMessages Queued messages transferred from connection
          */
         @Override
-        public void reattachOnResumeFailure() {
+        public void reattach(List<ConnectionManager.QueuedMessage> queuedMessages) {
+            final Map<String, List<ConnectionManager.QueuedMessage>> channelQueueMap  = new HashMap<>();
+            for (ConnectionManager.QueuedMessage queuedMessage : queuedMessages) {
+                final String channelName = queuedMessage.msg.channel;
+                if (!channelQueueMap.containsKey(channelName)){
+                    channelQueueMap.put(channelName, new ArrayList<>());
+                }
+                channelQueueMap.get(channelName).add(queuedMessage);
+            }
+
             for (Map.Entry<String, Channel> channelEntry : map.entrySet()) {
                 Channel channel = channelEntry.getValue();
                 if (channel.state == ChannelState.attaching || channel.state == ChannelState.attached || channel.state == ChannelState.suspended) {
                     Log.d(TAG, "reAttach(); channel = " + channel.name);
-                    channel.attach(true, null);
+
+                    if (channelQueueMap.containsKey(channel.name)){
+                        channel.reattach(channelQueueMap.get(channel.name));
+                    }
                 }
             }
         }
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -236,11 +236,11 @@ public class AblyRealtime extends AblyRest {
 
         /**
          * By spec RTN15c3
-         * Move queued messages from connection manager to their respective channel and reattach
-         * @param queuedMessages Queued messages transferred from connection
+         * Move queued messages from connection manager to their respective channel for them to be sent after reattach
+         * @param queuedMessages Queued messages transferred from ConnectionManager
          */
         @Override
-        public void reattach(List<ConnectionManager.QueuedMessage> queuedMessages) {
+        public void transferToChannels(List<ConnectionManager.QueuedMessage> queuedMessages) {
             final Map<String, List<ConnectionManager.QueuedMessage>> channelQueueMap  = new HashMap<>();
             for (ConnectionManager.QueuedMessage queuedMessage : queuedMessages) {
                 final String channelName = queuedMessage.msg.channel;
@@ -256,9 +256,9 @@ public class AblyRealtime extends AblyRest {
                     Log.d(TAG, "reAttach(); channel = " + channel.name);
 
                     if (channelQueueMap.containsKey(channel.name)){
-                        channel.reattach(channelQueueMap.get(channel.name));
+                        channel.transferQueuedMessages(channelQueueMap.get(channel.name));
                     }else {
-                        channel.reattach(null);
+                        channel.transferQueuedMessages(null);
                     }
                 }
             }
--- a/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
+++ b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
@@ -199,24 +199,26 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
      * */
     void reattach(List<QueuedMessage> messagesToTransfer) {
         state = ChannelState.attaching;
-        for (QueuedMessage queuedMessage : messagesToTransfer) {
-            if (queuedMessage.msg.action == Action.message) {
+        if (messagesToTransfer != null) {
+            for (QueuedMessage queuedMessage : messagesToTransfer) {
+                if (queuedMessage.msg.action == Action.message) {
 
-                queuedMessages.add(queuedMessage);
-            }else if (queuedMessage.msg.action == Action.presence) {
-                PresenceMessage[] presenceMessages = queuedMessage.msg.presence;
-                if (presenceMessages != null && presenceMessages.length > 0){
-                    for (PresenceMessage presenceMessage : presenceMessages) {
-                        String clientId;
-                        try {
-                            clientId = ably.auth.checkClientId(presenceMessage, false, true);
-                        } catch(AblyException e) {
-                            if(queuedMessage.listener != null) {
-                                queuedMessage.listener.onError(e.errorInfo);
+                    queuedMessages.add(queuedMessage);
+                }else if (queuedMessage.msg.action == Action.presence) {
+                    PresenceMessage[] presenceMessages = queuedMessage.msg.presence;
+                    if (presenceMessages != null && presenceMessages.length > 0){
+                        for (PresenceMessage presenceMessage : presenceMessages) {
+                            String clientId;
+                            try {
+                                clientId = ably.auth.checkClientId(presenceMessage, false, true);
+                            } catch(AblyException e) {
+                                if(queuedMessage.listener != null) {
+                                    queuedMessage.listener.onError(e.errorInfo);
+                                }
+                                return;
                             }
-                            return;
+                            this.presence.addPendingPresence(clientId, presenceMessage, queuedMessage.listener);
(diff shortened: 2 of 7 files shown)


lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -186,6 +186,36 @@ void attach(boolean forceReattach, CompletionListener listener) {
...
+        for (QueuedMessage queuedMessage : messagesToTransfer) {
+            if (queuedMessage.msg.action == Action.message) {
+                queuedMessages.add(queuedMessage);
+            }else if (queuedMessage.msg.action == Action.presence) {
+                PresenceMessage[] presenceMessages = queuedMessage.msg.presence;
+                if (presenceMessages != null && presenceMessages.length > 0){
+                    for (PresenceMessage presenceMessage : presenceMessages) {
+                        String clientId;
+                        try {
+                            clientId = ably.auth.checkClientId(presenceMessage, false, true);
```
ikbalkaya, line 203: I took this part from an existing code which puts pending presence.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -213,13 +214,13 @@ void reattach(List<QueuedMessage> messagesToTransfer) {
 
             }
         }
-        attach(true,null);
```
paddybyers: why remove this?

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -236,10 +236,11 @@ public void suspendAll(ErrorInfo error, boolean notifyStateChange) {
 
         /**
          * By spec RTN15c3
+         * Move queued messages from connection manager to their respective channel
          * @param queuedMessages Queued messages transferred from connection
          */
         @Override
-        public void reattachOnResumeFailure(List<ConnectionManager.QueuedMessage> queuedMessages) {
+        public void transferQueuedMessages(List<ConnectionManager.QueuedMessage> queuedMessages) {
```
paddybyers, line 243: As written, this is very inefficient. We're iterating through all queued messages multiple times, once for each channel.

ikbalkaya: aa1bc86296f22e21386630394f75f81ae22675c0

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: I think when running tests I had a wrong impression that this reattach will happen as a result of another state transition as I was seeing channels reattached. but now I see they are stuck in attaching state. It's a result of writing things in hurry without having time to check their effect. I'll reintroduce this.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -186,6 +193,40 @@ void attach(boolean forceReattach, CompletionListener listener) {
...
+    /**
+     * This method carries queued messages accumulated on connection manager while the channel
+     * isn't attached yet. It's added in the queue here and start a new attach call
+     * */
+    void reattach(List<QueuedMessage> messagesToTransfer) {
+        state = ChannelState.attaching;
+        if (messagesToTransfer != null) {
+            for (QueuedMessage queuedMessage : messagesToTransfer) {
+                if (queuedMessage.msg.action == Action.message) {
+
```
paddybyers, line 205: whitespace

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -186,6 +193,40 @@ void attach(boolean forceReattach, CompletionListener listener) {
...
+                    queuedMessages.add(queuedMessage);
+                }else if (queuedMessage.msg.action == Action.presence) {
```
paddybyers, line 207: please no `}else` - make whitespace consistent

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -186,6 +193,40 @@ void attach(boolean forceReattach, CompletionListener listener) {
...
+                    PresenceMessage[] presenceMessages = queuedMessage.msg.presence;
+                    if (presenceMessages != null && presenceMessages.length > 0){
+                        for (PresenceMessage presenceMessage : presenceMessages) {
+                            String clientId;
+                            try {
+                                clientId = ably.auth.checkClientId(presenceMessage, false, true);
```
paddybyers, line 213: why is this being checked here? If the message had previously been queued, don't we know that the `clientId` is valid?

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -84,6 +85,12 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
 
     private int retryCount = 0;
 
+    /**
+     * This flag is determining whether the attach should be reattampted when a setConnected() event us receuved
+     * This is a workaround and the mechanism to handle this should be rethought in the future
+     * */
+    private final AtomicBoolean reattachOnConnectionResume = new AtomicBoolean(false);
```
paddybyers, line 92: Can you explain what contention this is addressing?

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -406,6 +447,7 @@ synchronized private void clearAttachTimers() {
                 t.cancel();
                 t.purge();
             }
+
```
paddybyers, line 450: why a blank line?

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -624,7 +663,7 @@ public synchronized void setSuspended(ErrorInfo reason, boolean notifyStateChang
             Log.v(TAG, "setSuspended(); channel = " + name);
             presence.setSuspended(reason);
             setState(ChannelState.suspended, reason, false, notifyStateChange);
-            failQueuedMessages(reason);
+            // failQueuedMessages(reason);
```
paddybyers, line 666: why is this removed?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1252,19 +1255,16 @@ private synchronized void onConnected(ProtocolMessage message) {
     private void addPendingMessagesToQueuedMessages(boolean resetMessageSerial) {
         // Add messages from pending messages to front of queuedMessages in order to retry them
         queuedMessages.addAll(0, pendingMessages.queue);
-        //rewind start serial back to the first serial since we are clearing the queue
-        if (!pendingMessages.queue.isEmpty()){
-            //Reset current serial to the first pending message on previous queue as we are going to clear the queue now
+        //rewind start serial back to the first serial since we are clearing the queue if the queue is not empty
```
paddybyers, line 1258: I think this would be a lot clearer if it was:

```
if (resetMessageSerial){  // failed resume, so all new published messages start with msgSerial = 0
  msgSerial = 0; //msgSerial will increase in sendImpl when messages are sent
  pendingMessages.resetStartSerial(0);
} else if(!pendingMessages.queue.isEmpty()) { // pendingMessages needs to expect next msgSerial to be the earliest previously unacknowledged message
  msgSerial =  pendingMessages.queue.get(0).msg.msgSerial;
  pendingMessages.resetStartSerial((int) (msgSerial));
}

```

lib/src/test/java/io/ably/lib/test/realtime/RealtimeResumeTest.java
```diff
@@ -909,7 +924,14 @@ public void onConnectionStateChanged(ConnectionStateChange state) {
             assertNotNull(ably.connection.id);
             assertNotEquals("Connection has the same id", firstConnectionId, ably.connection.id);
 
-            /* wait for the publish callback to be called.*/
+            // wait for channel to get attached
+            (new ChannelWaiter(senderChannel)).waitFor(ChannelState.attached);
+            assertEquals("Connection has the same id", ChannelState.attached, senderChannel.state);
+
+            try {
```
paddybyers, line 931: why do we need this wait?

paddybyers (commented):
See inline comments

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: `setConnected()` and `reattach()` calls are done from different threads. setConnected() is called when a state transition happens - so it technically will always be done after `setConnected` , but as they arrive from different threads I wanted to minimize any sync issues by using an atomic boolean

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: c4f34de7056b6c90eeaf8a81a36bdb586bb6b157

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: c4f34de7056b6c90eeaf8a81a36bdb586bb6b157

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: c4f34de7056b6c90eeaf8a81a36bdb586bb6b157

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
paddybyers: which threads? What triggers `reattach()` other than connection state changes?

My concern is that we're mixing synchronisation primitives; most operations involving connection state are performed by the `ConnectionManager` thread, and there is a lock there that protects the `ConnectionManager` state and ensures that all state transitions occur strictly serially.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: 57c9a142260dd198c40f054787a7aea6cb0d9542

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: Here is a quick run to check in which thread this value is set and read.
```
2023-01-30 11:11:23.844  8588-8629  System.out              io.ably.lib.test                     I  thread_test reattachOnConnectionResume is set :WebSocketConnectReadThread-341
2023-01-30 11:11:23.845  8588-8617  System.out              io.ably.lib.test                     I  thread_test reattachOnConnectionResume is read :Thread-2
```
I think I misrenamed `reattach` as it isn't technically a reattach and will just set this flag so that this could be used when a state transition happens. I'm equally happy to make this non-atomic as I think operations still be ordered as you mentioned above.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: Sorry - this was a result of me furiously commenting out and trying different things and forgetting when cleaning up the code. I believe I commented this out when the channels would go into suspended state as attach wouldn't happen. I put it back f525dcac9fcaef94d1b7893860adb125d9eb307f

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: I took this from an existing code 
https://github.com/ably/ably-java/blob/1c41924f08dffb1e94adc4ee730584b14f16103e/lib/src/main/java/io/ably/lib/realtime/Presence.java#L743
As I did not fully know the condition of clientId being valid here, I thought this check might be sensible. But I can understand why it's not needed here.
Change: c41824dc911210265ea8f3b55c998206f74ce953

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1201,18 +1215,22 @@ private synchronized void onConnected(ProtocolMessage message) {
...
-                channels.reattachOnResumeFailure();
-                // Add any messages still pending from the previous transport (RTN19a) to the front of queued messages
-                // however, this time the pending messages have to have newly assigned `
-                // msgSerial`s. They can't simply be replayed, as they are in the successful resume case
+                //we are going to add pending messages and update pending queue state
                 addPendingMessagesToQueuedMessages(true);
+
+                //We are going to transfer those messages to channel level so that they are published after
+                //their respective channels are attached
+                channels.transferToChannels(new ArrayList<>(queuedMessages));
```
paddybyers, line 1231: I think we're going to have to change how we process unacknowledged `Message`s (as distinct from `PresenceMessage`s. According to the spec (https://sdk.ably.com/builds/ably/specification/main/features/#RTL6c2) all queued `Message`s (ie those that were queued to be sent, but not yet sent, in `ConnectionManager.queuedMessages`; plus those that were sent but are unacknowledged, in `ConnectionManager.pendingMessages`) should be re-sent immediately on establishment of the new connection.

paddybyers (changes requested):

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ikbalkaya: As discussed this is no longer the case as this information will be carried out with `StateIndication` c63183dbf8cf68ad1d5d0b60508205779f2defba

paddybyers (approved):
LGTM, thanks

PR #909, opened 2023-02-01
908 presence message superseded

### LINKED ISSUE #908: Presence messages superseded whilst channel in attaching state
presence.enter and presence.update (and the others) call presence.updatePresence.

updatePresence will put messages on a queue to be sent later, in the event that the channel is currently attaching. This queue is a HashMap that is keyed by clientid. This means if a presence.enter and presence.update occur in quick succession before the channel attaches, the former is cancelled out by the latter (and any listeners are not called).

There are two possible options:

1. Queue up all presence messages and then send in-order on channel attach.
2. Only send the latest message, but call all event listeners with the result.

The spec needs updating to provide more clarification on what the behaviour should be (see: https://github.com/ably/specification/issues/130), but the current implementation in ably-js is to just send all of the pending messages.

AndyTWF:
This change:

- Changes Presence.pendingPresence to a List from a HashMap
- When waiting for channel to attach, if multiple presence updates are received, these are all queued up for sending once attach completes (resulting in all completion listeners being called). This is a change from before, whereby only the most recent presence message for each client id was sent.
- The above behaviour is consistent with ably-js, though not specifically mandated in the specification

Fixes #908

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
@@ -1649,28 +1649,6 @@ public class RealtimePresenceTest extends ParameterizedTest {
         }
     }
 
-    private final class CountingCompletionListener implements CompletionListener
-    {
-        public int successfulListeners = 0;
-        public int failedListeners = 0;
-
-        @Override
-        public void onSuccess() {
-            synchronized(this) {
-                successfulListeners++;
-                notifyAll();
-            }
-        }
-
-        @Override
-        public void onError(ErrorInfo reason) {
-            synchronized(this) {
-                failedListeners++;
-                notifyAll();
-            }
-        }
-    }
-
     /**
      * <p>
      * Validates a client sending multiple presence updates when the channel is in the attaching
@@ -1704,7 +1682,7 @@ public class RealtimePresenceTest extends ParameterizedTest {
             channel2.attach();
             (new ChannelWaiter(channel2)).waitFor(ChannelState.attached);
 
-            CountingCompletionListener messageCompletionListener = new CountingCompletionListener();
+            CompletionWaiter messageCompletionListener = new CompletionWaiter();
 
             final ArrayList<PresenceMessage> receivedMessageStack = new ArrayList<>();
             channel2.presence.subscribe(actions, new Presence.PresenceListener() {
@@ -1746,24 +1724,9 @@ public class RealtimePresenceTest extends ParameterizedTest {
              * Validate that
              * - our listeners are called within 10 seconds
              */
-            long waitUntil = System.currentTimeMillis() + 10000;
-            try {
-                while (true) {
-                    synchronized (messageCompletionListener) {
-                        assertEquals(0, messageCompletionListener.failedListeners);
+            messageCompletionListener.waitFor(3, 10000);
+            assertTrue(messageCompletionListener.success);
 
-                        if (messageCompletionListener.successfulListeners != 3) {
-                            messageCompletionListener.wait(500);
-                            assertTrue(System.currentTimeMillis() < waitUntil);
-                            continue;
-                        }
-                    }
-
-                    break;
-                }
-            } catch (InterruptedException exception) {
-                fail();
-            }
         } finally {
             if (ably1 != null) ably1.close();
             if (ably2 != null) ably2.close();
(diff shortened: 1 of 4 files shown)


lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1649,6 +1649,125 @@ public void onPresenceMessage(PresenceMessage message) {
         }
     }
 
+    private final class CountingCompletionListener implements CompletionListener
```
paddybyers, line 1652: you could put this in with the helpers

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1649,6 +1649,125 @@ public void onPresenceMessage(PresenceMessage message) {
...
+             * Validate that
+             * - our listeners are called
+             */
+            try {
+                while (true) {
+                    synchronized (messageCompletionListener) {
+                        assertEquals(0, messageCompletionListener.failedListeners);
+
+                        if (messageCompletionListener.successfulListeners != 3) {
+                            messageCompletionListener.wait(5000);
```
paddybyers, line 1755: Isn't this going to loop forever if we don't reach `messageCompletionListener.successfulListeners == 3` ?

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1649,6 +1649,125 @@ public void onPresenceMessage(PresenceMessage message) {
...
+                public void onPresenceMessage(PresenceMessage message) {
+                    synchronized (receivedMessageStack) {
+                        receivedMessageStack.add(message);
+                        receivedMessageStack.notify();
+                    }
+                }
+            });
+
+            /* Start emitting channel with ably client 1 (emitter) */
+            channel1.presence.enter("Hello, #2!", messageCompletionListener);
```
paddybyers, line 1721: Nothing is happening here to ensure that the enter/update calls all happen before the channel is attached. Do you have any thoughts on how to do that?

paddybyers (commented):
See inline comments

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
AndyTWF: Ah, yes - you're right. I mistakenly thought the timeout would throw an exception if the timeout was exceeded. I will add a check.

jaley (approved):
LGTM. Think we could use some input from @paddybyers when he gets a moment though

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
AndyTWF: It turns out there's an existing listener that kinda already does this - I've adapted it to have timeouts and used that :)

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
AndyTWF: Ah I wasn't aware those existed - there's already a listener in there that is essentially the same, but without timeouts. I have adapted it to allow for timeouts and used that!

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
AndyTWF: It's quite a hard one to synchronize, but I believe it can be done by the following:

- Wrap the calls to presence.enter and presence.update in the test with a synchronize(channel1) call.
- setState (called by setAttached) also synchronizes on the channel, so we should prevent the channel from entering the attached state before we've had chance to sort out all the messages.

lib/src/test/java/io/ably/lib/test/realtime/RealtimePresenceTest.java
```diff
@@ -1696,9 +1696,11 @@ public void onPresenceMessage(PresenceMessage message) {
             });
 
             /* Start emitting channel with ably client 1 (emitter) */
-            channel1.presence.enter("Hello, #2!", messageCompletionListener);
-            channel1.presence.update("Lorem ipsum", messageCompletionListener);
-            channel1.presence.update("Dolor sit!", messageCompletionListener);
+            synchronized (channel1) {
```
paddybyers, line 1699: worth a comment here just to explain why you're synchronising on this

paddybyers (approved):
Just one comment but otherwise lgtm
Thanks

ikbalkaya (approved):

PR #914, opened 2023-02-06
Drop messages where channel does not exist

### LINKED ISSUE #913: Released channel re-added to the channel map after DETACHED message
If you call channels.release(), that removes the channel from the channel map and detaches it. However, the protocol messaged for DETACHED returning to us re-adds the channel to the channel map (via a call to channels.get()), therefore undoing the release action.

One way to fix this would be to not re-add the channel when processing DETACHED if its not there.

AndyTWF:
At the moment, if a message is received and a channel does not exist the channel is created (via channels.get) to process the message.

This can lead to issues with channels.release, which also detaches the channel after removing it from the channel map. If the timing is unfortunate, then the DETACH message will come back and re-create the channel, meaning it cannot be recreated afresh later.

This change fixes the issues by dropping any messages that are received for a non existent channel and logs the fact that this has happened.

Fixes #913

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -217,8 +217,12 @@ public class AblyRealtime extends AblyRest {
         @Override
         public void onMessage(ProtocolMessage msg) {
             String channelName = msg.channel;
-            Channel channel;
-            synchronized(this) { channel = channels.get(channelName); }
+            Channel channel = null;
+            synchronized(this) {
+                if (channels.containsKey(channelName)) {
+                    channel = channels.get(channelName);
+                }
+            }
             if(channel == null) {
                 Log.e(TAG, "Received channel message for non-existent channel");
                 return;
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
@@ -2008,34 +2008,31 @@ public class RealtimeChannelTest extends ParameterizedTest {
     }
 
     /*
-     * Without creating a channel, send a DETACHED protocol message to a named channel.
-     *
-     * Assert that the channel is not created when processing the message and that, therefore,
-     * the message is dropped. This prevents issues where releasing a channel - dropping it from
-     * the channel map and calling detach, can cause the channel to be re-created when the
-     * DETATCHED response comes back from ably.
+     * Checks that the DETACHED message sent by the server when a channel is released is dropped.
      */
     @Test
-    public void messages_to_non_existent_channels_are_dropped() throws AblyException {
+    public void detach_message_to_released_channel_is_dropped() throws AblyException {
         AblyRealtime ably = null;
         long oldRealtimeTimeout = Defaults.realtimeRequestTimeout;
-        final String channelName = "messages_to_non_existent_channels_are_dropped";
+        final String channelName = "detach_message_to_released_channel_is_dropped";
 
         try {
-            ClientOptions opts = createOptions(testVars.keys[0].keyStr);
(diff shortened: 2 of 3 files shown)


lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
```diff
@@ -217,8 +217,12 @@ public void release(String channelName) {
         @Override
         public void onMessage(ProtocolMessage msg) {
             String channelName = msg.channel;
-            Channel channel;
-            synchronized(this) { channel = channels.get(channelName); }
+            Channel channel = null;
+            synchronized(this) {
+                if (channels.containsKey(channelName)) {
```
ikbalkaya, line 222: I think that instead of checking the existing map - it would have been clearer and more explicit to check this against DETACHED protocol message.

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -2007,6 +2007,44 @@ public void no_messages_when_channel_state_not_attached() {
...
+    /*
+     * Without creating a channel, send a DETACHED protocol message to a named channel.
+     *
+     * Assert that the channel is not created when processing the message and that, therefore,
+     * the message is dropped. This prevents issues where releasing a channel - dropping it from
+     * the channel map and calling detach, can cause the channel to be re-created when the
+     * DETATCHED response comes back from ably.
+     */
+    @Test
+    public void messages_to_non_existent_channels_are_dropped() throws AblyException {
```
ikbalkaya, line 2019: I think for this case it would also be useful to show that there was an attached channel that was detached - (with call to `detach()`) and then send the `detached()` message afterwards

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
AndyTWF: This was discussed a bit on Slack (https://ably-real-time.slack.com/archives/C01EPJENRD0/p1675680904225679?thread_ts=1675427813.939969&cid=C01EPJENRD0) - this behaviour is also consistent with `ably-js` :)

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
AndyTWF: I'm a little confused as to what you want me to add to the test, or what that test case would be covering? This test aims to cover what should happen if a channel doesn't exist in the channel map, rather than what state the channel is in?

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
ikbalkaya: Ah thanks. Given Paddy approved this I'm happy with the approach. Just to be clear that I was worried that this might be covering much more cases than the original case where 'detached' causes an attach. So I think it would be good to answer this question: Wouldn't be there any other cases where map doesn't contain that channel and a protocol message would necessitate the creation /attachment of a new channel?

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
ikbalkaya: I meant to replay the whole scenario where there is an attach and the map contains a channel at some point and then the detach causing it to disappear - then test the afterwards :)

AndyTWF: Ah got you - I have pushed a test that allows that :)

lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
paddybyers: I don't believe so

paddybyers (approved):
LGTM, thanks

ikbalkaya (approved):

Issue #920, opened 2023-02-23
Provide an error code and error message for failed queued messages


ikbalkaya:
The code below creates an empty error info when a pending message fails. It must also contain a convenient code and message (check spec and other SDKs)

https://github.com/ably/ably-java/blob/94ecb6fada26982aaa698f3291f150f344eb2cd4/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1800

PR #922, opened 2023-02-24
Add reason to pending message instead of creating an ErrorInfo

ikbalkaya:
This PR passes `reason` to `fail()` from `failQueuedMessages` instead of creating an empty `ErrorInfo`.

Fixes https://github.com/ably/ably-java/issues/920

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -1682,7 +1682,7 @@ public class ConnectionManager implements ConnectListener {
             queuedMessages.clear();
 
             //also pending messages
-            pendingMessages.fail();
+            pendingMessages.fail(reason);
         }
     }
 
@@ -1795,9 +1795,9 @@ public class ConnectionManager implements ConnectListener {
         }
 
         //fail all pending queued emssages
-        synchronized void fail() {
+        synchronized void fail(ErrorInfo reason) {
             for (QueuedMessage queuedMessage: queue){
-                queuedMessage.listener.onError(new ErrorInfo());
+                queuedMessage.listener.onError(reason);
             }
             queue.clear();
         }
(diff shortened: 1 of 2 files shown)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1795,9 +1795,9 @@ public void resetStartSerial(int from) {
         }
 
         //fail all pending queued emssages
-        synchronized void fail() {
+        synchronized void fail(ErrorInfo reason) {
             for (QueuedMessage queuedMessage: queue){
-                queuedMessage.listener.onError(new ErrorInfo());
+                queuedMessage.listener.onError(reason);
```
AndyTWF, line 1800: Is this method tested anywhere? Is there a way we can add an assertion in a test somewhere that makes sure that the error message is set so we can avoid a future regression?

ikbalkaya: I had added tests for this so I think there must be. I will check the relevant test(s) and add assertions for this particular case. Thanks for bringing it up 👍

ikbalkaya: In retrospect, I think this reason comes from `stateIndication` of `enact` from `State` and not necessarily directly related to our case. I have seen some tests provoke failure call but these aren't the tests I had added. So even if I add a test for this specific case, it won't cover all cases. I think it will probably make sense to add tests for state transitions as general asserting all states will have error codes. But I feel like this is the material for another issue and PR

ikbalkaya: I ended up adding acae4782bc003b6c9a93182facb1d78b7f222ba0 As this is where tests are observing completion handling.

AndyTWF (approved):

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
ikbalkaya: Resolving this on Andy's behalf as he approved it now.

Issue #926, opened 2023-03-17
equals() for TokenDetails is broken


ikbalkaya:
Following is how `equals` has been implemented for `TokenDetails`

```
 @Override
   public boolean equals(Object obj) {
            TokenDetails details = (TokenDetails)obj;
            return equalNullableStrings(this.token, details.token) &
                    equalNullableStrings(this.capability, details.capability) &
                    equalNullableStrings(this.clientId, details.clientId) &
                    (this.issued == details.issued) &
                    (this.expires == details.expires);
        }    
```
This implementation doesn't take into account when obj is null or not the type of token details.

**Expected**
Return false when obj is null or a different type

Spotted:
Flutter plugin Stream writer does this check when writing types to streams
```
protected void writeValue(@NonNull ByteArrayOutputStream stream, @Nullable Object value) {
    if (value == null || value.equals(null)) {
      stream.write(NULL);
```

PR #927, opened 2023-03-17
Fix equals() on token details 

ikbalkaya:
Fixes https://github.com/ably/ably-java/issues/926

### DIFF
--- a/lib/src/main/java/io/ably/lib/rest/Auth.java
+++ b/lib/src/main/java/io/ably/lib/rest/Auth.java
@@ -358,6 +358,8 @@ public class Auth {
          */
         @Override
         public boolean equals(Object obj) {
+            if (!(obj instanceof TokenDetails)) return false;
+
             TokenDetails details = (TokenDetails)obj;
             return equalNullableStrings(this.token, details.token) &
                     equalNullableStrings(this.capability, details.capability) &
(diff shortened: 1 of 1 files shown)


AndyTWF (commented):
The cast on line 363 would throw a ClassCastException if you tried to compare anything other than a `TokenDetails` or subclass to a `TokenDetails`. Is it better to leave this behaviour in place, as it very quickly shows to anyone working on the library that they're trying to make an illogical comparison?

ikbalkaya:
> The cast on line 363 would throw a ClassCastException if you tried to compare anything other than a `TokenDetails` or subclass to a `TokenDetails`. Is it better to leave this behaviour in place, as it very quickly shows to anyone working on the library that they're trying to make an illogical comparison?

The conventional contract in for `equals` in Java is to check type and return false if type isn't what is expected by the object itself. I do not think raising an exception on `equals` would be appropriate.

AndyTWF (approved):

AndyTWF:
> > The cast on line 363 would throw a ClassCastException if you tried to compare anything other than a `TokenDetails` or subclass to a `TokenDetails`. Is it better to leave this behaviour in place, as it very quickly shows to anyone working on the library that they're trying to make an illogical comparison?
> 
> The conventional contract in for `equals` in Java is to check type and return false if type isn't what is expected by the object itself. I do not think raising an exception on `equals` would be appropriate. In the meantime, I realized that `hashcode` also wasn't implemented for this which I added here [565aed9](https://github.com/ably/ably-java/commit/565aed9d39b94e7f7f789b30ec471e10cfbba77f)

Fair play - I wasn't able to find anything too authoritative on conventions when I looked, but happy to go with convention :)

PR #928, opened 2023-03-17
Suspend timer is set when transport is unavailable and last state was connected

### LINKED ISSUE #925: Long-lived connections are immediately transitioned to `SUSPENDED` after disconnection
Relates to spec:

- `DF1a`
- `RTN14d`
- `RTN15g`

The default `connectionStateTtl` value (how long the connection is persisted on the Ably servers) is 120 seconds. This means that after 120 seconds of being `DISCONNECTED`, the connection state should transition to `SUSPENDED`.

We maintain a timer (`ConnectionManager.suspendTime`) in `ably-java` that represents the time at which suspension should occur. This is checked whenever:

- A `CONNECTING` state times out
- The transport is unavailable

If the time is not exceeded, the state is transitioned to `DISCONNECTED`. If it is exceeded, then the state becomes `SUSPENDED`.

The `suspendTime` is reset (to `now + connetionStateTtl`) only:

- On `ConnectionManager` initialisation
- On successful connection
- When a `DISCONNECTED` state is enacted, the previous state was `CONNECTED` and immediate retries are not suppressed.

If the only thing a connection has done is successfully connect... and transport becomes unavailable (via connection rebalancing, internet issues etc)... and we've been connected for >`connectionStateTtl`... then `ConnectionManager.onTransportUnavailable` will check the timer, see that it has passed, and immediately suspend the connection.

The correct behaviour here is to enter a `DISCONNECTED` state and immediately retry connections, only suspending at `disconnectTime + connectionStateTtl`.

AndyTWF:
Previously, a connection that had been connected for longer than the suspend timer period (set in onConnected) would have immediately transitioned to suspended when the transport became unavailable. This is because the timer is not reset the first time the transport fails.

This change fixes this by resetting the suspend timer when the transport becomes unavailable, if the current connection state is connected.

Also included are some unit tests on the connection manager which force the connection into a desired state, and then test the behaviour.

Fixes #925

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
@@ -20,14 +20,17 @@ import io.ably.lib.test.util.MockWebsocketFactory;
 import io.ably.lib.transport.ConnectionManager;
 import io.ably.lib.transport.Defaults;
 import io.ably.lib.transport.Hosts;
+import io.ably.lib.transport.ITransport;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
+import io.ably.lib.types.ErrorInfo;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
 import org.mockito.Mockito;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -731,4 +734,81 @@ public void onChannelStateChanged(ChannelStateChange stateChange) {
...
+        ClientOptions opts = createOptions(testVars.keys[0].keyStr);
+        try(AblyRealtime ably = new AblyRealtime(opts)) {
+            ConnectionManager connectionManager = ably.connection.connectionManager;
+            connectionManager.connect();
+            new Helpers.ConnectionManagerWaiter(ably.connection.connectionManager).waitFor(ConnectionState.connected);
+
+            // Here, we "fake" being online for 2 minutes - the suspendTime is set by onConnected and the default is 2 minutes
+            Field suspendTimeField = connectionManager.getClass().getDeclaredField("suspendTime");
+            suspendTimeField.setAccessible(true);
+            suspendTimeField.set(connectionManager, System.currentTimeMillis() - 10);
```
ikbalkaya, line 758: Can I ask what's the purpose of this being set to `System.currentTimeMillis() - 10` 🤔

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
```diff
@@ -731,4 +734,81 @@ public void onChannelStateChanged(ChannelStateChange stateChange) {
...
+
+            // We also have to grab the "real" transport to pass the superseded test
+            Field transportField = connectionManager.getClass().getDeclaredField("transport");
+            transportField.setAccessible(true);
+
+            connectionManager.onTransportUnavailable((ITransport) transportField.get(connectionManager), new ErrorInfo());
+            new Helpers.ConnectionManagerWaiter(connectionManager).waitFor(ConnectionState.disconnected);
+
+            assertTrue((long) suspendTimeField.get(connectionManager) >= System.currentTimeMillis());
+
```
ikbalkaya, line 768: Hmm, is that to assert that the suspend time will have reset after disconnection and that was the reason to set suspend time -10 in the first place?

ikbalkaya (approved):
Two questions, but looks good to me

lib/src/test/java/io/ably/lib/test/realtime/ConnectionManagerTest.java
AndyTWF: Yes - this is asserting that `onTransportUnavailable` will reset the `suspendTime` :)

paddybyers (changes requested):
Don't we also need to remove https://github.com/ably/ably-java/blob/a147df8701d5cc8bd13e70081a9b52c97f0f3ed1/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L313 ?

paddybyers (approved):
LGTM, thanks

PR #929, opened 2023-03-24
Release/1.2.27

AndyTWF:
Updates the changelog and other version references for the 1.2.27 release.


paddybyers (approved):
I realise it's self-evident in this case, but I do still think we should stick to our rule of every PR having a description.

Otherwise LGTM

PR #938, opened 2023-05-11
fix(ConnectionManager): don't check state before sending close message

owenpearson:
(probably) resolves an issue where, upon calling connection.close(), the connection state would first transition to `CLOSING` and then the `CLOSE` protocol message would only be sent if the connection state is `CONNECTED` (which will never be the case since we already transitioned to `CLOSING`)


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1553,18 +1553,14 @@ private boolean closeImpl() {
...
-                Log.v(TAG, "Requesting connection close");
-                transport.send(new ProtocolMessage(ProtocolMessage.Action.close));
-                return false;
-            } catch (AblyException e) {
-                /* we're closing, and the attempt to send the CLOSE message failed;
-                 * continue, because we're not going to reinstate the transport
-                 * just to send a CLOSE message */
-            }
+        try {
+            Log.v(TAG, "Requesting connection close");
```
AndyTWF, line 1557: Should we keep the check here, but change it to `connected || closing` - we probably don't want to be sending messages if we're in any other state?

owenpearson: that's a fair point, i did consider doing that - tbh i think it's safer to just send the close message indiscriminately just as long as there's a transport available to send it on. can't think of a situation where we have a transport and we don't want to send a close message when closing

AndyTWF: Fair :)

AndyTWF (approved):

PR #939, opened 2023-05-15
fix: transport not disconnecting after ttl passed

### LINKED ISSUE #932: RTN23a: Transport not disconnecting after TTL passed
Per RTN23a, a transport should be disconnected after receiving no activity for `maxIdleInterval` + `realtimeRequestTimeout`. This is implemented in `checkActivity()`: https://github.com/ably/ably-java/blob/deec71a40affe522c59160b9278334768b928d90/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java#L326

However, the timer that is set cannot do anything because it calls `checkActivity` again, which checks if a timer is set, which it is, so it does nothing (and starts no future timers).

This means on a network drop event, the library will wait over a minute to begin the process of trying fallback hosts etc.

AndyTWF:
Re-organises the activity timer logic to make sure that the timer runs. Previously the timer existence check in checkActivity() would cancel itself out because it would check (once the timer expires) that the timer was not set, when it was. So the disconnect logic would never get to run and a dead transport would linger until the underlying transport threw an exception which could be up to 2 minutes.

This change fixes this by moving the timer checking / timeout logic away from the existence check in checkActivity(). The arrangement of logic is more in line with the implementation in ably-js.

Fixes #932

To test:

- Connect up an Ably client, attach to channel etc
- Wait for a few websocket pings and "ok" logs from "onActivityTimerExpiry".
- Turn off the internet
- After ~15 seconds, the transport should be closed (when the bug was around, this would take > 2 minutes)
- Turn the internet back on
- Connection should resume somewhat swiftly

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -311,31 +311,29 @@ public class WebSocketTransport implements ITransport {
                 Log.v(TAG, "checkActivity: infinite timeout");
                 return;
             }
-            if(activityTimerTask != null) {
-                /* timer already running */
+
+            // Check if timer already running
+            if (activityTimerTask != null) {
                 return;
             }
-            timeout += connectionManager.ably.options.realtimeRequestTimeout;
-            long now = System.currentTimeMillis();
-            long next = lastActivityTime + timeout;
-            if (now < next) {
-                /* We have not reached maxIdleInterval+realtimeRequestTimeout
-                 * of inactivity.  Schedule a new timer for that long after the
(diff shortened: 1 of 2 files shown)


ikbalkaya (approved):
I couldn't test this since I'm currently on move but code looks good to me 👍

owenpearson (changes requested):
Code changes look good! 

I do think this is the kind of behaviour that's easy to break accidentally so I would encourage adding a test fixture if feasible. The test should just intercept the `CONNECTED` message, overwrite the `maxIdleInterval` to be arbitrarily small and then check that the client transitions to `DISCONNECTED` once this interval has elapsed.

PR #941, opened 2023-05-17
flakey tests 

AndyTWF:
**JWT re-auth:**

The test would sometimes fail because we'd end up with two re-auths instead of one due to the token expiring in the middle of the test.

This change fixes the issue, by only caring about the first token regeneration.

**Message data:**

The test was failing as it was setting the log handler and level, which is static and therefore global. This logger is shared between tests that are not running in isolation, so other tests could write to the log, causing the log assertion to fail.

**Pending messages:**

The test uses a sleep to ensure that the TTLs pass and that the connection is marked as stale. However, it does this whilst the channel is still active, and websocket ping/pongs could still be sent, therefore its possible that on the channel reconnecting, a resume will take place.

This change fixes the bug by performing the wait when the channel has been disconnected (with retries suppressed) which guarantees that the connection will be stale when the test reconnects.

**Realtime connect fail:**

The comparison in the test breaks if the number is exactly on the lower bound. This change fixes it by making the check inclusive.


owenpearson (approved):
All changes LGTM, thanks for the write-up 🙂

PR #951, opened 2023-05-31
fix: fallback hosts always being used on transport error

### LINKED ISSUE #950: Connection manager switches to fallback hosts on close
https://github.com/ably/ably-java/blob/bbbcb4d1ee6e9bc7a8b98a2b4990259030e50725/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1393

When the transport becomes unavailable due to a number of reasons (including a clean close, or an abnormal close) the connection immediately jumps to a fallback host if the internet is up.

This is because the check for whether to use fallback includes checking for a `pendingConnect`, as well as the error reason being "server related". The `pendingConnect` is set when the connecting state is enacted, but not cleared apart from `checkFallback` if it's decided not to use a fallback.

This means any time the connection closes for a number of reasons, then the connection will immediately fallback and not resume.

AndyTWF:
Prior to this change, a websocket transport error (whether that's a clean close, abnormal close or some others) would immediately result in a fallback host being used, even if the original host was entirely operational.

This is because the decision on whether to fallback is partly based on the presence of a "pending connection", which is set when the connecting state is entered. This pending connection is never cleared, except for when the decision is not to use a fallback. The result of this, is that a websocket error event (following a successful connection) will always have a lingering pending connection, so if the error from the websocket is "server-related" (the other part of the decision), it will automatically use a fallback host.

This change fixes the issue by clearing the pending connection whenever a connected state is reached. By doing this, it means that a simple disconnection (e.g. from a scaling event) will automatically retry the same host, wheras a disconnection during the connection process will try a fallback.

Fixes #950

To test this:

```
AblyRealtime ably = new AblyRealtime(options);
ably.connection.once(ConnectionEvent.connected, state -> {
    // Once connected, force the wss to drop out with abnormal close

    ConnectionManager connectionManager = ably.connection.connectionManager;
    try {
        Field transportField = connectionManager.getClass().getDeclaredField("transport");
        transportField.setAccessible(true);
        WebSocketTransport transport = (WebSocketTransport) transportField.get(connectionManager);
        transport.close();
    } catch (NoSuchFieldException e) {
        throw new RuntimeException(e);
    } catch (IllegalAccessException e) {
        throw new RuntimeException(e);
    }
});
ably.connect();

```

The above code will connect to Ably, then once connected, force a transport close. On the current `main` this will cause the client to immediately reconnect on a fallback host (failed resume), whereas on this branch, it will connect back (hopefully!) to the original host.

The behaviour implemented is in the spirit of RTN17c, but the spec could do with an update to make it clearer as to what lifecycle conditions in a Realtime connection constitutes the need to use a fallback host (our interpretation is that the spec references RSC17l, which refers to errors in the REST API, so it follows that we'd expect this to apply to the `CONNECTING` state for a Realtime connection.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
+++ b/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
@@ -265,6 +265,12 @@ public class ConnectionManager implements ConnectListener {
         void enactForChannel(StateIndication stateIndication, ConnectionStateChange change, Channel channel) {
             channel.setConnected(stateIndication.reattachOnResumeFailure);
         }
+
+        @Override
+        void enact(StateIndication stateIndication, ConnectionStateChange change) {
+            super.enact(stateIndication, change);
+            pendingConnect = null;
+        }
     }
 
     /**************************************************


lawrence-forooghian:
I'm not sure whether the management of the `pendingConnect` state is actually the main issue here.

> a websocket transport error (whether that's a clean close, abnormal close or some others) would immediately result in a fallback host being used, even if the original host was entirely operational

Might this not also suggest that the [code](https://github.com/ably/ably-java/blob/bbbcb4d1ee6e9bc7a8b98a2b4990259030e50725/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1394) for determining whether a given error should necessitate the use of a fallback host is incorrect? [RTN17f](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17f) lists all such errors, and the WebSocket transport errors that you've listed above don't qualify.

For example, in the example code that you've posted, what is the `ErrorInfo` that finds its way through to `ConnectionManager#checkFallback`?

AndyTWF:
> I'm not sure whether the management of the `pendingConnect` state is actually the main issue here.
> 
> > a websocket transport error (whether that's a clean close, abnormal close or some others) would immediately result in a fallback host being used, even if the original host was entirely operational
> 
> Might this not also suggest that the [code](https://github.com/ably/ably-java/blob/bbbcb4d1ee6e9bc7a8b98a2b4990259030e50725/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1394) for determining whether a given error should necessitate the use of a fallback host is incorrect? [RTN17f](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17f) lists all such errors, and the WebSocket transport errors that you've listed above don't qualify.
> 
> For example, in the example code that you've posted, what is the `ErrorInfo` that finds its way through to `ConnectionManager#checkFallback`?

In the case of abnormal close, the ErrorInfo passed back to the connection manager is `ConnectionManager.REASON_DISCONNECTED` which is the one described in RTN17f. It's determined in https://github.com/ably/ably-java/blob/bbbcb4d1ee6e9bc7a8b98a2b4990259030e50725/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java#L236. It's the same response for a normal close, which is what the code posted above does.

The status code of `ConnectionManager.REASON_DISCONNECTED` is 503, so would trigger the use of a fallback.

I guess there's a discussion as to whether normal websocket closes should count as this - but the REASON_DISCONNECTED potentially triggering a fallback does appear to be inline with RTN17f?

lawrence-forooghian:
> but the REASON_DISCONNECTED potentially triggering a fallback does appear to be inline with RTN17f?

Are you referring to [RTN17f1](https://sdk.ably.com/builds/ably/specification/main/features/#RTN17f1)? If so, my understanding is that it's describing the case where a `DISCONNECTED` `ProtocolMessage` is received from Realtime, not just any error generated by the client library. I think it relates to the broader point about what a `DISCONNECTED` `ProtocolMessage` means, described in [RTN15h](https://sdk.ably.com/builds/ably/specification/main/features/#RTN15h).

RTN17f1 was introduced by @SimonWoolf in https://github.com/ably/specification/commit/7c48827ffc72e8b971be446d6f00317bbcbee3d0 — he may be able to help us parse the meaning of that spec point.

lawrence-forooghian (approved):

PR #955, opened 2023-06-21
Fix incremental backoff jitter

### LINKED ISSUE #954: Fix incremental backoff while reconnecting + refactor tests
- As a part of PR https://github.com/ably/ably-java/pull/852/files we are just calculating backoff timeout and emitting it as an update event.
- Although, this timeout value is not being used internally.
- Newly calculated timeouts need to be fixed internally while reconnecting with the server.

sacOO7:
- Fixed #954


sacOO7:
Failing integration tests are being fixed with #949

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1165,6 +1156,7 @@ public void onMessage(ITransport transport, ProtocolMessage message) throws Ably
                     }
                     break;
                 case connected:
+                    disconnectedRetryAttempt = 0;
```
AndyTWF, line 1159: Curious as to why this has been added here, rather than keeping what we previously had in `setState` - would we not want to base the attempt on the state the library enters, rather than a protocol message received (what happens if the state changes under the hood in the meantime)?

This also looks like a change in behaviour from what was previously here - in the past we would also set the retry attempt to 0 if the channel entered the suspended state?

sacOO7: This is a good point. 
Although I had given a thought on this one and it seems once you receive connected message, state will destined to go into `connected` state and after `suspended` state can never go into `disconnected` state directly. It has to go through connected state first. So, adding check for `suspended` is actually redundant. 

I will see if I can make changes for this.

sacOO7: done

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -301,6 +302,7 @@ StateIndication validateTransition(StateIndication target) {
 
         @Override
         StateIndication onTimeout() {
+            this.timeout = ReconnectionStrategy.getRetryTime(ably.options.disconnectedRetryTimeout, ++disconnectedRetryAttempt);
```
AndyTWF, line 305: I'm not sure if moving this here is going to work as expected? If I'm reading this correctly:

- We transition to disconnected
- We have some number of timeouts (where the state goes disconnected -> connecting -> disconnected etc etc)
- We connect successfully
- We disconnect again... and now the timeout value is the same as it was at the last timeout in `disconnected`, and won't go back to the "correct" value until the first timeout happens again.

Whereas before - this would be handled and the first transition from disconnected to something else and then back again would set the retry delay back to its "first" value.

sacOO7: When we connect again, `disconnectedRetryAttempt` will be set to zero, so it will re-generate the timeout value again.

sacOO7: I think I got your point, till first timeout it will use the old value from last timeout. After that it will start following the right sequence.
But, I don't agree that it used to work before. We had a clear bug that was not able to set timeout value at right place.

sacOO7: I will reproduce this scenario using tests.

AndyTWF: That's correct - but we're not resetting the `timeout` value on the disconnected state when we transition back to `connected`.

The initial value set is in the constructor - and we only call that constructor once, in the ConnectionManager constructor.

The only other time it's set is when the Disconnected state times out (and we calculate the new value based on the number of attempts). For the first time we move back to disconnected again after being connected, the timeout value will be whatever it was the last time we were disconnected?

sacOO7: I have done the fix and pushing it soon.

sacOO7: Fixed.

sacOO7: I will check the scenario where there is forced `instant retry`. It should atleast not increment timeout value at that case

AndyTWF: Will this work if there's a delay in the state being enacted in a SynchronousStateChangeAction from onTransportUnavailable?

In this case:

- setState would be called in the SynchronousStateChangeAction constructor, with the `timeout` still set to whatever it was previously, which would set `currentState`
- The state change action would then be processed by the action loop (and still waiting on the `currentState` timeout)

sacOO7: Currently, the tests are passing, I will try testing this scenario too. 
@AndyTWF thanks for thorough review of the PR

sacOO7: Okay, fixed!

owenpearson (changes requested):
tests look fine to me, i'm not sure about the disconnect retry implementation though:

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -836,8 +836,13 @@ private synchronized ConnectionStateChange setState(ITransport transport, StateI
...
 
-        if (stateIndication.state == ConnectionState.suspended || stateIndication.state == ConnectionState.connected) {
-            this.disconnectedRetryCount = 0;
+        if (stateIndication.state == ConnectionState.connected || stateIndication.state == ConnectionState.suspended) {
+            this.disconnectedRetryAttempt = 0;
+        }
+
+        if (stateIndication.state == ConnectionState.disconnected) {
+            states.get(ConnectionState.disconnected).timeout =
+                ReconnectionStrategy.getRetryTime(ably.options.disconnectedRetryTimeout, ++disconnectedRetryAttempt);
```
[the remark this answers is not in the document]
sacOO7: We were mutating the state map either way, inside the `enact` method itself
https://github.com/ably/ably-java/pull/955/commits/e57f82f105dd2cef5ba946500674a862d27c6df1
As @AndyTWF mentioned, it was causing an edge case where for the first retry, it was using an old timeout value.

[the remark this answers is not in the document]
sacOO7: I am not sure what you meant by changing `ConnectionStateChange.timeout`. 
I would like to point out, the implementation in `java` and `dotnet` is completely different for setting timers from `js` one.
`ConnectionStateChange.timeout` just denotes a timeout value for the changed state, responsible for emitting notifs to external clients, I think it shouldn't be responsible for setting internal timers.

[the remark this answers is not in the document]
sacOO7: Implementation would have been very simple if we go by your approach.
Unfortunately, timers work differently in `ably-java` implementation : ( 
Being familiar with this code, @AndyTWF would agree to this.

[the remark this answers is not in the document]
sacOO7: I will point our where exactly timers are set into wait state and I don't think we can suspend it in between, like we cancel timers in `js` -> https://github.com/ably/ably-java/blob/41b42f42b4ee2dabc51c86438b1783fabfd76723/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L713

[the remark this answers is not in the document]
sacOO7: This too happens inside a `locked` state. Where no other thread can interrupt the flow without obtaining the lock.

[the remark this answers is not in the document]
owenpearson: ok well that's very annoying. i still think we need to refactor this because mutating `ConnectionManager.states` to communicate timeout changes is not the way to go, but i guess it's gonna take a bit of work so it can wait for now.

PR #964, opened 2023-07-13
Error log stack traces

### LINKED ISSUE #963: Stack traces not being sent to error logs
In a few places, we call `Throwable.printStackTrace` rather than sending it to the normal error logs (the former ends up in the system error log). This provides an inconsistent user experience for error handling.

This was reported in `HttpCore.java` but also happens a few places in `Message.java`.

AndyTWF:
Replaces a number of places where exception information is sent to the system error log rather than custom error handlers that users may be using.

Fixes #963

### DIFF
--- a/lib/src/main/java/io/ably/lib/http/HttpCore.java
+++ b/lib/src/main/java/io/ably/lib/http/HttpCore.java
@@ -251,7 +251,6 @@ public class HttpCore {
                 rawHttpListener.onRawHttpResponse(id, method, response);
             }
         } catch(IOException ioe) {
-            ioe.printStackTrace();
             if(rawHttpListener != null) {
                 rawHttpListener.onRawHttpException(id, method, ioe);
             }
--- a/lib/src/main/java/io/ably/lib/http/HttpCore.java
+++ b/lib/src/main/java/io/ably/lib/http/HttpCore.java
@@ -216,12 +216,14 @@ public class HttpCore {
             byte[] body = null;
             if(requestBody != null) {
                 body = prepareRequestBody(requestBody, conn);
+                // Logging level is checked before logging for performance reasons in building the entry
                 if (Log.level <= Log.VERBOSE)
                     Log.v(TAG, System.lineSeparator() + new String(body));
             }
 
             /* log raw request details */
             Map<String, List<String>> requestProperties = conn.getRequestProperties();
+            // Logging level is checked before logging for performance reasons in building the entry
             if (Log.level <= Log.VERBOSE) {
                 Log.v(TAG, "HTTP request: " + conn.getURL() + " " + method);
                 if (credentialsIncluded)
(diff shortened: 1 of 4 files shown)


lib/src/main/java/io/ably/lib/http/HttpCore.java
```diff
@@ -251,7 +251,6 @@ <T> T httpExecute(HttpURLConnection conn, String method, Param[] headers, Reques
                 rawHttpListener.onRawHttpResponse(id, method, response);
             }
         } catch(IOException ioe) {
-            ioe.printStackTrace();
```
sacOO7, line 254: aren't we supposed to use logger here?

sacOO7: atleast send a warning message

sacOO7: Got it 👍

sacOO7:
I can see repeated checks for 
```
if (Log.level <= Log.VERBOSE) {
Log.v(
```
Isn't this supposed to get handled as a part of `Log.v()` impl itself

sacOO7 (changes requested):
@AndyTWF It seems we are using a Singleton logger instance. This can lead to a buggy situation where

1. If the app is using more than one instance of `AblyRealtime` and is providing multiple logHandlers as a part of clientOptions.
2. Last assigned `logHandler` will get be set at https://github.com/ably/ably-java/blob/28ffd6b20fa1adc0ae64adb4cf105b2fb5fda416/lib/src/main/java/io/ably/lib/util/Log.java#L143C6-L143C6.
3. So, all of the logs will get mixed and only be available to the last handler. 
4. Afaik, we don't have singleton loggers in other SDKs to avoid situations like these. 
5. As per spec, `LogLevel` and `LogHandler` are supposed to be passed from `clientOptions` and the same handler should be used to log the errors. There shouldn't be a case for overriding LogLevel/handlers for multiple `AblyRealtime` instances.

AndyTWF:
> @AndyTWF It seems we are using a [Singleton logger instance](https://github.com/ably/ably-java/blob/main/lib/src/main/java/io/ably/lib/util/Log.java). This can lead to a buggy situation where
> 
> 1. If the app is using more than one instance of `AblyRealtime` and is providing multiple logHandlers as a part of clientOptions.
> 2. Last assigned `logHandler` will get be set at https://github.com/ably/ably-java/blob/28ffd6b20fa1adc0ae64adb4cf105b2fb5fda416/lib/src/main/java/io/ably/lib/util/Log.java#L143C6-L143C6.
> 3. So, all of the logs will get mixed and only be available to the last handler.
> 4. Afaik, we don't have singleton loggers in other SDKs to avoid situations like these.
> 5. As per spec, `LogLevel` and `LogHandler` are supposed to be passed from `clientOptions` and the same handler should be used to log the errors. There shouldn't be a case for overriding LogLevel/handlers for multiple `AblyRealtime` instances.

Yes this is something we've been aware of for a while - I'd propose we do this in a separate branch though, as it's not an insignificant amount of work to un-static everything :)

sacOO7:
> 

@AndyTWF maybe this can be done as a part of this PR. It should be a small work to replace the checks, wdyt ?

AndyTWF:
Comments to explain verbose logging done in 5176e80a

sacOO7:
> Comments to explain verbose logging done in [5176e80](https://github.com/ably/ably-java/commit/5176e80af769a04fcb481b1217c99beac3fa5db7)

mm.. the message is not that clear, can we make it like 
```
Verbose LogLevel check avoids extra computation cycles needed to build the logging message.
```

owenpearson (approved):
LGTM 👍

sacOO7 (approved):
LGTM

PR #966, opened 2023-07-19
Message interactions

AndyTWF:
This change implements the "subscription" aspect of the message interactions feature.

JIRA links:

- Epic: https://ably.atlassian.net/browse/CON-6
- Story: https://ably.atlassian.net/browse/CON-2

Spec points:

- RTL22
- MFI1
- MFI2

It contains the following changes:

- Introduction of the "MessageFilter" class as per MFI1/2. This allows the user to create and customise the filters they wish to apply to messages. The implementation is aimed to be idiomatic to how `ClientOptions` is instantiated in the library.
- Introduction of a class (MessageExtrasFilter) that is responsible for applying the above filter settings given a message.
- A manager class for MessageListeners that are "filtered" by the client - for management of listeners separate to the main channel object.
- Addition of additional `subscribe` and `unsubscribe` overloads to ChannelBase to provide the necessary methods for users to use the new filters, as per RTL22.
- README examples of publishing and subscribing to message interactions.

To test:

- Create an instance of `MessageFilter` and specify whatever filters you'd like to use
- Subscribe to a channel with that filter
- Publish messages to the channel that match your filter, they should be received by your subscriber
- Publish messages to the channel that do not match your filter, they should not be received by the connection (ie onMessage will be called), but not passed on to your subscriber.

Filters that can be tested:

- Whether the event "name" matches that of the filter
- Whether the publishing "clientId" matches that of the filter
- Whether (or not) the message references another
- Whether the referenced message timeserial matches the filter
- Whether the reference type matches the filter

JS implementation for reference:

https://github.com/ably/ably-js/pull/1003

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/MessageExtrasFilter.java
+++ b/lib/src/main/java/io/ably/lib/realtime/MessageExtrasFilter.java
@@ -1,0 +1,93 @@
+package io.ably.lib.realtime;
+
+import com.google.gson.JsonObject;
+import io.ably.lib.types.Message;
+
+/**
+ * A filter for inspecting the extras of a message to check if they match.
+ * Spec: RTL22c
+ */
+final class MessageExtrasFilter implements FilteredListeners.IMessageFilter {
+    public final io.ably.lib.types.MessageFilter filter;
+
+    public MessageExtrasFilter(io.ably.lib.types.MessageFilter filter) {
+        this.filter = filter;
+    }
+
+    @Override
+    public boolean onMessage(Message message) {
+        return clientIdMatchesFilter(message) &&
+            nameMatchesFilter(message) &&
+            referenceMatchesFilter(message);
+    }
+
+    private boolean nameMatchesFilter(Message message) {
+        return filter.name == null || message.name.equals(filter.name);
+    }
+
+    private boolean clientIdMatchesFilter(Message message) {
+        return filter.clientId == null || message.clientId.equals(filter.clientId);
+    }
+
+    private boolean referenceMatchesFilter(Message message) {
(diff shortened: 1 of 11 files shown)


AndyTWF:
Popped this back to draft for a second as I'm investigating a potential issue

lib/src/main/java/io/ably/lib/types/MessageFilter.java
```diff
@@ -0,0 +1,57 @@
...
+     * Value to check against `extras.ref.type`.
+     * Spec: MFI2c
+     */
+    public String refType;
+
+    /**
+     * Value to check against the `name` of a message.
+     * Spec: MFI2d
+     */
+    public String name;
```
sacOO7, line 32: Are we also allowing regex for this?

sacOO7: I mean it would make sense to have a powerful filtering capability,

lib/src/main/java/io/ably/lib/realtime/MessageExtrasFilter.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public boolean onMessage(Message message) {
+        return clientIdMatchesFilter(message) &&
+            nameMatchesFilter(message) &&
+            referenceMatchesFilter(message);
+    }
+
+    private boolean nameMatchesFilter(Message message) {
+        return filter.name == null || message.name.equals(filter.name);
```
sacOO7, line 25: I mean it would make sense to have a powerful filtering capability like regex here and other places too : )

PR #968, opened 2023-09-15
fix: do not clear ably client on reset

ttypic:
Looks like it unnecessary to delete ably client on reset, it breaks push logic

### DIFF
--- a/android/src/main/java/io/ably/lib/push/ActivationContext.java
+++ b/android/src/main/java/io/ably/lib/push/ActivationContext.java
@@ -116,8 +116,6 @@ public class ActivationContext {
     public void reset() {
         Log.v(TAG, "reset()");
 
-        ably = null;
-
         getActivationStateMachine().reset();
         activationStateMachine = null;
 


sacOO7:
I would be honest, this `push.reset()` itself feels like breaking the state machine as mentioned here https://github.com/ably/ably-flutter/pull/479#discussion_r1327326981.
Spec mentions proper flow of `activation state machine`. Every stateChange reflects in a flow that can go in different directions.
https://sdk.ably.com/builds/ably/specification/main/features/#RSH3.
I think right solution would be to provide a `stateChange` that is a part of `activation machine` and should resemble [called deactivate](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3b2)

sacOO7:
Though, you can try and check if the hacky solution works!
It shouldn't introduce a state where it disrupts the natural flow of the activation machine though!

ViniciusFSSilva:
Vinícius Felipe dos Santos Silva

sacOO7:
We planning to close this PR right @ttypic

ttypic:
Yes, this PR is just healing the symptom not the original problem

ViniciusFSSilva:
Enviado do meu Galaxy

PR #972, opened 2023-11-20
fix: deviceId and deviceToken consistence

ttypic:
Resolves [SDK-3950](https://ably.atlassian.net/browse/SDK-3950)

According to the logs, in some situations, `deviceId` doesn't agree with `deviceToken`. 

I added additional checks when we setup `deviceToken` to prevent this inconsitency.
Also during `LocalDevice#create` method, where we create `deviceId` I also clear `deviceToken` if it exists.

[SDK-3950]: https://ably.atlassian.net/browse/SDK-3950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/src/androidTest/java/io/ably/lib/test/android/EventTest.java
+++ b/android/src/androidTest/java/io/ably/lib/test/android/EventTest.java
@@ -44,7 +44,7 @@ public class EventTest {
 
     @Test
     public void events_with_constructor_parameter_do_not_have_persisted_name() {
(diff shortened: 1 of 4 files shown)


ttypic:
Probably we can additionally check that only one instance of `LocalDevice` exists, it potentially can create some other problems

AndyTWF (approved):

PR #973, opened 2023-11-21
feat: throw exception when trying to attach on released channel

ttypic:


lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -487,6 +502,10 @@ public void run() {
                 }, Defaults.realtimeRequestTimeout);
     }
 
+    private void checkChannelIsNotReleased() {
+        if (released) throw new IllegalStateException("Can't perform any operation on released channel");
```
AndyTWF, line 506: Should this exception be an instance of `AblyException` with associated `ErrorInfo`?

ttypic: `AblyException` is checked exception unfortunately. I can use "sneakyThrow" workaround to do this, but not sure it will be expected behaviour

AndyTWF: Ah, so this would change the method signature and break BC?

paddybyers: ISTM you could just add a `throws` [here](https://github.com/ably/ably-java/blob/6dd4461a2e5a705408ae00add4fafae96eeded7b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java#L452) and it would work?

AndyTWF (approved):

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
paddybyers: Ok, I think I'd prefer to have the `IllegalStateException` than do the sneaky throw :(

PR #977, opened 2023-11-22
fix: prevent reattaching of detached channels

ttypic:
Resolves [SDK-3959](https://ably.atlassian.net/browse/SDK-3959)

[SDK-3959]: https://ably.atlassian.net/browse/SDK-3959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
+++ b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
@@ -586,7 +586,7 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
     /* State changes provoked by ConnectionManager state changes. */
 
     public void setConnected(boolean reattachOnResumeFailure) {
-        if (reattachOnResumeFailure){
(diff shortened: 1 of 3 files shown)


lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -586,7 +586,7 @@ public void run() {
     /* State changes provoked by ConnectionManager state changes. */
 
     public void setConnected(boolean reattachOnResumeFailure) {
-        if (reattachOnResumeFailure){
+        if (reattachOnResumeFailure && state.isReattachable()){
```
ttypic, line 589: `reattachOnResumeFailure` flag becomes `true` when `connectionId` changes, don't know why we unconditionally attach all channels in this case.

AndyTWF (approved):

Issue #976, opened 2023-11-22
Implement no-connection-serial - ( remove all references to connection serial )


sacOO7:
- Remove all references to connection serial
- Update tests for the same

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4182) by [Unito](https://www.unito.io)

Issue #974, opened 2023-11-22
Implement no-connection-serial - ( recovery key )


sacOO7:
- Refactor creating and passing of recovery key and related spec implementation.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4185) by [Unito](https://www.unito.io)

Issue #846, opened 2022-10-20
Should send `DETACH` after receiving `ATTACHED` while in the `DETACHING` or `DETACHED` state (`RTL5k`)


QuintinWillison:
[RTL5k](https://sdk.ably.com/builds/ably/specification/main/features/#RTL5k) was added in https://github.com/ably/docs/pull/1044, amended in https://github.com/ably/docs/pull/1160 and then only implemented in a single SDK (https://github.com/ably/ably-js/pull/784).

It needs to be implemented in this SDK, as a specification/protocol version 1.2 requirement.

PR #980, opened 2023-11-27
Feature/no connection serial recovery key

sacOO7:
Related #974 
Related #846
Related #976


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -1193,19 +1194,19 @@ public void onMessage(ITransport transport, ProtocolMessage message) throws Ably
...
         boolean reattachOnResumeFailure = false; // this will indicate that channel must reattach when connected
         // event is received
 
+        boolean isConnectionResumeOrRecoverAttempt = !StringUtils.isNullOrEmpty(connection.key) ||
+            !StringUtils.isNullOrEmpty(ably.options.recover);
+        ably.options.recover = null; // RTN16k, explicitly setting null, so it won't be used for subsequent connection requests
+
         connection.reason = error;
         if (connection.id != null) { // there was a previous connection, so this is a resume and RTN15c applies
             Log.d(TAG, "There was a connection resume");
```
ttypic, line 1212: Should we reset `msgSerial` here under `else` branch of `message.connectionId.equals(connection.id)` condition according to RTN16f

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -131,6 +132,11 @@ private void setState(ChannelState newState, ErrorInfo reason, boolean resumed,
             this.retryCount = 0;
         }
 
+        // RTP5a1
+        if (newState == ChannelState.detached || newState == ChannelState.suspended || newState == ChannelState.failed) {
+            properties.channelSerial = null;
```
ttypic, line 137: Maybe I miss understood but [(RTN15c7)](https://sdk.ably.com/builds/ably/specification/main/features/#RTN15c7) states: "The internal msgSerial counter should be reset so that the first message published to Ably will contain a msgSerial of 0". Doesn't this mean reseted `channelSerial` is `0L` and not `null`?

sacOO7: Actually fixed as a part of https://github.com/ably/ably-java/pull/980/commits/b42ff87ad98c55eeefb13d6f372e843fd5511cf2

sacOO7: channelSerial is a string unline msgserial which is int

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: yes https://github.com/ably/ably-java/commit/b42ff87ad98c55eeefb13d6f372e843fd5511cf2

lib/src/main/java/io/ably/lib/types/RecoveryKeyContext.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final String TAG = "RecoveryKeyContext";
+
+    private final String connectionKey;
+    private final long msgSerial;
+    private final Map<String, String> channelSerials = new HashMap<>();
+
+    public RecoveryKeyContext(String connectionKey, long msgSerial, Map<String, String> channelSerials) {
+        this.connectionKey = connectionKey;
+        this.msgSerial = msgSerial;
+        this.channelSerials.putAll(channelSerials);
```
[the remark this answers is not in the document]
sacOO7: General principal of immutability. We have marked channelSerials field as immutable, so it shouldn't get influenced by 
changes in external channelserials and should only capture snapshot of serials at the time of creating the instance.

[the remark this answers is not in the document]
ttypic: Got you! (one more point to rewrite lib in Kotlin) I personally prefer using analog of guava's `ImmutableMap` which just wraps around existing `Map` rather than make `Map`'s copy. But it this particular code probably no harm to copy

ttypic (approved):

PR #982, opened 2023-11-29
no-connection-serial-presence

### LINKED ISSUE #975: Implement no-connection-serial - ( update internal presence )
- Implemented and refactor internal presence as per no connection serial

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4183) by [Unito](https://www.unito.io)

sacOO7:
Fixed #975


lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -354,26 +354,6 @@ private void sendDetachMessage(CompletionListener listener) throws AblyException
         }
     }
 
-    public void sync() throws AblyException {
```
ttypic, line 357: Could you please explain why we deleted this method?

sacOO7: It's unused method : (

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -627,21 +610,9 @@ public void run() {
...
-             * a SUSPENDED channel will initiate an attach operation. If the
-             * attach operation for the channel times out and the channel
-             * returns to the SUSPENDED state (see #RTL4f)
-             */
-            try {
-                attachWithTimeout(null);
-            } catch (AblyException e) {
-                Log.e(TAG, "setConnected(): Unable to initiate attach; channel = " + name, e);
-            }
+    public void setConnected() {
```
ttypic, line 613: Could you please explain why we changed method signature?

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ttypic: But it's public method for the `Channel`, even if it's not mentioned in docs, somebody might've used it, and it will be breaking change

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
sacOO7: Btw, I wanted to add that there was a discussion for `ably-java` where lots of methods are marked public which are not necessarily used.

sacOO7: Because we don't need `reattachOnResumeFailure`, new spec explicitly mentions to attach channels irrespective of resume success/failure.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
sacOO7: Btw, I wanted to add, there was a discussion for `ably-java` where lots of methods are marked public which are not necessarily used.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ttypic: Can we mark old method as deprecated as well? I don't want to introduce breaking changes if we can avoid them

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ttypic: Oh, that's bad. Can we agree to add `@VisibleForTesting` annotation when we find such methods? Also we should prefer package-private methods over public to ease testing, in some rare cases we can access private fields using reflection, but making methods public for testing should be avoided in future.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
sacOO7: As I mentioned => https://github.com/ably/ably-java/pull/982#discussion_r1431570929, this is not being used as public. If someone is using it as a public method, then they are tampering with library internals. I think we should force users not to use this method rather than deprecating it. They should create an explicit issue if they want a new functionality instead of using library internals.

sacOO7: I feel there's a plan as a part of `java-2.0` release, where we refactor code to remove deprecated public methods and mark public methods as private which are not part of the spec.

sacOO7: Also related to comment -> https://ably-real-time.slack.com/archives/CTQELK4NM/p1612514956037700?thread_ts=1612354862.033900&cid=CTQELK4NM

sacOO7: So, it's not just method name, rather implementation can also change for internal methods. So, it's not recommended to use internal methods. I think we should mark those private as such wherever possible.

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
sacOO7: Seems this is an actual method which was in use sometime back. We can it's usage in `ably-js` 
https://github.com/ably/ably-js/blob/fdd5a566eafac0a7fae9592f890361a608794487/src/common/lib/client/realtimechannel.ts#L914C7-L914C7

sacOO7: https://github.com/ably/ably-js/blob/fdd5a566eafac0a7fae9592f890361a608794487/src/common/lib/client/realtimechannel.ts#L914

ttypic (approved):

PR #1005, opened 2024-05-07
replaced java8 compatible code with java7 compatible code

### LINKED ISSUE #1004: v1.2.34 is incompatible with older Android API versions
The issue is related to `getOrDefault` method call which looks like was added in the 1.2.34 ably-java release specifically the Feature/no connection serial in ably-java https://github.com/ably/ably-java/pull/983/files.

It looks like `getOrDefault` needs API level 24 (android 7). Seems this method is supported in java 1.8 onwards. so older version of android using older JRE will not be able to find it.  We should explicitly mention minimum JDK/JRE version required for the ably-java project in both .gradle and README file.

They can either add android support library ( used for backward compatibility to support older APIs ) or we can revert the change from our side.

Quick fix: https://developer.android.com/studio/write/java8-support#library-desugaring

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4793) by [Unito](https://www.unito.io)

sacOO7:
Fixed #1004

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
+++ b/lib/src/main/java/io/ably/lib/realtime/AblyRealtime.java
@@ -266,7 +266,11 @@ public class AblyRealtime extends AblyRest {
             for (Channel channel : map.values()) {
                 if (channel.state.isReattachable()) {
                     Log.d(TAG, "reAttach(); channel = " + channel.name);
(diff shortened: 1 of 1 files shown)


umair-ably (approved):

sacOO7:
Thanks for approval 👍

PR #1008, opened 2024-05-14
chore: update `CHANGELOG.md`

ttypic:
- update `CHANGELOG.md`
- simplify expression for `getOrDefault` workaround

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,9 @@
 
 [Full Changelog](https://github.com/ably/ably-java/compare/v1.2.37...v1.2.38)
 
-**Breaking changes:**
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,12 +4,9 @@
 
 [Full Changelog](https://github.com/ably/ably-java/compare/v1.2.37...v1.2.38)
 
-**Breaking changes:**
```
ttypic, line 7: I think we shouldn't put this fix in the "breaking changes" section, because we fixed this breaking change and treated it as a bug

sacOO7 (approved):
Lgtm

Issue #1015, opened 2024-07-01
 For REST clients, all requests should include an `X-Ably-ClientId` header when basic auth is to be used (RSA7e2)


ttypic:
RSA7e2 is not implemented, `X-Ably-ClientId` is missing for Rest client

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4862) by [Unito](https://www.unito.io)

PR #1019, opened 2024-07-16
feat: include `X-Ably-ClientId` for each request (RSA7e2)

ttypic:
Resolves https://github.com/ably/ably-java/issues/1015

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/rest/HttpHeaderTest.java
+++ b/lib/src/test/java/io/ably/lib/test/rest/HttpHeaderTest.java
@@ -50,7 +50,7 @@ public class HttpHeaderTest extends ParameterizedTest {
      * should be included in all REST requests to the Ably endpoint
      * see {@link io.ably.lib.transport.Defaults#ABLY_AGENT_PARAM}
      * <p>
-     * Spec: RSC7d, G4
+     * Spec: RSC7d, G4, RSA7e2
      * </p>
      */
     @Test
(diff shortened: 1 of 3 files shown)


sacOO7 (approved):
Lgtm
You can check for failing tests though

lib/src/test/java/io/ably/lib/test/rest/HttpHeaderTest.java
```diff
@@ -83,12 +83,50 @@ public void header_lib_channel_publish() {
...
+            /* Publish message */
+            String messageName = "test message";
+            String messageData = String.valueOf(System.currentTimeMillis());
+
+            Channel channel = ably.channels.get("test");
+            channel.publish(messageName, messageData);
+
+            /* Get last headers */
+            Map<String, String> headers = server.getHeaders();
+            Assert.assertEquals(headers.get("x-ably-clientid"), "dGVzdCBjbGllbnQ=");
```
sacOO7, line 123:
```suggestion
            Assert.assertEquals(headers.get("x-ably-clientid"),Base64Coder.encodeString("test client"));
```

ttypic:
thanks @sacOO7, yeah, looking at it right now, oddly it passes locally

lib/src/test/java/io/ably/lib/test/rest/HttpHeaderTest.java
ttypic: I prefer not to put logic for expected value. Potential bug in `Base64Coder#encodeString` method can make test unreliable

ttypic: I'll add comment that it is in fact base64 encoded "test client" value, to make it more readable

sacOO7: Sure, no issues. Just that, the assertion was more readable. Agree that assertions should be dumb and shouldn't be any logic into them.

sacOO7: lg

PR #1021, opened 2024-07-18
chore(Auth): get rid of unnecessary padding removal for Auth tokens

ttypic:
Makes ably-java SDK consistent with other SDKs. In other SDKs we use base64 encoding with padding for Auth tokens


sacOO7 (approved):
lgtm.

PR #1023, opened 2024-07-23
tests: Assert connection error code rather than message

lmars:
The error messages are subject to change, but the codes are not, so assert the code instead.


ttypic (approved):
LGTM

PR #1034, opened 2024-09-25
chore: update gradle wrapper

ttypic:


java/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+    exclude("io/ably/lib/BuildConfig.java")
+}
+
+tasks.register<Test>("testRealtimeSuite") {
+    filter {
+        includeTestsMatching("*RealtimeSuite")
+    }
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
+    jvmArgs("--add-opens", "java.base/java.lang=ALL-UNNAMED")
+    beforeTest(closureOf<TestDescriptor> { logger.lifecycle("-> $this") })
```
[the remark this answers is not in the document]
ttypic: unfortunately you are wrong, you can't do this, `beforeTest` argument is groovy's closure, you can't treat it like functional interface

.github/workflows/check.yml
```diff
@@ -12,4 +12,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: ./gradlew checkstyleMain checkstyleTest checkWithCodenarc runUnitTests
+      - name: Set up the JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
```
sacOO7, line 18: Why are we not using latest java LTS version, java-21 ?

CONTRIBUTING.md
```diff
@@ -201,8 +201,7 @@ implementation files('libs/ably-android-1.2.42.aar')
 This library uses [semantic versioning](http://semver.org/). For each release, the following needs to be done:
 
 1. Create a branch for the release, named like `release/1.2.4` (where `1.2.4` is what you're releasing, being the new version)
-2. Replace all references of the current version number with the new version number (check the [README.md](./README.md) and [common.gradle](./common.gradle)) and commit the changes
-    a. Increment the `versionCode` in the Android project's `build.gradle` by 1
```
sacOO7, line 205: Can we just double check if updated steps work for release process ?

sacOO7:
Overall looks good. Few points to be added
1. Make sure new build scripts work as expected, run all tasks under CONTRIBUTING.md
2. Update newly/updated added gradle tasks under `CONTRIBUTING.md`( if needed )
3. Ably-java lint/lint-format task is missing, will be great if we can add it.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,49 @@
...
+java-websocket = "1.5.3"
+vcdiff = "0.1.1"
+hamcrest = "1.3"
+nanohttpd = "2.3.0"
+mockito = "1.10.19"
+concurrentunit = "0.4.2"
+slf4j = "1.7.30"
+build-config = "5.4.0"
+firebase-messaging = "22.0.0"
+android-test = "0.5"
```
sacOO7, line 15: Have we added firebase as a new dependency? Couldn't find it in old dependencies

sacOO7: Okay, seems this is a Android specific dependency. Just wanted to make sure this doesn't get added into ably-java jar

dependencies.gradle
```diff
@@ -1,16 +0,0 @@
-// These dependencies have to be in lib/build.gradle for compilation _and_
-// in java/build.gradle and android/build.gradle for maven.
-dependencies {
```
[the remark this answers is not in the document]
ttypic: it's not a gradle way to do things, because of this we have two dependencies blocks in android build

CONTRIBUTING.md
ttypic: yes, I double checked this, now we can even create github workflow to make a release for us

.github/workflows/check.yml
ttypic: Actually we are using Java 8 everywhere :) Java 17 is used only to run gradle process. I prefer not to rush in here and use latest LTS, because gradle adds java support with some delay. We always can update with next gradle update

sacOO7 (approved):
LGTM

Issue #1041, opened 2024-10-14
Race condition when calling`AblyRealtime#connect()` on terminated state


ttypic:
There is a race condition during connection termination in the `ConnectionManager`. If `connect()` is called shortly after the connection is terminated, a new connection is not created because the old one still exists.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5033) by [Unito](https://www.unito.io)

PR #1042, opened 2024-10-14
[ECO-5033] fix: race condition when calling`AblyRealtime#connect()` on terminated state

ttypic:
Resolves https://github.com/ably/ably-java/issues/1041


lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -790,7 +791,12 @@ public synchronized State getConnectionState() {
     public synchronized void connect() {
         /* connect() is the only action that will bring the ConnectionManager out of a terminal currentState */
         if(currentState.terminal || currentState.state == ConnectionState.initialized) {
-            startup();
+            try {
+                startup();
+            } catch(InterruptedException e) {
+                Thread.currentThread().interrupt();
+                Log.e(TAG, "Failed to start up connection", e);
+            }
```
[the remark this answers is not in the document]
ttypic: nice catch! fixed

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -790,7 +791,13 @@ public synchronized State getConnectionState() {
...
         /* connect() is the only action that will bring the ConnectionManager out of a terminal currentState */
         if(currentState.terminal || currentState.state == ConnectionState.initialized) {
-            startup();
+            try {
+                startup();
+            } catch(InterruptedException e) {
+                Thread.currentThread().interrupt();
+                Log.e(TAG, "Failed to start up connection", e);
+                throw new RuntimeException("Failed to start up connection", e);
+            }
```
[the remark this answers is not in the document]
ttypic: makes sense, fixed

sacOO7 (changes requested):
Can we add test around this covering happy and failing path?

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -71,6 +71,11 @@ public class ConnectionManager implements ConnectListener {
     static ErrorInfo REASON_REFUSED = new ErrorInfo("Access refused", 401, 40100);
     static ErrorInfo REASON_TOO_BIG = new ErrorInfo("Connection closed; message too large", 400, 40000);
 
+    /**
+     * Flag that indicating that we entered termination state
+     */
+    private boolean terminating = false;
```
sacOO7, line 77: Since, this is being accessed from multiple threads, I guess we should use 
[AtomicBoolean](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicBoolean.html)

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -790,7 +797,13 @@ public synchronized State getConnectionState() {
     public synchronized void connect() {
         /* connect() is the only action that will bring the ConnectionManager out of a terminal currentState */
         if(currentState.terminal || currentState.state == ConnectionState.initialized) {
-            startup();
+            try {
+                startup();
+            } catch(InterruptedException e) {
```
sacOO7, line 802: Which thread exactly throws this exception and which one catches it?

sacOO7: I believe external thread is responsible for catching it.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -790,7 +797,13 @@ public synchronized State getConnectionState() {
...
+                Thread.currentThread().interrupt();
```
sacOO7, line 803:
```
     Thread.currentThread().interrupt();
```
Does this mean we are interrupting external thread

sacOO7: Maybe we should rethrow exception and let user handle it externally wdyt

ttypic: It's just the way java works, we are restoring interruption flag, we don't interrupt any thread. We can't rethrow, we can wrap in some `RuntimeException`, but it probably won't be useful at all. If somebody interrupts caller thread when we are waiting termination, I believe it's expecting behavior to ignore call to `connect()`. It's extremely rare scenario anyway, it also can be harmful to throw unexpected runtime exception

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: Yeah, I kind of agree, `terminating` is not being set at lot of places.

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: Thanks going through this 👍

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: Okay, so we are following https://www.baeldung.com/java-interrupted-exception#2-restore-the-interrupt. 
This means we will restore the status of being interrupted and caller method will be responsible for handling it right

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: Okay, lgtm. 
Okay, so we are following https://www.baeldung.com/java-interrupted-exception#2-restore-the-interrupt.
The concept is similar to cancellationToken` in dotnet, `cancellationContext` in go .. etc, 
Feels like one of the thread synchronization mechanism.

sacOO7 (approved):
LGTM

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7, line 77:
```suggestion
    /**
     * Flag indicating existing active connection with the server
     */
    private boolean activeConnection = false;
```
`terminating` word does feel a bit misleading.

sacOO7 (changes requested):

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
sacOO7: @coderabbitai wdyt about this?

sacOO7: wdyt about this @coderabbitai

sacOO7: Well, this will ofc need inversion of `terminating` checks/assignment at other places as well.

sacOO7: I will leave it to @ttypic for renaming the variable. There is ofc an option to rename it to `activeConnection=false`
or `active=false`

sacOO7 (approved):
LGTM

PR #1048, opened 2024-11-27
[ECO-5139] feat: add `action` and `serial` fields

ttypic:
Add `action` and `serial` fields to the message object for the Chat SDK

## Summary by CodeRabbit

- **New Features**
	- Introduced a method for reading optional integer values from JSON.
	- Added new fields (`serial`, `version`, `action`, `createdAt`) to the Message class for enhanced message details.
	- Created a new enum (`MessageAction`) to represent various message actions.
	- Enhanced message handling logic to improve tracking of message versions during transmission.
	- Added a new capability in the JSON configuration to allow all actions.

- **Bug Fixes**
	- Improved serialization and deserialization processes to handle new fields effectively.
	- Refined error handling and state management during message processing and recovery.

- **Tests**
	- Expanded test coverage for the Message class, including new tests for serialization and deserialization with the new fields.
	- Added tests to validate message properties in the RealtimeMessageTest class.

### DIFF
--- a/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
+++ b/lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
@@ -25,6 +25,7 @@ import io.ably.lib.types.DecodingContext;
 import io.ably.lib.types.DeltaExtras;
 import io.ably.lib.types.ErrorInfo;
 import io.ably.lib.types.Message;
+import io.ably.lib.types.MessageAction;
 import io.ably.lib.types.MessageDecodeException;
 import io.ably.lib.types.MessageSerializer;
 import io.ably.lib.types.PaginatedResult;
@@ -843,6 +844,12 @@ public abstract class ChannelBase extends EventEmitter<ChannelEvent, ChannelStat
             if(msg.connectionId == null) msg.connectionId = protocolMessage.connectionId;
             if(msg.timestamp == 0) msg.timestamp = protocolMessage.timestamp;
             if(msg.id == null) msg.id = protocolMessage.id + ':' + i;
+            // (TM2p)
+            if(msg.version == null) msg.version = String.format("%s:%03d", protocolMessage.channelSerial, i);
+            // (TM2k)
+            if(msg.serial == null && msg.action == MessageAction.MESSAGE_CREATE) msg.serial = msg.version;
+            // (TM2o)
+            if(msg.createdAt == null && msg.action == MessageAction.MESSAGE_CREATE) msg.createdAt = msg.timestamp;
 
             try {
                 msg.decode(options, decodingContext);
--- a/lib/src/test/java/io/ably/lib/types/MessageTest.java
+++ b/lib/src/test/java/io/ably/lib/types/MessageTest.java
@@ -46,4 +46,69 @@ public class MessageTest {
         assertEquals("test-data", serializedObject.get("data").getAsString());
         assertEquals("test-name", serializedObject.get("name").getAsString());
     }
+
+    @Test
+    public void serialize_message_with_serial() {
+        // Given
+        Message message = new Message("test-name", "test-data");
+        message.clientId = "test-client-id";
+        message.connectionKey = "test-key";
+        message.action = MessageAction.MESSAGE_CREATE;
+        message.serial = "01826232498871-001@abcdefghij:001";
+
+        // When
+        JsonElement serializedElement = serializer.serialize(message, null, null);
+
+        // Then
+        JsonObject serializedObject = serializedElement.getAsJsonObject();
+        assertEquals("test-client-id", serializedObject.get("clientId").getAsString());
+        assertEquals("test-key", serializedObject.get("connectionKey").getAsString());
+        assertEquals("test-data", serializedObject.get("data").getAsString());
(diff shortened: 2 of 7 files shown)


sacOO7 (commented):
Related to https://github.com/ably/specification/pull/210/files and https://github.com/ably/specification/pull/235/files. I'm not sure why we didn't address other materialization fields as a part of this PR

ttypic:
@sacOO7, 

`version` is not important for now, originally I added `version` as well, but decided to rollback it, this PR focused on fields that's important for Chat Beta. For Chat Beta we don't need to implement all parts of the spec, we just should be able to filter non "message.create" actions and use `serial` field instead of `timeserial` from extras.

> why we didn't address other materialization fields as a part of this PR

It requires some thinking, more tests and it will be better to implement as part of separate PR

sacOO7:
Sure, but the spec says `the SDK must set it equal to its @TM2p@ @version@ if @MESSAGE_CREATE@ does not contain a @serial@`. Won't this cause issue while reading `serial` for `MESSAGE_CREATE`
Also, we should create issue to address other fields, same as [Cocoa-Add new properties to ARTMessage](https://github.com/ably/ably-cocoa/issues/2000) if current implementation is sufficient

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -843,6 +843,10 @@ private void onMessage(final ProtocolMessage protocolMessage) {
             if(msg.connectionId == null) msg.connectionId = protocolMessage.connectionId;
             if(msg.timestamp == 0) msg.timestamp = protocolMessage.timestamp;
             if(msg.id == null) msg.id = protocolMessage.id + ':' + i;
+            // (TM2p)
+            if(msg.version == null) msg.version = String.format("%s:%03d", protocolMessage.channelSerial, i);
+            // (TM2k)
+            if(msg.serial == null) msg.serial = msg.version;
```
AndyTWF, line 849: Also needed here is `TM2o` for the `createdAt` timestamp (needs to be specified if not provided by the server)

lib/src/test/java/io/ably/lib/types/MessageTest.java
```diff
@@ -46,4 +46,69 @@ public void serialize_message_with_name_and_data() {
         assertEquals("test-data", serializedObject.get("data").getAsString());
         assertEquals("test-name", serializedObject.get("name").getAsString());
     }
+
+    @Test
+    public void serialize_message_with_serial() {
```
AndyTWF, line 51: Do we also need a test (not this file) that handles the inbound stuff - e.g. a quick integration test that ensures fields are set?

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
```diff
@@ -843,6 +843,12 @@ private void onMessage(final ProtocolMessage protocolMessage) {
...
+            // (TM2o)
+            if(msg.createdAt == null) msg.createdAt = msg.timestamp;
```
AndyTWF, line 851: To be pedantic on TM2o - we should only set this if the msg.action is `MESSAGE_CREATE`

lib/src/test/java/io/ably/lib/types/MessageTest.java
sacOO7: Yeah, agree to this one 👍

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
sacOO7: @AndyTWF not sure why this was closed? 
We do the same in case of swift -> https://github.com/ably/ably-cocoa/blob/9f1cbd92c0456ccca486a6efedfceb572a0d7bb2/Source/ARTJsonLikeEncoder.m#L289

AndyTWF: Ah sorry, I'd opened a separate thing about the `msg.action` check for `createdAt` - but yes, the same applies to the `serial field`

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ttypic: Fixed for serial as well! Not sure why we make it more complicated by adding this additional condition :)

lib/src/main/java/io/ably/lib/realtime/ChannelBase.java
ttypic: It makes no sense to me, especially in java, there NPE is most common reason of app crash, but I fixed this everywhere

lib/src/test/java/io/ably/lib/types/MessageTest.java
ttypic: @sacOO7 integration test has been added here: [lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java](https://github.com/ably/ably-java/pull/1048/files#diff-308e503e1c8c03d7814aa64771fdd96e4d51bb00f0ab44c945cfacf3e396a08f)

sacOO7 (approved):
Lgtm

AndyTWF (approved):

Issue #1050, opened 2024-12-03
Java SKD - Duplicate messages on rewind after 1.2.34


leni-kirilov:
With the help of Mike from Ably Support , it turns out there is a regression since 1.2.34 (I caught it with 1.2.43)

An integration test of mine sends 3 messages and reads them with rewind. 
**It started receiving them 6 times...**

My code is roughly the following:
```
assertRewindOrHistory("?rewind=10", 3, 3, testInfo);
private void assertRewindOrHistory(String rewindString, int existingMessagesCount, int expectedMessageCount, TestInfo testInfo) throws Exception {

    //GIVEN a channel with some messages exist
    String testSpecificChannel = HARDCODED_PUBLIC_CHANNEL_NAME + testInfo.getDisplayName() + System.currentTimeMillis();
    List<String> testMessages = List.of("message 1", "message 2", "message 3");
    ablyService.publishMessage(ablyProperties.rootApiKey(), testInfo.getDisplayName(), testSpecificChannel, testMessages);
    TimeUnit.SECONDS.sleep(1);

    //AND a customer is waiting to consume data via Websocket using a temp auth token
    var customerToken = ablyAuthService.getAuthTokenForPublicConsumption(testInfo.getDisplayName(), testSpecificChannel);
    var customerAbly = ablyService.connect(customerToken);

    //WHEN registering a consumer and rewind x messages in the past
    Consumer<Message> messageConsumer = (message) -> {
        Assertions.assertNotNull(message.data, "This is the message: " + message);
        actualMessageCount.incrementAndGet();
    };

    ablyService.consumeMessage(customerAbly, rewindString + testSpecificChannel, messageConsumer);

    //THEN expect all past messages to be read
    assertAsync(expectedMessageCount, actualMessageCount, "Reading different messages count from " + testSpecificChannel);
}
// AblyService.java
public void consumeMessage(AblyRealtime ably, String channelName, Consumer<Message> assertion) {
    logger.info("Consuming messages from channel " + channelName);

    connectAnd(ably, (AblyRealtime ably2) -> {
        Channel channel = ably2.channels.get(channelName);

        try {
            logger.info("Subscribed to the channel" + channel.name);
            channel.subscribe(assertion::accept);
        } catch (AblyException e) { throw new RuntimeException(e); }
    });
}
private void connectAnd(AblyRealtime ably, Consumer<AblyRealtime> consumer) {
    ably.connection.on(ConnectionState.connected, state -> {
        switch (state.current) {
[...]

Issue #1018, opened 2024-07-11
Avoid duplicate ATTACH sent


sacOO7:
- `Attach` related changes were introduced as a part of protocol 2.
- As per protocol 2, ATTACH msg should be sent for every channel in `attaching`, `attached` and `suspended` states once `connected`.
- There are various mechanisms by which duplicate ATTACH can be sent
1. `ATTACH` is added to the connection or global queue before being sent, check description on -> https://github.com/ably/ably-ruby/pull/416
2. `ATTACH` is sent periodically when channel is in `SUSPENDED` state as per https://sdk.ably.com/builds/ably/specification/main/features/#RTL13b. 
3. In case of `ably-java` we had observed duplicate `ATTACHED` being received in the response during re-connection. Need to check if this is the case and how can we avoid it.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4872) by [Unito](https://www.unito.io)

sacOO7:
Proposed fix at https://github.com/ably/ably-java/pull/1052

PR #1052, opened 2024-12-05
[ECO-5163] fix: duplicated messages because of duplicated attach message

ttypic:
Resolves https://github.com/ably/ably-java/issues/1050

Implemented missing [(RTL4i)](https://sdk.ably.com/builds/ably/specification/main/features/#RTL4i) spec point to prevent duplicated attach message

## Summary by CodeRabbit

- **New Features**
	- Added a conditional check to enhance channel attachment logic, preventing premature attach requests during specific connection states.
	- Introduced a new test to ensure message integrity by verifying that messages are not duplicated under certain subscription conditions.

- **Bug Fixes**
	- Improved message handling to prevent duplication of attach messages during connection events.

- **Tests**
	- Added a test method to validate message delivery without duplication.
	- Enhanced the flexibility of connection failure tests by using dynamic key assignments.

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
@@ -13,12 +13,14 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import io.ably.lib.types.ChannelOptions;
 import io.ably.lib.types.MessageAction;
 import io.ably.lib.types.MessageExtras;
 import io.ably.lib.types.Param;
@@ -995,6 +997,10 @@ public class RealtimeMessageTest extends ParameterizedTest {
                 msgComplete.onSuccess();
             });
 
+            CompletionWaiter attachListener = new CompletionWaiter();
+            channel.attach(attachListener);
+            assertNull(attachListener.waitFor(1, 10_000));
+
             /* publish to the channel */
             JsonObject chatMessage = new JsonObject();
             chatMessage.addProperty("text", "hello world!");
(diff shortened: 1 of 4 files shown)


sacOO7 (commented):
Thanks for raising the PR to fix the issue!
checks for `ATTACH` are bit tricky and has corner cases mentioned here https://github.com/ably/ably-java/issues/1018
So, I will take my time to carefully review the PR to make sure all cases are covered

sacOO7:
Also, it seems some of the realtime tests failing

lib/src/test/java/io/ably/lib/test/realtime/RealtimeMessageTest.java
```diff
@@ -1010,4 +1014,42 @@ public void should_have_serial_action_createdAt() throws AblyException {
...
+                .collect(Collectors.toList());
+            channel.publish(messages.toArray(Message[]::new));
+        }
+
+        try (AblyRealtime realtime = new AblyRealtime(opts)) {
+            final ChannelOptions options = new ChannelOptions();
+            options.params = Map.of("rewind", "10");
+            final Channel channel = realtime.channels.get(testChannelName, options);
+            final CompletionWaiter completionWaiter = new CompletionWaiter();
+            final AtomicInteger counter = new AtomicInteger();
```
sacOO7, line 1038: Also, we can avoid using `AtomicInteger` since the internal `emit` call for channel messages is blocking in nature and the next event will only be sent when the previous one is processed

sacOO7: Seems, IDE suggests to make it either atomic or final : (

sacOO7: I think test can be further simplified to 
```
        ClientOptions opts = createOptions(testVars.keys[0].keyStr);
        opts.logLevel = 2;
        String testChannelName = "my-channel" + System.currentTimeMillis();
        try (AblyRest rest = new AblyRest(opts)) {
            final io.ably.lib.rest.Channel channel = rest.channels.get(testChannelName);
            Message[] messages = new Message[3];
            for (int i = 0; i < messages.length; i++) {
                messages[i] = new Message("name", "message" + i);
            }
            channel.publish(messages);
        }

        try (AblyRealtime realtime = new AblyRealtime(opts)) {
            final ChannelOptions options = new ChannelOptions();
            options.params = Map.of("rewind", "10");
            final Channel channel = realtime.channels.get(testChannelName, options);

            final int[] receivedMessages = {0};
            channel.subscribe((Channel.MessageListener) message -> receivedMessages[0] = receivedMessages[0] + 1);

            Exception conditionError = new Helpers.ConditionalWaiter().wait(() -> receivedMessages[0] == 3, 10000);
            assertNull(conditionError);

            assertEquals("Should be exactly 3 messages",3, receivedMessages[0]);

            Thread.sleep(1500);

            assertEquals("Should be exactly 3 messages even after 1.5 sec wait",3, receivedMessages[0]);
```
`ConditionalWaiter` is same as `assertWaiter`

ttypic: it's a test, we just need wrapper around counter :) we don't care about atomic's overhead here (it's probably less than millisecond). Overhead from our `CompletionWaiter` are incomparable to it

sacOO7: IDE is saying, `messages.toArray(Message[]::new)` is only compatible with java 11+

ttypic: once again it's a test, this code will be never executed by library users we still target 1.8

sacOO7: Yeah, but not sure why IDE is signaling warning for use of `List.of` and `Map.of` ( java 9+ compatible )

sacOO7: I think we can ignore those warnings, but shouldn't change IDE settings for the same.
Doing so can introduce java 8+ code in the source code as well.

ttypic: I rewrote to use 1.8, to get rid of IDE warnings

sacOO7 (commented):
Okay, not sure if this is a sandbox issue, but it seems `realtime.RealtimeMessageTest` are consistently failing on both `okhttp` and `websocket` connection.
You can check what's going wrong there : (

sacOO7:
If this is a sandbox issue, then I will approve PR right away ( since `should_not_duplicate_messages` seems to be passing )

sacOO7 (approved):
LGTM

PR #1060, opened 2025-02-05
[CHAT-5063] feat: sdk proxy wrapper for agent tracking

ttypic:
We want to be able to distinguish requests made inside the wrapper from those made by the core PubSub SDK without the wrapper, allowing us to track agents across wrapper SDKs such as the Chat SDK or Asset Tracking. To achieve this, we introduce special proxy Realtime and REST clients that inject additional `DerivedClientOptions` parameters into the underlying SDK.


lib/src/main/java/io/ably/lib/rest/DerivedClientOptions.java
```diff
@@ -0,0 +1,33 @@
+package io.ably.lib.rest;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class DerivedClientOptions {
```
lawrence-forooghian, line 6: A couple of notes about naming — people didn't like the "derived client" naming on the DR, so in the ably-cocoa PR (https://github.com/ably/ably-cocoa/pull/2014) I settled on the following more explicit naming:

- the options class is called `WrapperSDKProxyOptions`
- the method is called `createWrapperSDKProxy`

lawrence-forooghian:
To check I've understood the scope of this PR — this only addresses adding the wrapper SDK's agents to the `Ably-Agent` header, right? Doing the `agent` channel param will come separately?

lawrence-forooghian:
Would you mind explaining to me _which_ HTTP requests will have the wrapper SDK proxy agent added to the `Ably-Agent` header? For example, if the library fetches an auth token, might that request contain a wrapper SDK proxy agent? I ask because on the DR [we decided we wanted to be specific about which requests would be attributed to a wrapper SDK](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued#Which-REST-requests-do-we-want-to-be-able-to-attribute-to-a-modern-wrapper-SDK%3F).

lib/src/main/java/io/ably/lib/rest/DerivedClientOptions.java
ttypic: Got it! Do you think it would be less obvious if we removed ‘SDK’ from the name, e.g., `WrapperProxyOptions` and `createWrapperProxy`?

ttypic:
> To check I've understood the scope of this PR — this only addresses adding the wrapper SDK's agents to the `Ably-Agent` header, right? Doing the `agent` channel param will come separately?

Yes, I think it's a bit lesser priority for now, but it shouldn't be hard to add

ttypic:
> Would you mind explaining to me which HTTP requests will have the wrapper SDK agent added to the Ably-Agent header? For example, if the library fetches an auth token, might that request contain a wrapper SDK agent? I ask because on the DR [we decided we wanted to be specific about which kinds of request would be attributable to a wrapper SDK](https://ably.atlassian.net/wiki/spaces/ENG/pages/3413016589/ADR-117+Tracking+wrapper+SDK+usage+continued#Which-REST-requests-do-we-want-to-be-able-to-attribute-to-a-modern-wrapper-SDK%3F).

The agent will be added to all requests initiated by the wrapper instance, including potentially token requests initiated by the wrapper. Personally, I believe the more we cover, the better.  I think we should cover billable API calls "at a minimum", I’m not sure it would be a problem if we included all requests. Additionally, distinguishing between billable and non-billable APIs would require extra code, which we would then need to maintain.

lib/src/main/java/io/ably/lib/rest/DerivedClientOptions.java
lawrence-forooghian: Personally I feel that it's clearer with "SDK"…

ttypic:
Closed in favor of https://github.com/ably/ably-java/pull/1066

PR #1066, opened 2025-02-13
[CHAT-5063] feat: add `createWrapperSdkProxy` implementation for kotlin adapter of ably-java

ttypic:
We want to be able to distinguish requests made inside the wrapper from those made by the core PubSub SDK without the wrapper, allowing us to track agents across wrapper SDKs such as the Chat SDK or Asset Tracking. To achieve this, we introduce special proxy Realtime and Rest clients that inject additional agents parameters into the underlying SDK.

To do this we introduced several package-private methods that make it possible to use custom http module to invoke Rest API, so we can inject agent information.

Also special `SdkWrapperCompatible` introduced, allowing us to hide `createWrapperSdkProxy` from the public interface and use extension function instead

## Summary by CodeRabbit

- **New Features**
	- Upgraded dependency versions and added new libraries ensuring improved performance.
	- Introduced proxy wrapper capabilities that enable customizable HTTP handling and dynamic agent injection.
- **Refactor**
	- Streamlined and standardized HTTP handling across client methods for more consistent parameter management.
	- Enhanced modularity in channels and presence operations with explicit HTTP context support.
- **Tests**
	- Expanded testing with an embedded HTTP server, new test suites, and utility functions for robust asynchronous operations.
- **Chores**
	- Updated build configurations and dependency management to support the enhanced testing framework.

### DIFF
--- a/android/src/main/java/io/ably/lib/rest/AblyRest.java
+++ b/android/src/main/java/io/ably/lib/rest/AblyRest.java
@@ -1,8 +1,6 @@
 package io.ably.lib.rest;
 
 import android.content.Context;
-import io.ably.lib.http.Http;
-import io.ably.lib.http.HttpCore;
 import io.ably.lib.push.LocalDevice;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
@@ -36,24 +34,6 @@ public class AblyRest extends AblyBase {
      */
     public AblyRest(ClientOptions options) throws AblyException {
         super(options, new AndroidPlatformAgentProvider());
-    }
-
-    /**
-     * Constructor implementation to be able to have shallow copy of the client,
-     * allowing us to modify certain fields while implementing a proxy for the Realtime/Rest SDK wrapper
-     */
-    protected AblyRest(AblyRest underlyingClient, HttpCore httpCore, Http http) {
-        super(underlyingClient, httpCore, http);
-    }
-
-    /**
-     * [Internal Method]
-     * <p/>
-     * We use this method to create a shallow copy of the client, allowing us to modify certain fields
-     * while implementing a proxy for the Realtime/Rest SDK wrapper
-     */
-    public AblyRest createShallowCopy(HttpCore httpCore, Http http) {
-        return new AblyRest(this, httpCore, http);
     }
 
     /**
--- a/android/src/main/java/io/ably/lib/rest/AblyRest.java
+++ b/android/src/main/java/io/ably/lib/rest/AblyRest.java
@@ -1,6 +1,8 @@
 package io.ably.lib.rest;
 
 import android.content.Context;
+import io.ably.lib.http.Http;
+import io.ably.lib.http.HttpCore;
 import io.ably.lib.push.LocalDevice;
 import io.ably.lib.types.AblyException;
 import io.ably.lib.types.ClientOptions;
@@ -34,6 +36,24 @@ public class AblyRest extends AblyBase {
      */
     public AblyRest(ClientOptions options) throws AblyException {
         super(options, new AndroidPlatformAgentProvider());
+    }
+
+    /**
+     * Constructor implementation to be able to have shallow copy of the client,
+     * allowing us to modify certain fields while implementing a proxy for the Realtime/Rest SDK wrapper
+     */
+    protected AblyRest(AblyRest underlyingClient, HttpCore httpCore, Http http) {
+        super(underlyingClient, httpCore, http);
+    }
+
+    /**
+     * [Internal Method]
+     * <p/>
+     * We use this method to create a shallow copy of the client, allowing us to modify certain fields
+     * while implementing a proxy for the Realtime/Rest SDK wrapper
+     */
+    public AblyRest createShallowCopy(HttpCore httpCore, Http http) {
+        return new AblyRest(this, httpCore, http);
     }
 
     /**
--- a/pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
+++ b/pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
@@ -1,0 +1,20 @@
+package com.ably.pubsub
+
+data class WrapperSdkProxyOptions(val agents: Map<String, String>)
+
+interface SdkWrapperCompatible<T> {
+
+  /**
+   * Creates a proxy client to be used to supply analytics information for Ably-authored SDKs.
+   * The proxy client shares the state of the `RealtimeClient` or `RestClient` instance on which this method is called.
+   * This method should only be called by Ably-authored SDKs.
+   */
+  fun createWrapperSdkProxy(options: WrapperSdkProxyOptions): T
+}
+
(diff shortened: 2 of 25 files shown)


sacOO7:
Had a quick review of the PR, seems I also need to have spec PR https://github.com/ably/specification/pull/274/files as a reference. I will do more deeper review once I understand the context properly.

lawrence-forooghian:
@sacOO7 the spec PR that you linked to is for the chat spec; the relevant one here is https://github.com/ably/specification/pull/273 🙂

pubsub-adapter/src/main/kotlin/io/ably/lib/realtime/WrapperRealtimeClient.kt
```diff
@@ -1,50 +1,87 @@
...
 import io.ably.lib.types.*
 
-internal class WrapperRealtimeClient(private val javaClient: AblyRealtime, private val httpModule: Http) : RealtimeClient by RealtimeClientAdapter(javaClient) {
-  override val channels: Channels<out RealtimeChannel>
-    get() = WrapperRealtimeChannels(javaClient.channels, httpModule)
+internal class WrapperRealtimeClient(
+    private val javaClient: AblyRealtime,
+    private val httpModule: Http,
+    private val agents: Map<String, String>,
+) : RealtimeClient by RealtimeClientAdapter(javaClient) {
```
lawrence-forooghian, line 16: Does the `request` method need to be overridden here to inject the wrapper SDK agents, or is there some other mechanism that handles this that I haven't understood?

lawrence-forooghian (commented):
Took a first look at it and, whilst not understanding all of the details, I _think_ it makes sense roughly. But have asked a question that should help my understanding :)

android/src/main/java/io/ably/lib/rest/AblyRest.java
```diff
@@ -36,6 +38,24 @@ public AblyRest(ClientOptions options) throws AblyException {
...
+        super(underlyingClient, httpCore, http);
+    }
+
+    /**
+     * [Internal Method]
+     * <p/>
+     * We use this method to create a shallow copy of the client, allowing us to modify certain fields
+     * while implementing a proxy for the Realtime/Rest SDK wrapper
+     */
+    public AblyRest createShallowCopy(HttpCore httpCore, Http http) {
```
sacOO7, line 55: We only have `createWrapperSDKProxy` as a public method right? maybe we should mark this as protected.
If it's not accessible, we can access it using [invoke private java method](https://github.com/ably/ably-chat-kotlin/blob/71b20c4b70c93fd91d3356d743843da514df70cb/chat-android/src/test/java/com/ably/chat/TestUtils.kt#L110) ( just need to remove suspendCancellableCoroutine )

sacOO7: This is to make sure we don't create out of spec public methods

pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Creates a proxy client to be used to supply analytics information for Ably-authored SDKs.
+   * The proxy client shares the state of the `RealtimeClient` or `RestClient` instance on which this method is called.
+   * This method should only be called by Ably-authored SDKs.
+   */
+  fun createWrapperSdkProxy(options: WrapperSdkProxyOptions): T
+}
+
+fun RealtimeClient.createWrapperSdkProxy(options: WrapperSdkProxyOptions): RealtimeClient =
```
sacOO7, line 15: As per `WP4`, wrapper proxy client shouldn't have `createWrapperSdkProxy` method on it

sacOO7: When looked into the code, seems it's only used in tests, so maybe we should move it into test helpers

android/src/main/java/io/ably/lib/rest/AblyRest.java
ttypic: Good spot, I guess we can make it protected, I think we don't use it outside the package. Let me check. But "invoke private java method" relies on reflection and tampering with method visibility, which is always best to avoid

pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
sacOO7: Or `createWrapperSdkProxy` should return interface that doesn't have `createWrapperSdkProxy` method on it. We shouldn't be able to apply `createWrapperSdkProxy` on client created by the same method.

sacOO7: I guess current kotlin module can be published and recommended to be used as public API pubsub library instead of `ably-java`, since it exposes right public API methods.

pubsub-adapter/src/main/kotlin/io/ably/lib/realtime/WrapperRealtimeClient.kt
ttypic: I used to patch AblyRealtime client itself, but I agree it's not very obvious.  Now I override all methods instead and do nothing with underlying realtime client

android/src/main/java/io/ably/lib/rest/AblyRest.java
ttypic: refactored a bit, get rid of this code

pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
ttypic: > When looked into the code, seems it's only used in tests, so maybe we should move it into test helpers

These are special extension functions for the high-level SDK (currently, specifically for Chat).

> We shouldn't be able to call `createWrapperSdkProxy` on the client created by the same method

The spec states that we don’t need to support this (as it might be difficult to implement), but we definitely could. The current implementation allows nested calls like:
`client.createWrapperSdkProxy(agents1).createWrapperSdkProxy(agents2)`,
which results in both agent groups being dynamically added to all requests.

> I guess current kotlin module can be published and recommended to be used as public API pubsub library instead of ably-java, since it exposes right public API methods.

That’s the long-term goal, but the current public API interfaces are not quite right yet. I’ll continue improving them.

android/src/main/java/io/ably/lib/rest/AblyRest.java
sacOO7: Great

pubsub-adapter/src/main/kotlin/com/ably/pubsub/WrapperSdkProxy.kt
sacOO7: Okay, I think this shouldn't be a issue for the time being.
Looks okay for now

sacOO7 (approved):
Looks good to me 👍

Issue #1067, opened 2025-02-24
Warning on Android Proguard


fanwgwg:
I'm seeing this warning when using ably-java in an android project with proguard enabled
```

WARNING: R8: Missing class lombok.Generated (referenced from: io.ably.lib.network.ProxyConfig io.ably.lib.network.HttpEngineConfig$HttpEngineConfigBuilder.proxy and 101 other contexts)
```

In my proguard rules, I've already added this line as mentioned in the doc
```
-keep class io.ably.lib.** { *; } -keep class org.msgpack.core.** { *; }
```

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5228) by [Unito](https://www.unito.io)

ttypic:
Hey @fanwgwg,

Thanks for reporting this! The issue should be resolved in the latest ably-android `1.2.50`. Feel free to reopen it if you’re still able to reproduce it

PR #1069, opened 2025-02-24
[CHAT-5228] fix: add lombok to ProGuard rules

ttypic:
Fixes https://github.com/ably/ably-java/issues/1067

We use Lombok deep down internally for code generation but Lombok itself should not be included in the final APK.

Effect of `-dontwarn lombok.**`:
- It tells ProGuard to ignore missing Lombok classes, preventing unnecessary warnings.
- This avoids potential build failures when ProGuard is set to treat warnings as errors.

## Summary by CodeRabbit

- **Chores**
  - Updated configuration settings to suppress non-critical warnings from an external utility library, helping to streamline the build process and improve overall stability.

### DIFF
--- a/android/proguard.txt
+++ b/android/proguard.txt
@@ -5,3 +5,4 @@
 -keep class com.google.gson.** {*;}
 -dontwarn org.msgpack.core.buffer.**
 -dontwarn org.slf4j.**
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
lgtm

PR #1076, opened 2025-03-11
[ECO-5248] fix: make query params URL encoded

ttypic:
Fixed inconsistent encoding/decoding in HTTP requests. Value of `Param` object shouldn't be URL encoded, but if it used in as query params it should be URL encoded in the request.


pubsub-adapter/src/test/kotlin/com/ably/Utils.kt
```diff
@@ -15,3 +18,29 @@ suspend fun waitFor(timeoutInMs: Long = 10_000, block: suspend () -> Boolean) {
     }
   }
 }
+
+fun createAblyRealtime(port: Int): AblyRealtime {
```
sacOO7, line 22: Is it okay if we don't mark methods public explicitly 🤔. Or the check not applicable on tests?

sacOO7 (approved):
LGTM

pubsub-adapter/src/test/kotlin/com/ably/Utils.kt
ttypic: yeah, this rule applies for the main source sets only

Issue #1074, opened 2025-03-10
RTN11d spec point implemented incorrectly


ttypic:
RTN11d spec point implemented incorrectly, it states:

> [(RTN11d)](https://sdk.ably.com/builds/ably/specification/main/features/#RTN11d) If the state is CLOSED or FAILED, transitions all the channels to INITIALIZED and unsets their RealtimeChannel.errorReason, unsets the Connection.errorReason, clears all connection state (including in particular Connection.recoveryKey), and resets the msgSerial to 0

Current implementation:

```java
 connection.on(ConnectionEvent.closed, new ConnectionStateListener() {
            @Override
            public void onConnectionStateChanged(ConnectionStateListener.ConnectionStateChange state) {
                channels.clear();
            }
        });
```

It doesn't handle "failed" termination state and clears channels instead of reseting. Because of that channels can become orphan and  won't receive state changes

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5246) by [Unito](https://www.unito.io)

PR #1077, opened 2025-03-14
[ECO-5246] fix: Realtime Client Reconnection Logic

ttypic:
Resolves https://github.com/ably/ably-java/issues/1074

Previously, we cleared all channels on the close event, causing all previously acquired channels to become orphaned. We have now removed this logic and fixed the reconnection behavior to align with the spec

## Summary by CodeRabbit

- **Refactor**
  - Modified channel management to prevent automatic clearing of channels upon connection closure, allowing for persistent channel states.

- **New Features**
  - Introduced a mechanism to reinitialize channels, resetting their states and error conditions during reconnection for smoother recovery.

- **Tests**
  - Added tests to validate that channels properly reset when reconnecting from closed or closing states.

### DIFF
--- a/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
+++ b/lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
@@ -979,6 +979,7 @@ public class RealtimeChannelTest extends ParameterizedTest {
             assertEquals("Simulated connection failure", channel.reason.message);
 
             ably.connect();
+            new ConnectionWaiter(ably.connection).waitFor(ConnectionState.connected);
 
             Helpers.CompletionWaiter attachListener = new Helpers.CompletionWaiter();
             channel.attach(attachListener);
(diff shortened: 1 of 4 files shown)


sacOO7:
Not sure why tests are failing

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -2462,6 +2462,98 @@ public void detach_message_to_released_channel_is_dropped() throws AblyException
...
+            new ChannelWaiter(channel).waitFor(ChannelState.attached);
+            assertEquals("Verify attached state reached", channel.state, ChannelState.attached);
+
+            /* push a message to increase msgSerial */
+            channel.publish("test", "test");
+            assertEquals(1, ably.connection.connectionManager.msgSerial);
+
+            ably.close();
+            new ChannelWaiter(channel).waitFor(ChannelState.detached);
+
```
sacOO7, line 2494: check for connection closed state and msgserial != 0

ttypic: not sure these are important assertions for this particular test, but added

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
```diff
@@ -2462,6 +2463,100 @@ public void detach_message_to_released_channel_is_dropped() throws AblyException
...
+            List<ConnectionState> observedConnectionStates = new ArrayList<>();
+            ably.connection.on(stateChange -> observedConnectionStates.add(stateChange.current));
+
+            ably.close();
+            new ConnectionWaiter(ably.connection).waitFor(ConnectionState.closing);
+            ably.connect();
+
+            new ConnectionWaiter(ably.connection).waitFor(ConnectionState.connected);
+            assertEquals(List.of(ConnectionState.closing, ConnectionState.connecting, ConnectionState.connected), observedConnectionStates);
```
sacOO7, line 2548: Not sure, but it feels like the test would flake because the connection can go into `CLOSED` state, before we call 
`ably.connect` at L2545

sacOO7: Okay, ran multiple times on the local, works good 👍

sacOO7 (approved):
LGTM

lib/src/test/java/io/ably/lib/test/realtime/RealtimeChannelTest.java
ttypic: I rewrote test a little, make `connect()` instantly after `close()` so there will be no chance that closed message from network comes faster than two sequential statement

Issue #1079, opened 2025-03-31
`Java-WebSocket` holds lock while invoking listeners, it may cause deadlock


ttypic:
The `Java-WebSocket` library holds a lock while invoking listeners, and when a connection is terminated ungracefully, it can cause a deadlock. Here’s an explanation:

- The **WebSocketLibrary**’s connection checker thread executes `WebSocket.close()`.
- The **WebSocketLibrary** acquires a lock and starts invoking listeners while holding the lock.
- The **WebSocketHandler** has an internal Timer that holds a lock on the handler. This Timer decides to close the connection as well and calls `WebSocket.close()` while holding the lock.
- The **WebSocketLibrary** tries to dispose of the **Timer**  but can’t because the **Timer**  holds the lock.
- The **Timer**  tries to close the connection but can’t because the  **WebSocketLibrary** still holds the lock.

There are two issues here:

1. The **WebSocketLibrary** shouldn’t hold a lock while calling listeners.
2. The **Timer** shouldn’t hold a lock while calling `close()`.

We can’t fix the first issue, which is why we need to address the second one.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5276) by [Unito](https://www.unito.io)

PR #1080, opened 2025-03-31
fix: deadlock in `WebSocket.close()`

ttypic:
Fixes https://github.com/ably/ably-java/issues/1079

The `Java-WebSocket` library holds a lock while invoking listeners, and when a connection is terminated ungracefully, it can cause a deadlock. Here’s an explanation:

- The **WebSocketLibrary**’s connection checker thread executes `WebSocket.close()`.
- The **WebSocketLibrary** acquires a lock and starts invoking listeners while holding the lock.
- The **WebSocketHandler** has an internal Timer that holds a lock on the handler. This Timer decides to close the connection as well and calls `WebSocket.close()` while holding the lock.
- The **WebSocketLibrary** tries to dispose of the **Timer**  but can’t because the **Timer**  holds the lock.
- The **Timer**  tries to close the connection but can’t because the  **WebSocketLibrary** still holds the lock.

There are two issues here:

1. The **WebSocketLibrary** shouldn’t hold a lock while calling listeners.
2. The **Timer** shouldn’t hold a lock while calling `close()`.

We can’t fix the first issue, which is why we need to address the second one.

## Summary by CodeRabbit

- **Refactor**
  - Improved connection handling to enhance overall stability and performance, ensuring a more robust and responsive experience during periods of activity.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -381,8 +381,7 @@ public class WebSocketTransport implements ITransport {
             }
         }
 
-        private synchronized void onActivityTimerExpiry() {
-            activityTimerTask = null;
+        private void onActivityTimerExpiry() {
             long timeSinceLastActivity = System.currentTimeMillis() - lastActivityTime;
             long timeRemaining = getActivityTimeout() - timeSinceLastActivity;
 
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
LGTM
I went through the explanation, seems it can cause deadlock.
It would have been better if we could have a simple unit test to check this works as expected. This will be useful in regression testing for future changes.

PR #1082, opened 2025-04-17
Rename action META_OCCUPANCY->META

SimonWoolf:
https://github.com/ably/specification/pull/292/commits/35bbeb7f8ed9f384fb7105eac8adab7430254f45

Technically a breaking change. We're not considering it such because `Message.action` was only introduced recently as part of the annotations/materialization work, which isn't yet part of a publicly released feature yet, and isn't added to our [main documentation](https://ably.com/docs/api/realtime-sdk/types#message). So it's just very unlikely that anyone is relying on the action yet. Our inband occupancy events docs still all say to use the message name, which is not changing.

## Summary by CodeRabbit

- **Refactor**
  - Renamed an enum constant from "META_OCCUPANCY" to "META" in the user-facing API.


ttypic (approved):
LGTM

PR #1083, opened 2025-05-13
WIP: liveobjects experimental implementation

ttypic:
Draft implementation to showcase liveobject plugins setup


lib/src/main/java/io/ably/lib/objects/LiveObjects.java
```diff
@@ -0,0 +1,8 @@
+package io.ably.lib.objects;
+
+import io.ably.lib.types.Callback;
+
+public interface LiveObjects {
```
sacOO7, line 5: Why can't we call this singular `LiveObject` instead? 
Like it's holding only one parent `rootObject`

sacOO7: We don't have methods suggesting any plural values as well

ttypic: Don't have any opinion here, both singular and plural are good for me, this PR just a reference for the plugin setup in gradle

sacOO7: Guess it represents type of returned `getObjects` method

PR #1119, opened 2025-07-15
[ECO-5450] fix: async connection state transition side effects

ttypic:
Side effects for close and reconnect were invoked asynchronously on a different thread, so the sequence of operations: `client.close(); client.connect(); channel.attach();` produced unexpected results, with the channel ending up in the initialized state instead of attaching.

To fix this, we now proactively reset the connection and channel states. If `connect()` is called while in a terminal state, we can synchronously clear channel states.

Additionally, we need to handle the case where `close()` is called but the connection has not yet transitioned to the closing state, and `connect()` is invoked immediately afterward. In this situation, it is also safe to detach from channels and clean up their state.


sacOO7 (commented):
not sure why integration tests are failing

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -835,10 +831,16 @@ public synchronized void connect() {
...
+        if (closeRequested || currentState.terminal) {
+            // (RTN11d)
+            reinitializeChannelsAfterReconnect();
+            closeRequested = false;
+        }
         requestState(ConnectionState.connecting);
     }
 
     public void close() {
+        closeRequested = true;
```
sacOO7, line 843: instead of `closeRequested`, can we use existing `cleaningUpAfterEnteringTerminalState = true`

lib/src/main/java/io/ably/lib/transport/ConnectionManager.java
```diff
@@ -835,10 +831,16 @@ public synchronized void connect() {
                 return;
             }
         }
+        if (closeRequested || currentState.terminal) {
```
sacOO7, line 834: Can we update connect method to 
```
    public synchronized void connect() {
        /* connect() is the only action that will bring the ConnectionManager out of a terminal currentState */
        if (closeRequested) { // wait till we reach terminal state, will be notified by onConnectionStateChanged()
          while(!currentState.terminal) wait();
          closeRequested = false;
        }
        if(currentState.terminal || currentState.state == ConnectionState.initialized) {
            try {
                startup(); // this also handles cleanup after terminal state is reached
            } catch(InterruptedException e) {
                Thread.currentThread().interrupt();
                Log.e(TAG, "Failed to start up connection", e);
                return;
            }
        }
        requestState(ConnectionState.connecting);
    }
```
And update startup to 
```
    private synchronized void startup() throws InterruptedException {
       if (currentState.terminal) {
          while (cleaningUpAfterEnteringTerminalState) {
              Log.v(TAG, "Waiting for termination action to clean up handler thread");
              wait();
          }
           // (RTN11d)
           reinitializeChannelsAfterReconnect();
        }

        if (handlerThread == null) {
            (handlerThread = new Thread(new ActionHandler())).start();
            startConnectivityListener();
        } else {
            Log.v(TAG, "`connect()` has been called twice on uninitialized or terminal state");
        }
    }
```
This makes sure `connect` is processed only when terminal state is reached along with processing necessary cleanups. Also, code looks clean and easy to follow wdyt

sacOO7: `startup` method also freezes thread till cleanup is completed.

ttypic: yes, but there is no IO operation going while it's frozen, it can't hang for significant time, just simple sync between two threads

sacOO7 (approved):
LGTM, you can double check for failing integration tests

PR #1120, opened 2025-07-15
[ECO-5076] Implemented code for getRoot method

sacOO7:


live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
```diff
@@ -41,6 +34,11 @@ internal class DefaultLiveObjects(private val channelName: String, internal val
   private val sequentialScope =
     CoroutineScope(Dispatchers.Default.limitedParallelism(1) + CoroutineName(channelName) + SupervisorJob())
 
+  /**
+   * Coroutine scope for handling callbacks asynchronously.
+   */
+  private val callbackScope = CoroutineScope(Dispatchers.Default + CoroutineName("LiveObjectsCallback-$channelName"))
```
ttypic, line 40: I think we can use single scope to manage live objects, I don't think we need two. Scope is mostly to group coroutines together and cancel them at the same time. For sequential execution we just need context, I suggest:

```kotlin
private val scope =
    CoroutineScope(Dispatchers.Default + CoroutineName("LiveObjects-$channelName") + SupervisorJob())
private val sequentialContext =  Dispatchers.Default.limitedParallelism(1)  
// ...

override fun getRootAsync(callback: Callback<LiveMap>) {
    scope.launchWithCallback(callback) { getRootAsync() }
}

private suspend fun getRootAsync(): LiveMap = withContext(sequentialContext) {
       adapter.throwIfInvalidAccessApiConfiguration(channelName)
       objectsManager.ensureSynced(state)
       objectsPool.get(ROOT_OBJECT_ID) as LiveMap
}
```

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
```diff
@@ -94,6 +89,21 @@ internal class DefaultLiveObjects(private val channelName: String, internal val
     TODO("Not yet implemented")
   }
 
+  override fun on(event: ObjectsStateEvent, listener: ObjectsStateChange.Listener): ObjectsSubscription =
+    objectsManager.on(event, listener)
+
+  override fun off(listener: ObjectsStateChange.Listener) = objectsManager.off(listener)
+
+  override fun offAll() = objectsManager.offAll()
```
ttypic, line 97: let's not expose `offAll()` at least for now. `offAll`, `unsubscribeAll` - are "gigantic foot guns" (c) Andy

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
sacOO7: Actually, I was planning to do something like that, but we are already using `sequentialScope` in non-suspend methods `initializeHandlerForIncomingObjectMessages` and `handleStateChange`.

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
sacOO7: Same exists in case of ably-js https://github.com/ably/ably-js/blob/837afe6585f68bd9fabf6bf8fd14b723258e9b30/src/plugins/objects/objects.ts#L187, need to disusss with team yes

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
sacOO7: Added `GlobalCallbackScope` to utils.
This will be shared by
`DefaultLiveObjects`, `DefaultLiveMap` and `DefaultLiveCounter` async public API methods.

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
sacOO7: Also, we have `unsubscribeAll` in case of `object-subscriptions`.  You can check https://github.com/ably/specification/pull/346

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
```diff
@@ -186,22 +188,6 @@ internal class DefaultLiveObjects(internal val channelName: String, internal val
...
-
-    state = newState
-    Log.v(tag, "Objects state changed to: $newState")
-
-    // TODO: Emit state change events
-  }
-
   // Dispose of any resources associated with this LiveObjects instance
   fun dispose(reason: String) {
     val cancellationError = CancellationException("Objects disposed for channel $channelName, reason: $reason")
```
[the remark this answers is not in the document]
sacOO7: Also, problem is `cancel` and `cancelChildren` methods primary accept `CancellationException`

live-objects/src/main/kotlin/io/ably/lib/objects/DefaultLiveObjects.kt
sacOO7: Okay, updated code to have channel/objects lifecycle specific scope -> https://github.com/ably/ably-java/pull/1120/commits/8b5ea817ae433aae00ae2791f2cb7d19c29a8f39

ttypic (approved):

sacOO7:
Closing in favor of https://github.com/ably/ably-java/pull/1138

PR #1149, opened 2025-08-14
chore[live-objects]: change `:java` dependency to `compileOnly` in `live-objects` module

ttypic:
live-objects can be used both with ably-java and ably-android, making dependency `compileOnly` we prevent adding specific module (ably-java or ably-android) to the classpath

## Summary by CodeRabbit

* **Chores**
  * Adjusted an internal module to be compile-time only, reducing runtime dependencies and improving isolation.
  * Enabled that module on the test classpath so tests continue to run unchanged.
  * No changes to functionality, public APIs, or user experience.

### DIFF
--- a/live-objects/build.gradle.kts
+++ b/live-objects/build.gradle.kts
@@ -11,10 +11,11 @@ repositories {
 }
 
 dependencies {
-    implementation(project(":java"))
+    compileOnly(project(":java"))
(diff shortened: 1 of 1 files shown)


sacOO7:
Btw, how do we make sure `ably-java`/`ably-android` version in sync with `liveobjects` module

PR #1152, opened 2025-08-19
fix[live-objects]: replace unsafe array cast with `.map`

ttypic:
`Array<Object>` can't be cast to `Array<ObjectMessages>` even if every element is an instance of `ObjectMessages`, so it's safer to use `map`

## Summary by CodeRabbit

* **Refactor**
  * Improved serialization safety by replacing unchecked array casts with per-element casting and safe iteration.
  * Removed suppression of unchecked-cast warnings for clearer type safety.
  * No changes to public APIs or user-visible behavior; improves internal robustness and maintainability.

### DIFF
--- a/live-objects/src/main/kotlin/io/ably/lib/objects/serialization/DefaultSerialization.kt
+++ b/live-objects/src/main/kotlin/io/ably/lib/objects/serialization/DefaultSerialization.kt
@@ -1,5 +1,3 @@
-@file:Suppress("UNCHECKED_CAST")
-
 package io.ably.lib.objects.serialization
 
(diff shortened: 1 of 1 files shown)


sacOO7 (approved):
lgtm

sacOO7 (commented):
Can you also remove 
```
@file:Suppress("UNCHECKED_CAST")
```
from top of the file

sacOO7 (changes requested):
Can you also remove 
```
@file:Suppress("UNCHECKED_CAST")
```
from top of the file

PR #1154, opened 2025-08-20
fix: add missing ProGuard rules for LiveObjects plugin

ttypic:
Add rules for Gson serializers and reflection used during plugin initialization.


sacOO7 (approved):
LGTM

Issue #1144, opened 2025-08-07
Thread leak if there's no internet connection


mattias-berglund:
If you run ably-java:1.2.54 on a system with a flaky internet connection, ably-java will leak WebSocket-related threads under certain circumstances. Here's a sample program to demonstrate the behavior: the program sets up an AblyRealtime instance and logs the number of WebSocket-related threads an each connection-state change.

```
package se.tremil.ablyleak;

import io.ably.lib.realtime.AblyRealtime;
import io.ably.lib.realtime.Channel;
import io.ably.lib.types.ClientOptions;
import io.ably.lib.types.ErrorInfo;
import io.ably.lib.util.Log;
import java.util.Optional;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class Ablyleak {
    public static void main(String[] args) throws Exception {
        final Logger logger = LoggerFactory.getLogger(Ablyleak.class);

        final ClientOptions clientOptions = new ClientOptions();
        clientOptions.clientId = "********";
        clientOptions.key = "********";
        clientOptions.echoMessages = false;
        clientOptions.logLevel = Log.VERBOSE;
        
        final AblyRealtime ably = new AblyRealtime(clientOptions);
        
        ably.connection.on(csc -> {
            final var webSocketCount = Thread.getAllStackTraces().keySet().stream().map(Thread::getName).filter(s -> s.startsWith("WebSocket")).count();
            final Optional<ErrorInfo> optionalReason = Optional.ofNullable(csc.reason);
            final String errorCode = optionalReason.map(e -> e.code).map(Object::toString).orElse("<n/a>");
            final String errorStatus = optionalReason.map(e -> e.statusCode).map(Object::toString).orElse("<n/a>");
            
            logger.info(String.format("connection state changed from %s to %s, web socket count: %d (code=%s, statusCode=%s)", csc.previous, csc.current, webSocketCount, errorCode, errorStatus));
        });

        final Channel channel = ably.channels.get("remote-punches");
        channel.attach();
    }
}

```

Running this on a system connected to a WiFi-network without internet connection will produce something like this:

```
19:46:00.280 [Thread-0] INFO  (Ablyleak.java:30) - connection state changed from connecting to disconnected, web socket count: 0 (code=80000, statusCode=503)
19:46:13.646 [Thread-0] INFO  (Ablyleak.java:30) - connection state changed from disconnected to connecting, web socket count: 0 (code=<n/a>, statusCode=<n/a>)
... 
[...]

PR #1165, opened 2025-09-24
[ECO-5514] fix: improve WebSocket transport lifecycle and activity management

ttypic:
Resolves https://github.com/ably/ably-java/issues/1144

* Force-cancel hanging socket connections
* Replace synchronized blocks with fine-grained locking using `activityTimerMonitor` to avoid deadlocks.
* Refactor `close()` for safer access to shared fields and handle uninitialized cases gracefully.
* Simplify activity timer logic and ensure consistent disposal of `WebSocketClient` and `WebSocketHandler`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fewer deadlocks/races during connect/close; improved cleanup and warnings for already-closed or uninitialized connections.
  * More robust activity/idle handling to prevent stray timers and improper cancels.

* **Refactor**
  * Enforced single-connect semantics and reduced synchronization hotspots for more reliable lifecycle and activity handling.

* **New Features**
  * Added ability to retrieve auth parameters and check whether a transport is active.

* **Tests**
  * Added unit tests for connect/close and activity-timer scenarios and a test helper to set private fields.

### DIFF
--- a/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
+++ b/lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
@@ -52,7 +52,10 @@ public class WebSocketTransport implements ITransport {
     private ConnectListener connectListener;
     private WebSocketClient webSocketClient;
     private final WebSocketEngine webSocketEngine;
+    private WebSocketHandler webSocketHandler;
     private boolean activityCheckTurnedOff = false;
+
+    private boolean connectHasBeenCalled = false;
 
     /******************
      * protected constructor
@@ -94,8 +97,13 @@ public class WebSocketTransport implements ITransport {
      * ITransport methods
      ******************/
 
+    /**
+     * Connect is called once when we create transport;
+     * after transport is closed, we never call `connect` again
+     */
     @Override
     public void connect(ConnectListener connectListener) {
+        ensureConnectCalledOnce();
         this.connectListener = connectListener;
         try {
             boolean isTls = params.options.tls;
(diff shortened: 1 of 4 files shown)


lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -305,6 +328,9 @@ public void onClose(final int wsCode, final String wsReason) {
             }
             connectListener.onTransportUnavailable(WebSocketTransport.this, reason);
             dispose();
+
+            webSocketClient = null;
+            webSocketHandler = null;
```
sacOO7, line 333: Since we're not using any synchronization primitives, there's a possibility that connect is called externally—setting `webSocketHandler` and `webSocketClient`—while a race condition causes them to be set to null here after that.

sacOO7: Okay got it

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -52,6 +52,7 @@ public class WebSocketTransport implements ITransport {
     private ConnectListener connectListener;
     private WebSocketClient webSocketClient;
     private final WebSocketEngine webSocketEngine;
+    private WebSocketHandler webSocketHandler;
     private boolean activityCheckTurnedOff = false;
```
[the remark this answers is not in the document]
ttypic: we don't really need volatile for this fields, we don't really use them in any synchronization logic

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
ttypic: I put additional checks for connect anyway to ensure that it can be called twice, because it will result in memory leak

lib/src/main/java/io/ably/lib/transport/WebSocketTransport.java
```diff
@@ -217,10 +245,15 @@ class WebSocketHandler implements WebSocketListener {
          * WsClient private members
          ***************************/
 
-        private Timer timer = new Timer();
+        private final Timer timer = new Timer();
         private TimerTask activityTimerTask = null;
         private long lastActivityTime;
```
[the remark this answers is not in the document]
ttypic: it's only used in critical sections, no need volatile here

sacOO7 (approved):
Lgtm

Issue #1167, opened 2025-09-29
Decoding error in the logs when `data` field is empty


ttypic:
If `data` field in message is empty SDK throws decoding exception and log ERROR message, although it's valid payload

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5570) by [Unito](https://www.unito.io)

PR #1168, opened 2025-09-29
fix: handle null check for empty `data` in `Message` decoding logic

ttypic:
Resolves https://github.com/ably/ably-java/issues/1167

Bug only shown in the log, but still annoying 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a rare crash when processing messages with no payload, preventing unnecessary exceptions and improving stability in edge cases.
  * Improved error handling for message payload types to avoid false positives.
  * Previously, some operations could throw an error if a message payload was absent; these cases are now handled gracefully, reducing log noise and avoiding interruptions for end users.


sacOO7 (approved):
Lgtm

PR #1190, opened 2026-02-11
docs: add path-based LiveObjects API proposal

ttypic:
New API for LO

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation for a new path-based LiveObjects API for Java/Kotlin and Python, covering core concepts (path-based access, deferred resolution, atomic creation), a typed PathObject model, type-safe usage patterns, subscription semantics, collection accessors, migration guidance from the current API, usage examples, performance considerations, and a roadmap with open questions.

### DIFF
--- a/liveobjects/PATH_BASED_API_JAVA_PYTHON.md
+++ b/liveobjects/PATH_BASED_API_JAVA_PYTHON.md
@@ -1,0 +1,943 @@
+# LiveObjects Path-Based API Implementation Plan
+
+## Executive Summary
+
+This document outlines the implementation plan for building a **new path-based LiveObjects API from scratch**.
+This is a clean-slate implementation that replaces the current object-manipulation approach with a modern, intuitive path-based design.
+
+**Target SDKs:** Java/Kotlin, Python
+
+---
+
+## Core Architecture
+
+### Fundamental Concepts
+
+**Path-Based Access**
+Objects are accessed via paths from the root, not by creating and managing object instances:
+
+**Java:**
+```java
+// Path to nested value
+StringPathObject colour = myObject.at("shape.colour").asStringPrimitive();
+String value = colour.value(); // "red"
+```
+
+**Python:**
+```python
+# Path to nested value
+colour = my_object.at("shape.colour").as_string_primitive()
+value = colour.value()  # "red"
+```
+
+**Deferred Resolution**
+`PathObject` wraps a path; resolution happens when methods are called:
+
+**Java:**
+```java
+LiveMapPathObject foo = myObject.get("foo").asLiveMap(); // No resolution yet
+foo.set("bar", Primitive.create("baz")); // Resolution happens here
+```
+
+**Python:**
+```python
+foo = my_object.get("foo").as_live_map()  # No resolution yet
+foo.set("bar", Primitive.create("baz"))  # Resolution happens here
+```
+
+**Single Object Per Channel**
+Each channel has one root object (always a LiveMap):
+
+**Java:**
+```java
+LiveMapPathObject myObject = channel.getObject().get(); // The singular object
+```
+
+**Python:**
+```python
+my_object = await channel.object.get()  # The singular object
+```
+
+**Atomic Deep Creation**
+Create nested structures in one operation:
+
+**Java:**
+```java
+myObject.set("shape", LiveMap.create(Map.of(
+  "name", Primitive.create("circle"),
+  "colour", LiveMap.create(Map.of(
+    "border", Primitive.create("red"),
+    "fill", Primitive.create("blue")
+  ))
+)));
+```
+
+**Python:**
+```python
+my_object.set("shape", LiveMap.create({
+    "name": Primitive.create("circle"),
+    "colour": LiveMap.create({
+        "border": Primitive.create("red"),
+        "fill": Primitive.create("blue")
+    })
+}))
+```
+
+### Type System
+
+```
+Value (all types that can be stored)
+├── Primitive (leaf values)
+│   ├── String
+│   ├── Number
+│   ├── Boolean
+│   ├── Binary (byte[])
+└── Live (collaborative types)
+    ├── LiveCounter
+    ├── LiveMap
+    └── LiveList
+```
+
+### Core Types
+
+**PathObject**
+Represents a path to a location in the object tree.
+
+**Java:**
+```java
+public interface PathObject {
+  String path();
+  JsonValue compact() throws AblyException;
+
+  StringPathObject asStringPrimitive();
+  NumberPathObject asNumberPrimitive();
+  BooleanPathObject asBooleanPrimitive();
+  BinaryPathObject asBinaryPrimitive();
+
+  LiveMapPathObject asLiveMap();
+  LiveCounterPathObject asLiveCounter();
+  LiveListPathObject asLiveList();
+
+  Subscription subscribe(ObjectChangeListener listener, SubscriptionOptions options);
+}
+
+public interface StringPathObject extends PathObject {
+  String value() throws AblyException;
+}
+
+public interface NumberPathObject extends PathObject {
+  Double value() throws AblyException;
+}
+
+public interface BooleanPathObject extends PathObject {
+  Boolean value() throws AblyException;
+}
+
+public interface BinaryPathObject extends PathObject {
+  byte[] value() throws AblyException;
+}
+
+public interface LiveMapPathObject extends PathObject {
+  PathObject get(String key);
+  PathObject at(String path);
+  LiveMap instance() throws AblyException;
+
+  void set(String key, Value value, MessageOptions options) throws AblyException;
+  void remove(String key, MessageOptions options) throws AblyException;
+
+  Iterable<Map.Entry<String, PathObject>> entries();
+  Iterable<String> keys();
+  Iterable<PathObject> values();
+  long size();
+}
+
+public interface LiveListPathObject extends PathObject {
+  PathObject get(int index);
+  long size();
+}
+
+public interface LiveCounterPathObject extends PathObject {
+  Double value() throws AblyException;
+  void increment(Number amount, MessageOptions options) throws AblyException;
+  void decrement(Number amount, MessageOptions options) throws AblyException;
+}
+```
+
+**Python:**
+```python
+from abc import ABC, abstractmethod
+from typing import Optional, Iterable, Tuple, Any
+
+class PathObject(ABC):
+    @abstractmethod
+    def path(self) -> str:
+        pass
+
+    @abstractmethod
+    def compact(self) -> Any:
+        pass
+
+    @abstractmethod
+    def as_string_primitive(self) -> 'StringPathObject':
+        pass
+
+    @abstractmethod
+    def as_number_primitive(self) -> 'NumberPathObject':
+        pass
+
+    @abstractmethod
+    def as_boolean_primitive(self) -> 'BooleanPathObject':
+        pass
+
+    @abstractmethod
+    def as_binary_primitive(self) -> 'BinaryPathObject':
+        pass
+
+    @abstractmethod
+    def as_live_map(self) -> 'LiveMapPathObject':
+        pass
+
+    @abstractmethod
+    def as_live_counter(self) -> 'LiveCounterPathObject':
+        pass
+
+    @abstractmethod
+    def as_live_list(self) -> 'LiveListPathObject':
+        pass
+
+    @abstractmethod
+    def subscribe(
+        self,
+        listener: ObjectChangeListener,
+        options: Optional[SubscriptionOptions] = None
+    ) -> Subscription:
+        pass
+
+
+class StringPathObject(PathObject):
+    @abstractmethod
+    def value(self) -> str:
+        pass
+
+
+class NumberPathObject(PathObject):
+    @abstractmethod
+    def value(self) -> float:
+        pass
+
+
+class BooleanPathObject(PathObject):
+    @abstractmethod
+    def value(self) -> bool:
+        pass
+
+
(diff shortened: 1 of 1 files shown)


liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+}
+
+public interface BinaryPathObject extends PathObject {
+  byte[] value() throws AblyException;
+}
+
+public interface LiveMapPathObject extends PathObject {
+  PathObject get(String key);
+  PathObject at(String path);
+  LiveMap instance() throws AblyException;
```
lawrence-forooghian, line 143: `LiveMapInstance`?

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+  void set(String key, Value value, MessageOptions options) throws AblyException;
+  void remove(String key, MessageOptions options) throws AblyException;
+
+  Iterable<Map.Entry<String, PathObject>> entries();
+  Iterable<String> keys();
+  Iterable<PathObject> values();
+  long size();
+}
+
+public interface LiveListPathObject extends PathObject {
```
lawrence-forooghian, line 154: these two also need `instance()`

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+### Core Types
+
+**PathObject**
+Represents a path to a location in the object tree.
+
+**Java:**
+```java
+public interface PathObject {
+  String path();
+  JsonValue compact() throws AblyException;
```
lawrence-forooghian, line 110: this needs updating to represent the new `compact()` / `compactJson()` split in JS

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+public interface BinaryPathObject extends PathObject {
+  byte[] value() throws AblyException;
+}
+
+public interface LiveMapPathObject extends PathObject {
+  PathObject get(String key);
+  PathObject at(String path);
+  LiveMap instance() throws AblyException;
+
+  void set(String key, Value value, MessageOptions options) throws AblyException;
```
lawrence-forooghian, line 145: these `options` don't exist in JS

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+
+  LiveMapPathObject asLiveMap();
+  LiveCounterPathObject asLiveCounter();
+  LiveListPathObject asLiveList();
+
+  Subscription subscribe(ObjectChangeListener listener, SubscriptionOptions options);
+}
+
+public interface StringPathObject extends PathObject {
+  String value() throws AblyException;
```
lawrence-forooghian, line 125: let's make it clear what the failure modes are here and when this throws vs returning `null`:

- nothing at the path: returns `null`?
- resolved value not a string: throws?

lawrence-forooghian:
Wrote up discussion re this in standup today; see https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1778683232335829?thread_ts=1778503442.294519&cid=C09SY1AQGK0

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,987 @@
...
+foo = my_object.get("foo").as_live_map()  # No resolution yet
+foo.set("bar", Primitive.create("baz"))  # Resolution happens here
+```
+
+**Single Object Per Channel**
+Each channel has one root object (always a LiveMap):
+
+**Java:**
+```java
+LiveMapPathObject myObject = channel.getObject().get(); // The singular object
```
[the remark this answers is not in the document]
sacOO7: Seems, at the bottom, we agree to use `LiveMapPathObject myObject = channel.object().get()` instead of
`channel.object().get()`

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+
+### Fundamental Concepts
+
+**Path-Based Access**
+Objects are accessed via paths from the root, not by creating and managing object instances:
+
+**Java:**
+```java
+// Path to nested value
+StringPathObject colour = myObject.at("shape.colour").asStringPrimitive();
```
sacOO7, line 22: `PathObject` as a concept is making sure we handle dynamic instances at given path, feels odd when we apply concrete typing here,
Maybe it would make sense to have `type` specific instances, i.e. `StringInstance`, `LiveMapInstance` etc wdyt

sacOO7: types for given `instance` remains same throughout the lifetime, in case of `PathObject`, type at given path can change

sacOO7: Though we need to give a thought about we can make `PathObjects` strongly typed.
This should be consistent across SDKs : )

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+value = colour.value()  # "red"
+```
+
+**Deferred Resolution**
+`PathObject` wraps a path; resolution happens when methods are called:
+
+**Java:**
+```java
+LiveMapPathObject foo = myObject.get("foo").asLiveMap(); // No resolution yet
+foo.set("bar", Primitive.create("baz")); // Resolution happens here
```
sacOO7, line 39: We currently don't have `Primitive` as explicit type anywhere in the spec, maybe there's need to define it in the spec, so we can publicly expose and use the same.
This will be consistent across SDKs

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+colour = my_object.at("shape.colour").as_string_primitive()
+value = colour.value()  # "red"
+```
+
+**Deferred Resolution**
+`PathObject` wraps a path; resolution happens when methods are called:
+
+**Java:**
+```java
+LiveMapPathObject foo = myObject.get("foo").asLiveMap(); // No resolution yet
```
sacOO7, line 38: Not sure if this should be `LiveMapPathObject`. Puts developer in a limbo when someone tries to set `key/value` on `foo` but code throws error that `foo` isn't a `LiveMap` or it's of different type. 
Since, `PathObject` itself represents dynamic types as runtime, defining types for it will be contradiction at multiple levels.

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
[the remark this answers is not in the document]
sacOO7: We previously had `channel.getObjects().getRoot()`, this used to make sense, but with the new public API change where `getObjects` is now just `getObject` and `getRoot` changed to `get`, it would make sense to update `java` API to make it more sensible to read i.e. `channel.object().get()`.
In case of `kotlin`, it will be `channel.realtimeObject.get()` or `channel.Object.get()`, since uppercase `Object` is not a reserved keyword in Kotlin.

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+```python
+my_object = await channel.object.get()  # The singular object
+```
+
+**Atomic Deep Creation**
+Create nested structures in one operation:
+
+**Java:**
+```java
+myObject.set("shape", LiveMap.create(Map.of(
```
sacOO7, line 66: [As per spec](https://sdk.ably.com/builds/ably/specification/main/objects-features/#:~:text=class%20LiveCounter%20extends%20LiveObject%3A%20//%20RTLC*%2C%20RTLC1%2C%20internal) `LiveMap`/`LiveCounter` is now a internal type. This is also true for `ably-js`, but `ably-js` exports following types
1. `LiveCounterValueType` ->  `LiveCounter`
2. `LiveMapValueType` -> `LiveMap`

Need to check how we can keep this naming consistent across `ably-js` and other SDKs.
As per public docs, we do expose `LiveCounter`, `LiveMap` -> https://ably.com/docs/liveobjects/concepts/path-object#typing

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
sacOO7: I don't think we can represent `compact` for `ably-java`. In case of `ably-js`, it returns anonomous object  
(native js object) without any concrete typing. In case of other SDKs like java/swift, it would be impossible to handle anonomous typing specifically needed by `compact` / `compactJson` methods

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+**PathObject**
+Represents a path to a location in the object tree.
+
+**Java:**
+```java
+public interface PathObject {
+  String path();
+  JsonValue compact() throws AblyException;
+
+  StringPathObject asStringPrimitive();
```
sacOO7, line 112: Similar to comment -> https://github.com/ably/ably-java/pull/1190/changes#r3323890816, feels odd to return specific types for `PathObject`, applicable for instances where type for given instance will remain same for given lifetime.

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
sacOO7: Maybe, it's possible to return `Map<String, object>` where `object` can be `Primitive`, `LiveMap` or `LiveCounter`. I feel we should have superclass for `Primitive`, `LiveMap` and `LiveCounter` or something like it to make this work. But it's surely tricky. 
Value returned by `compact` method can be printed as `json` tree in `ably-js`. Also, each value is directly accessible from root object itself, like `value.key1.key2` etc

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
sacOO7, line 143:
```suggestion
  LiveMapInstance instance() throws AblyException;
```

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
sacOO7: `LiveListPathObject` doesn't exist

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+
+### Getting Started
+
+**Java:**
+```java
+// Get channel
+RealtimeChannel channel = client.channels.get("my-channel");
+
+// Get root object (waits for sync)
+LiveMapPathObject myObject = channel.getObject().get();
```
sacOO7, line 328: Root object will always be `LiveMapPathObject`, type of root will never change.
But `PathObjects` at child locations, their types can always change dynamically.

sacOO7: It's not frequent that types can change `dynamically` for children, still need to consider cases where we store type in a variable and then try performing some action on it when it can possible change internally

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+
+// Access nested primitive value
+StringPathObject name = myObject.at("user.name").asStringPrimitive();
+String nameValue = name.value(); // "Alice"
+
+// Access nested counter
+LiveCounterPathObject visits = myObject.at("stats.visits").asLiveCounter();
+Double visitsValue = visits.value(); // 42.0
```
sacOO7, line 336: `visits.value` can return `string` or `livemap` based on underlying changed type

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
sacOO7: Maybe this should be called `RootPathObject` instead

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+visits_value = visits.value()  # 42.0
+```
+
+### Creating Objects
+
+**Java:**
+```java
+// Create simple map
+myObject.set("user", LiveMap.create(Map.of(
+    "name", Primitive.create("Alice"),
```
sacOO7, line 362: Need explicit `Primitive` type in the spec, maybe should be called `LivePrimitive`

liveobjects/PATH_BASED_API_JAVA_PYTHON.md
```diff
@@ -0,0 +1,943 @@
...
+This provides type safety and prevents invalid operations (e.g., can't call `increment()` on a string).
+
+### 2. Explicit Type Assertions
+Navigation returns a generic `PathObject`, which must be explicitly cast to the expected type:
+```java
+PathObject generic = myObject.get("user");
+LiveMapPathObject user = generic.asLiveMap(); // Explicit cast
+```
+
+This makes the developer's expectations clear and fails fast if the type doesn't match.
```
sacOO7, line 887: This defeats purpose of updating reading/writing dynamic types at given paths.
Static types doesn't ensure we are updating or reading right values.

PR #1207, opened 2026-05-01
fix: update `testAppSpec.json` to the right format

ttypic:
Right now ably-java has copy of common's test app spec with some modifications, but the format may change we shouldn't rely on it, and use actual format instead

for now as a quick fix we update format for presence

## Summary by CodeRabbit

* **Tests**
  * Consolidated duplicate test data entries to improve test resource efficiency and reduce redundancy in test configurations.


sacOO7 (approved):
LGTM

PR #1209, opened 2026-05-15
[AIT-767] uts: live objects ground work

ttypic:
This PR includes:

- MockHttpClient
- MockWebSocketClient
- Clock interface that allows fake timers
- Skill to translate UTS tests into kotlin 

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit-test infrastructure, deterministic virtual-clock and timer helpers, mock WebSocket/HTTP transports, test client factories, and new UTS test suites covering connection-recovery and realtime behavior.

* **Chores**
  * Introduced a Clock/AblyTimer abstraction and wiring across the SDK to centralize time handling and improve testability.

* **CI**
  * Test pipeline updated to run the new UTS test module.

* **Documentation**
  * Added end-to-end guidance for translating pseudocode specs into runnable Kotlin tests and running them.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-java/pull/1209?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,323 @@
+---
+description: "Translate a UTS pseudocode test spec into Kotlin tests in the uts module. Usage: /uts-to-kotlin <path-to-spec-file>"
+allowed-tools: Bash, Read, Edit, Write
+---
+
+You are translating a UTS pseudocode test spec file into a runnable Kotlin test in the `uts` module. Follow these steps in order.
+
+---
+
```
paddybyers, line 50: Should there be a reference to https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md ?

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,323 @@
...
+## Step 1 — Read the spec
+
+Read the file at `$ARGUMENTS`. Identify:
+- All test cases (each has an ID like `RTN4a`, `RSC1`, etc. and a description)
```
paddybyers: There's a structured ID for each test, eg `realtime/unit/RSA4c2/callback-error-connecting-disconnected-0` (eg https://github.com/ably/specification/blob/main/uts/realtime/unit/auth/auth_callback_errors_test.md#rsa4c2---authcallback-error-during-connecting-transitions-to-disconnected)

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: Added

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: Yeah it's weird, because later in the doc it uses structured ID, updated

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -1036,7 +1038,7 @@ public void onAuthError(ErrorInfo err) {
             clearTokenDetails();
     }
 
-    public static long timestamp() { return System.currentTimeMillis(); }
+    public long timestamp() { return clock.currentTimeMillis(); }
```
sacOO7, line 1041: **ABI break — `timestamp()` changed from `static` to instance method.**

```diff
- public static long timestamp() { return System.currentTimeMillis(); }
+ public long timestamp() { return clock.currentTimeMillis(); }
```

`Auth` is `public class` (see `Auth.java:41`), so this is part of the published API. Any external caller compiled against the previous version (downstream apps, `ably-android`, etc.) will get `NoSuchMethodError` at runtime — and source callers using `Auth.timestamp()` will fail to compile.

This change isn't required for the FakeClock test path either: `FakeClock` plugs in via `DebugOptions.clock`, which `Auth` already reads through `SystemClock.clockFrom(options)` for internal use.

**Suggested fix — preserve the static API for backwards compatibility:**

```java
/** @deprecated Use the instance method on an Auth instance. Retained for ABI compatibility. */
@Deprecated
public static long timestamp() { return System.currentTimeMillis(); }

public long instanceTimestamp() { return clock.currentTimeMillis(); }
```

Or, if you want to commit to the rename, audit downstream consumers (`ably-android` etc.) and call it out in the CHANGELOG as a breaking change.

lib/src/main/java/io/ably/lib/rest/Auth.java
```diff
@@ -1050,6 +1052,8 @@ public void onAuthError(ErrorInfo err) {
      */
     Auth(AblyBase ably, ClientOptions options) throws AblyException {
         this.ably = ably;
+        this.clock = SystemClock.clockFrom(options);
+        this.nanoTimeDelta = clock.currentTimeMillis() - System.nanoTime()/(1000*1000);
```
sacOO7: **`nanoTimeDelta` mixes virtual `Clock` time with real `System.nanoTime()`.**

```java
this.nanoTimeDelta = clock.currentTimeMillis() - System.nanoTime()/(1000*1000);
```

When `clock` is a `FakeClock` initialised at `t=0`, this computes `0 - System.nanoTime()/1e6` — a large negative value. The drift-detection at line ~924 recomputes the same formula and checks `Math.abs(currentNanoTimeDelta - oldNanoTimeDelta) > 500`. Virtual time stays at 0 while real time advances, so the drift exceeds 500 ms quickly and `Auth` will incorrectly conclude the device clock changed.

Dormant for RTN16 tests (no token-query path), but the first auth UTS test using FakeClock will hit it.

**Suggested fix:** add a `nanoTimeMillis()` (or similar) method to the `Clock` SPI, implement consistently for both `SystemClock` (delegate to `System.nanoTime()/1e6`) and `FakeClock` (return current virtual time minus a constructor-captured baseline). Then change both call sites here to use it, so the two ends of the drift subtraction always come from the same clock.

```java
this.nanoTimeDelta = clock.currentTimeMillis() - clock.nanoTimeMillis();
```

uts/src/test/kotlin/io/ably/lib/uts/infra/FakeClock.kt
```diff
@@ -0,0 +1,86 @@
...
+class FakeClock(initialTimeMs: Long = 0L) : Clock {
+    @Volatile private var time = initialTimeMs
+    private val timers = mutableMapOf<String, FakeNamedTimer>()
+    private val waiters = mutableListOf<Waiter>()
+
+    override fun currentTimeMillis() = time
+
+    override fun newTimer(name: String): NamedTimer {
+        val t = FakeNamedTimer(name)
+        timers[name] = t
```
sacOO7, line 26: **Name-keyed map silently clobbers concurrent timers.**

```kotlin
override fun newTimer(name: String): NamedTimer {
    val t = FakeNamedTimer(name)
    timers[name] = t   // ← overwrites any existing timer with this name
    return t
}
```

The SDK creates a new timer per attach/reattach/detach attempt (`ChannelBase.java:543,607,646`). If two channels are attached concurrently, the second `clock.newTimer("attach-timer")` overwrites the first in this map. The SDK's own `NamedTimer` reference still works, but `FakeClock.advance()` only iterates `timers.values` — the orphaned timer's tasks are silently never advanced.

The current `ConnectionRecoveryTest` doesn't expose this because it only ever has one of each timer alive at a time, but the first multi-channel test will silently fail to advance some retry/timeout schedules.

**Suggested fix — store timers in a list, not a map. The `name` is for diagnostic logging, not identity:**

```kotlin
private val timers = mutableListOf<FakeNamedTimer>()

override fun newTimer(name: String): NamedTimer =
    FakeNamedTimer(name).also { synchronized(timers) { timers += it } }

fun advance(ms: Long) {
    time += ms
    synchronized(timers) { timers.toList() }.forEach { it.fireDue(time) }
    // ...
}
```

uts/src/test/kotlin/io/ably/lib/test/mock/FakeClock.kt
```diff
@@ -0,0 +1,86 @@
...
+        timers[name] = t
+        return t
+    }
+
+    @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+    override fun waitOn(target: Any, timeout: Long) {
+        synchronized(waiters) {
+            waiters.add(Waiter(target as Object, time + timeout))
+        }
+        (target as Object).wait()
```
sacOO7: **`waitOn` calls `Object.wait()` with no timeout — diverges from `SystemClock.waitOn` semantics.**

```kotlin
override fun waitOn(target: Any, timeout: Long) {
    synchronized(waiters) {
        waiters.add(Waiter(target as Object, time + timeout))
    }
    (target as Object).wait()   // ← no timeout passed
}
```

`SystemClock.waitOn` does `target.wait(timeout)` (returns after `timeout` ms regardless of whether `notify` fires). This implementation waits **indefinitely** — if `fakeClock.advance(...)` is never called past the deadline, the SDK thread hangs forever. That's a behavioural divergence from `SystemClock` and a footgun for any test that uses a blocking `Future.get(timeout)` without driving the clock.

**Suggested fix — preserve the timeout as a safety fallback so the contract matches `SystemClock`:**

```kotlin
(target as Object).wait(timeout)
```

The virtual `notifyAll()` from `advance()` will usually fire first in tests that drive the clock properly; the real-time fallback is a safety net only.

(Note: this is distinct from CodeRabbit's monitor-contract comment on line 34 — the caller `HttpScheduler.AsyncRequest.get` does invoke `clock.waitOn(this, remaining)` from inside `synchronized(this)`, so the monitor *is* held. The issue here is the missing timeout argument.)

uts/src/test/kotlin/io/ably/lib/uts/infra/FakeClock.kt
```diff
@@ -0,0 +1,86 @@
...
+            return TimerInstance { task.cancel(); pending -= s }
+        }
+
+        override fun cancel() {
+            pending.forEach { it.task.cancel() }
+            pending.clear()
+        }
+
+        fun fireDue(now: Long) {
+            val due = pending.filter { it.fireAt <= now }
```
sacOO7, line 78: **Two latent issues in `fireDue` worth addressing together:**

```kotlin
fun fireDue(now: Long) {
    val due = pending.filter { it.fireAt <= now }
    pending -= due.toSet()
    due.forEach { it.task.run() }
}
```

1. **No chain-scheduling fan-out.** If a fired task calls `schedule(_, 0)` to defer-to-next-tick, the new task is added to `pending` after `due` was filtered — it sits until the next `advance()`. Real `java.util.Timer` fires chain-scheduled tasks promptly within the same tick.

2. **Cancelled tasks still fire.** `TimerTask.cancel()` called directly on the SDK's reference doesn't remove the task from `pending` — `fireDue` will still call `task.run()` on it. Real `Timer` skips cancelled tasks at fire time.

**Suggested fix:**

```kotlin
fun fireDue(now: Long) {
    while (true) {
        val due = synchronized(this) {
            val d = pending.filter { it.fireAt <= now && it.task.state == TimerTask.SCHEDULED }
            pending -= d.toSet()
            d
        }
        if (due.isEmpty()) return
        due.forEach { it.task.run() }   // may schedule more tasks; loop picks them up
    }
}
```

(`TimerTask.state` isn't public — substitute a `cancelled: AtomicBoolean` on the `Scheduled` wrapper that the `TimerInstance.cancel` lambda sets.)

uts/src/test/kotlin/io/ably/lib/uts/infra/DefaultPendingConnection.kt
```diff
@@ -0,0 +1,34 @@
...
+        listener.onOpen()
+        onConnected(listener)
+    }
+
+    override fun respondWithSuccess(message: ProtocolMessage) {
+        listener.onOpen()
+        onConnected(listener)
+        // Async delivery per spec: the library must store the WS reference before processing CONNECTED.
+        val encoded = Serialisation.gson.toJson(message)
+        Thread { listener.onMessage(encoded) }.apply { isDaemon = true }.start()
```
sacOO7: **`Thread { … }.start()` per CONNECTED is the anti-pattern the spec explicitly calls out.**

```kotlin
Thread { listener.onMessage(encoded) }.apply { isDaemon = true }.start()
```

Per `specification/uts/docs/writing-derived-tests.md` §"No real timers in unit tests":

> For waiting on async event delivery (mock message propagation, promise settlement): yield to the event loop with a zero-delay mechanism like `setImmediate`, `process.nextTick`, or equivalent. Define a `flushAsync()` helper and use it everywhere instead of `setTimeout(resolve, N)`.

Practical consequences here:
- Each test that simulates reconnect (e.g. `RTN16k`) creates multiple daemon threads, all lingering until they complete.
- No ordering guarantee between "CONNECTED #1 delivered" and the test's next action (e.g. `simulateDisconnect()`). The SDK happens to serialise inbound on its action handler, but the test-side race is still real.

**Suggested fix:** use a single-thread daemon `ExecutorService` shared across the mock (cheap, deterministic ordering), or expose a `flushAsync()` / `awaitDeliveryQuiescent()` helper that tests call after mock interactions:

```kotlin
private val deliveryExecutor = Executors.newSingleThreadExecutor { r ->
    Thread(r, "mock-ws-delivery").apply { isDaemon = true }
}

override fun respondWithSuccess(message: ProtocolMessage) {
    listener.onOpen()
    onConnected(listener)
    val encoded = Serialisation.gson.toJson(message)
    deliveryExecutor.submit { listener.onMessage(encoded) }
}
```

uts/src/test/kotlin/io/ably/lib/test/mock/MockWebSocketEngineFactory.kt
```diff
@@ -0,0 +1,58 @@
...
+    private val onClientClose: (Int, String) -> Unit,
+) : WebSocketClient {
+    override fun connect() {
+        val uri = URI(url)
+        val tls = uri.scheme == "wss"
+        val port = if (uri.port == -1) (if (tls) 443 else 80) else uri.port
+        onConnect(DefaultPendingConnection(uri.host, port, tls, parseQueryString(uri.rawQuery), listener, onConnected))
+    }
+
+    override fun close() {}
```
sacOO7: **Parameter-less `close()` is a no-op, and the parameterised one doesn't drive `listener.onClose`.**

```kotlin
override fun close() {}
override fun close(code: Int, reason: String) { onClientClose(code, reason) }
```

Neither variant calls `listener.onClose(code, reason)`, so when the SDK calls `connection.close()` the state machine never observes the close from the transport side. This is why `RTN16g2` has to do:

```kotlin
client.connection.close()
…
mock.simulateDisconnect()                            // ← only way to drive CLOSED
awaitState(client, ConnectionState.closed)
```

A correct mock would let `client.connection.close()` drive the SDK to CLOSED via the natural close path: SDK calls transport `close(code, reason)` → mock echoes `listener.onClose(code, reason)` → SDK transitions.

**Suggested fix:**

```kotlin
override fun close() = close(1000, "Normal closure")
override fun close(code: Int, reason: String) {
    onClientClose(code, reason)
    listener.onClose(code, reason)   // drive the SDK
}
```

If any test ever needs to inspect the close frame *without* auto-acking it (rare), add a config flag to suppress the listener echo.

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+**Env-gated skip (preferred)** — test contains spec-correct assertions but is skipped by default:
+
+```kotlin
+/**
+ * @UTS realtime/unit/RSA4c2/callback-error-connecting-disconnected-0
+ */
+@Test
+fun `RSA4c2 - callback error connecting disconnected`() = runTest {
+    // DEVIATION: see deviations.md
+    if (System.getenv("RUN_DEVIATIONS") != null) return@runTest
```
sacOO7: **Polarity is backwards — this example skips when `RUN_DEVIATIONS` is set, not when it's unset.**

```kotlin
if (System.getenv("RUN_DEVIATIONS") != null) return@runTest
```

The surrounding text on line 326 says *"test contains spec-correct assertions but is **skipped by default**"*, which means:
- Default (`RUN_DEVIATIONS` unset) → **skip**
- `RUN_DEVIATIONS=1` set → **run** the spec-correct assertions and prove the deviation exists

As written, this example does the opposite. A future translator following the SKILL verbatim will produce backwards deviation tests that pass by default *because they're skipping the assertion when they should be running it*.

Note: the actual test at `ConnectionRecoveryTest.kt:304` gets it right (`if (… != null) { assertEquals(...) }` — assertion only runs when env var is set). Only the SKILL example is wrong.

**Suggested fix:**

```kotlin
if (System.getenv("RUN_DEVIATIONS") == null) return@runTest
```

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+    awaitChannelState(testChannel, ChannelState.attached)
+
+    // Verify msgSerial was initialized from the recoveryKey (RTN16f).
+    // SDK deviation: ConnectionManager.onConnected() resets msgSerial = 0 when connection.id == null
+    // (fresh client, ConnectionManager.java:1316), even when using the recover option.
+    // Spec requires msgSerial to be preserved from the recovery key (42). See deviations.md.
+    val currentRecoveryKey = RecoveryKeyContext.decode(client.connection.createRecoveryKey()!!)
+    assertNotNull(currentRecoveryKey)
+    if (System.getenv("RUN_DEVIATIONS") != null) {
+      assertEquals(42L, currentRecoveryKey.msgSerial)
```
sacOO7, line 308: **RTN16f deviation assertion doesn't verify what the spec actually requires.**

The spec (`connection_recovery_test.md` §RTN16f, lines 437-458) requires the test to **publish a message** after recovery, **capture the outgoing `MESSAGE` frame**, and assert:

```pseudo
ASSERT message_frame.message.msgSerial == 42
```

The current test only re-reads `createRecoveryKey()` and inspects the in-memory `msgSerial` field — it never exercises the publish flow. A future SDK fix that preserves the recovery-key `msgSerial` *but still resets the outgoing-frame `msgSerial`* would pass this test, which defeats the point of the deviation gate.

(This is related to but distinct from CodeRabbit's vacuous-pass comment on line 306 — that's about the default path; this is about the deviation path not asserting the right thing.)

**Suggested fix — inside the `RUN_DEVIATIONS` gate, do the spec's publish + capture flow:**

```kotlin
if (System.getenv("RUN_DEVIATIONS") != null) {
    testChannel.publish("event", "data")
    val msg = mock.awaitNextMessageFromClient()
    assertEquals(ProtocolMessage.Action.message, msg.action)
    assertEquals(42L, msg.msgSerial)
}
```

Also record this in `uts/src/test/kotlin/io/ably/lib/deviations.md` — it's currently referenced (line 301) but the file doesn't exist.

sacOO7 (commented):
Lead-dev review of the UTS groundwork (cross-checked against `specification/uts` and the SDK changes in this PR). The overall direction — `Clock`/`NamedTimer` SPI, `DebugOptions` injection points, mock WS+HTTP engines, fake clock, and a `uts-to-kotlin` skill — is sound, and the RTN16 translation faithfully matches the spec sections that exist. Comments below are correctness or robustness issues; a couple are blockers.

**Strategic findings that don't belong on a single line:**

- **`deviations.md` is referenced but missing.** `ConnectionRecoveryTest.kt:301` and the SKILL both point at `uts/src/test/kotlin/io/ably/lib/deviations.md`. The file doesn't exist. RTN16f is a real deviation that should be recorded there.
- **Spec coverage gap is upstream, not here.** The spec header at `connection_recovery_test.md` lists `RTN16d` / `RTN16i` / `RTN16l`, but only `RTN16i` is actually encoded (inline within `RTN16j`); `RTN16d` and `RTN16l` have no dedicated sections in the spec itself. The translation here is faithful — the gap should be filed against `ably/specification`.
- **No `@AfterEach` cleanup.** `RTN16g2` creates three clients and only closes the last one on the success path; if any earlier assertion throws, all three leak their internal Timer threads into subsequent tests. Recommend a tracked-client `@AfterEach` (or a `useClient { … }` helper).
- **`useJUnitPlatform()` + bare `kotlin("test")`** in `uts/build.gradle.kts`. Kotlin 2.1.10 + Gradle variant metadata *should* resolve this to `kotlin-test-junit5` automatically, but the failure mode if it doesn't is "0 tests executed, build green". Recommend pinning explicitly: `testImplementation(kotlin("test-junit5"))`, then verify with `./gradlew :uts:test -i` that tests actually execute.

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+      install(mock)
+    }
+
+    client.connect()
+    awaitState(client, ConnectionState.connected)
+
+    assertEquals(ConnectionState.connected, client.connection.state)
+    assertEquals("fresh-conn", client.connection.id)
+    assertEquals("fresh-key", client.connection.key)
+    assertNull(capturedQueryParams!!["recover"])
```
sacOO7, line 348: **RTN16f1 — missing log-handler assertion called for by the spec.**

The spec has an explicit implementation note at `connection_recovery_test.md:533-536`:

> Implementations should verify this by capturing log output (e.g., via a log handler) and asserting that an error-level log message was emitted mentioning the malformed recovery key.

The current test asserts the connection succeeds and that no `recover` / `resume` query param was sent, but doesn't verify the log emission. Confirmed `RecoveryKeyContext.decode` does `Log.e(TAG, …)` on `JsonSyntaxException`, so a handler attached via `ClientOptions.logHandler` would observe it.

**Suggested fix — capture log events and assert one was emitted at error level mentioning the malformed key:**

```kotlin
val logEvents = mutableListOf<Pair<Int, String>>()
val client = TestRealtimeClient {
    autoConnect = false
    recover = "this-is-not-valid-json!!!"
    logHandler = Log.LogHandler { sev, _, msg, _ -> logEvents += sev to msg }
    install(mock)
}
…
assertTrue(
    logEvents.any { (sev, msg) -> sev >= Log.ERROR && msg.contains("recovery", ignoreCase = true) },
    "Expected an error log mentioning the malformed recovery key"
)
```

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+      autoConnect = false
+      recover = recoveryKeyJson
+      install(mock)
+    }
+
+    client.connect()
+    awaitState(client, ConnectionState.connected)
+
+    mock.simulateDisconnect()
+    awaitState(client, ConnectionState.connected)
```
sacOO7, line 249: **RTN16k — missing assertion on the SDK's post-reconnect state.**

The mock returns `connectionKey: "resumed-key"` on the second CONNECTED (line 224). The test verifies the *outgoing* `resume` query param (line 250) but never checks what the SDK *applied* on receiving the second CONNECTED. A bug where the SDK keeps the old key (or fails to update `connection.id`) would pass this test silently.

The SDK updates these from `connectionDetails` at `ConnectionManager.java:1319,1323`:
```java
connection.id = message.connectionId;
connection.key = connectionDetails.connectionKey; // RTN16d
```

**Suggested fix — add after the second `awaitState(connected)`:**

```kotlin
assertEquals("resumed-key", client.connection.key)
assertEquals("recovered-conn-id", client.connection.id)
```

This also gives partial coverage of RTN16d (which has no dedicated section in the spec).

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
```diff
@@ -0,0 +1,415 @@
...
+          connectionStateTtl = 800L
+        }
+      })
+    }
+
+    clientSuspended.connect()
+    awaitState(clientSuspended, ConnectionState.connected)
+
+    mockSuspended.simulateDisconnect()
+    awaitState(clientSuspended, ConnectionState.disconnected)
```
sacOO7: **Three related issues in `RTN16g2`'s SUSPENDED branch:**

1. **Transient-state observation.** `awaitState(disconnected)` on line 192 is exactly what `writing-test-specs.md` §"Verifying Transient States (Record-and-Verify Pattern)" calls out as an anti-pattern — DISCONNECTED is supposed to be transient (RTN15a immediate-reconnect). It's safe today only because FakeClock gates the SDK's retry timer (`ConnectionManager.java:316,912`); any future SDK change introducing an immediate-reconnect path on transport failure will make this race-y. Use record-and-verify instead.

2. **`refuseJob` is launched *after* `simulateDisconnect()` and after `awaitState(disconnected)`.** Under FakeClock this is structurally clean (the SDK can't move until the test advances time), but moving the launch *before* the disconnect makes the ordering self-evident and removes the forward-compatibility hazard above.

3. **Client leaks.** This test creates three clients (`client`, `clientFailed`, `clientSuspended`) but only the last is closed on the success path. If any assertion above throws, all three leak their internal Timer threads into subsequent tests.

**Suggested fix — tracked-list `@AfterEach` + record-and-verify pattern:**

```kotlin
private val tracked = mutableListOf<AblyRealtime>()
@AfterEach fun tearDown() = tracked.forEach { runCatching { it.close() } }

// in the test:
tracked += clientSuspended
val stateChanges = mutableListOf<ConnectionState>()
clientSuspended.connection.on { stateChanges += it.current }

// start the refuse job FIRST
val refuseJob = launch { /* ... */ }
mockSuspended.simulateDisconnect()
awaitState(clientSuspended, ConnectionState.suspended)
refuseJob.cancel()

assertContainsInOrder(stateChanges,
    ConnectionState.connected,
    ConnectionState.disconnected,
    ConnectionState.suspended)
```

Apply the same `tracked +=` pattern to `client` and `clientFailed` earlier in the test.

lib/src/main/java/io/ably/lib/rest/Auth.java
sacOO7: We can't take a chance on breaking public API changes

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
sacOO7: Seems, we should explicitly call out these type of cases in `.claude/skills/uts-to-kotlin/SKILL.md`
Maybe, it's because `writing-test-specs.md` is not correctly loaded as a part of the skill.

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
+---
+description: "Translate a UTS pseudocode test spec into Kotlin tests in the uts module. Usage: /uts-to-kotlin <path-to-spec-file>"
+allowed-tools: Bash, Read, Edit, Write
+---
+
+Translate the UTS pseudocode test spec at `$ARGUMENTS` into a runnable Kotlin test in the `uts` module.
+
+Reference: [Writing Derived Tests](https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md)
```
sacOO7:
```suggestion
Reference: [Writing Derived Tests](https://raw.githubusercontent.com/ably/specification/refs/heads/main/uts/docs/writing-derived-tests.md)
```
It's easier for LLM to fetch/understand raw markdown than HTML page

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
+---
+description: "Translate a UTS pseudocode test spec into Kotlin tests in the uts module. Usage: /uts-to-kotlin <path-to-spec-file>"
+allowed-tools: Bash, Read, Edit, Write
+---
+
+Translate the UTS pseudocode test spec at `$ARGUMENTS` into a runnable Kotlin test in the `uts` module.
```
sacOO7, line 6: If `$ARGUMENTS` not provided, skill should either explicitly prompt for `$ARGUMENTS` with format of the give file.

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: When skill is loaded, it doesn't necessarily load `Reference` markdown. 
It's better we make fetch mandatory.


> *Note* - Always fetch [Writing Derived Tests](https://raw.githubusercontent.com/ably/specification/refs/heads/main/uts/docs/writing-derived-tests.md) as a part of SKILL load.

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Also, if required/sufficient information is not available in `$ARGUMENTS`, then skill should log error and return early

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+Read ALL of these before generating any code (you need exact method signatures):
+
+```
+uts/src/test/kotlin/io/ably/lib/ClientFactories.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/MockWebSocket.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/MockHttpClient.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/PendingRequest.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/PendingConnection.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/FakeClock.kt
+uts/src/test/kotlin/io/ably/lib/test/mock/MockEvent.kt
```
sacOO7: Can't we put all of those in single `uts.infra` package? We need to update skill everytime infrastructure changes wdyt

sacOO7: or atleast include `uts.infra.*`/ uts/src/test/kotlin/io/ably/lib/test/mock/*` package

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Yes, as per https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md#3-deviation-test-patterns, this makes sense.

sacOO7: Can you double check if this was an intentional change?

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+**Adapted assertion** — when you still want to assert on the SDK's actual behaviour to prevent regressions:
+
+```kotlin
+// DEVIATION: spec requires error code 40106, SDK returns 40160 — see deviations.md
+assertEquals(40160, error.errorInfo.code)
+```
+
+**Never use the accommodate-both pattern** (accept either spec or SDK behaviour). Every test must assert either spec behaviour or the SDK's actual behaviour — never both at once.
```
sacOO7, line 375: I don't like idea of accommodating `SDK behaviour` when we accept `UTS spec` as source of truth. 
We should only accept `UTS spec behaviour`
Maybe this should explicitly go into `deviations.md`. 
Not sure why we agreed on it here -> https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md#3-deviation-test-patterns:~:text=Avoid%20the%20accommodate%2Dboth%20pattern

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Deviations file
+
+Append to `uts/src/test/kotlin/io/ably/lib/deviations.md`. Each entry needs:
```
sacOO7, line 379:
```suggestion
Deviation file should be created at `uts/src/test/kotlin/io/ably/lib/deviations.md` ( Required for `Evaluation` phase). Each entry is appended with:
```

lib/src/main/java/io/ably/lib/util/NamedTimer.java
```diff
@@ -0,0 +1,8 @@
+package io.ably.lib.util;
+
+import java.util.TimerTask;
+
+public interface NamedTimer {
```
sacOO7:
```suggestion
public interface AblyTimer {
```
Since, this is `Ably` SDK specific timer, would make sense to rename it.
This also conveys that it's a custom timer with ably requirement specific functionality.

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livecounter/DefaultLiveCounter.kt
```diff
@@ -22,7 +23,7 @@ import kotlinx.coroutines.runBlocking
 internal class DefaultLiveCounter private constructor(
   objectId: String,
   private val realtimeObjects: DefaultRealtimeObjects,
-) : LiveCounter, BaseRealtimeObject(objectId, ObjectType.Counter) {
+) : LiveCounter, BaseRealtimeObject(objectId, ObjectType.Counter, SystemClock.clockFrom(realtimeObjects.adapter.clientOptions)) {
```
sacOO7: Can't we add extension method to `SystemClock`, that builds clock from given `realtimeObjects.adapter.clientOptions`. Currently, long method call is polluting the constructor

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livemap/DefaultLiveMap.kt
```diff
@@ -29,7 +30,7 @@ internal class DefaultLiveMap private constructor(
   objectId: String,
   private val realtimeObjects: DefaultRealtimeObjects,
   internal val semantics: ObjectsMapSemantics = ObjectsMapSemantics.LWW
-) : LiveMap, BaseRealtimeObject(objectId, ObjectType.Map) {
+) : LiveMap, BaseRealtimeObject(objectId, ObjectType.Map, SystemClock.clockFrom(realtimeObjects.adapter.clientOptions)) {
```
sacOO7: Can't we add extension method to `SystemClock`, that builds clock from given `realtimeObjects.adapter.clientOptions`. Currently, long method call is polluting the constructor

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
sacOO7: https://github.com/ably/ably-java/pull/1209#discussion_r3309925483

lib/src/main/java/io/ably/lib/rest/Auth.java
ttypic: It's not a public API, our users never should've used this and I just can't imagine situation where they decided to use it, we can have this weird api available, but I would prefer removing it

lib/src/main/java/io/ably/lib/rest/Auth.java
ttypic: that's valid, I'll add

uts/src/test/kotlin/io/ably/lib/uts/infra/FakeClock.kt
ttypic: shouldn't be a problem, we never should create timers concurrently, if we do it's a bug, we need to fix it anyway

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: yes, that's valid

sacOO7: I mean, do we intend to correct this? Like spec https://github.com/ably/specification/blob/main/uts/docs/writing-derived-tests.md#3-deviation-test-patterns says to change it

lib/src/main/java/io/ably/lib/rest/Auth.java
ttypic: done

uts/src/test/kotlin/io/ably/lib/test/mock/FakeClock.kt
ttypic: done

uts/src/test/kotlin/io/ably/lib/uts/infra/FakeClock.kt
ttypic: we assuming schedule(_, 0) is not possible, it's just wrong to use timer thread for that.

Cancelled tasks still fire - is not true, it removes itself from the list once canceled

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: fixed

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livemap/DefaultLiveMap.kt
ttypic: done

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livecounter/DefaultLiveCounter.kt
ttypic: done

lib/src/main/java/io/ably/lib/util/NamedTimer.java
ttypic: done

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: updated

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: source of truth is spec anyway, if llm should check spec compliance

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: done

.claude/skills/uts-to-kotlin/SKILL.md
ttypic: fixed

lib/src/main/java/io/ably/lib/rest/Auth.java
sacOO7: Okay, I can see
```
    Auth(AblyBase ably, ClientOptions options) throws AblyException {
        this.ably = ably;
        this.clock = SystemClock.clockFrom(options);
        this.nanoTimeDelta = clock.currentTimeMillis() - clock.nanoTime()/(1000*1000);
```
We are okay to divide by `(1000*1000)` right?

.claude/skills/uts-to-kotlin/SKILL.md
```diff
@@ -37,19 +39,7 @@ Package: derived from the output path under `kotlin/`.
...
-uts/src/test/kotlin/io/ably/lib/test/mock/MockWebSocket.kt
-uts/src/test/kotlin/io/ably/lib/test/mock/MockHttpClient.kt
-uts/src/test/kotlin/io/ably/lib/test/mock/PendingRequest.kt
-uts/src/test/kotlin/io/ably/lib/test/mock/PendingConnection.kt
-uts/src/test/kotlin/io/ably/lib/test/mock/FakeClock.kt
-uts/src/test/kotlin/io/ably/lib/test/mock/MockEvent.kt
-```
-
----
+Read ALL of files in `uts/src/test/kotlin/io/ably/lib.uts/infra` before generating any code (you need exact method signatures).
```
sacOO7: `uts/src/test/kotlin/io/ably/lib/uts/infra` instead of `uts/src/test/kotlin/io/ably/lib.uts/infra` right?

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: This doesn't seem like fixed yet?

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: yeah, maybe some reason we kept it

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Can't see the change : (

uts/src/test/kotlin/io/ably/lib/uts/infra/FakeClock.kt
sacOO7: Okay got it 👍

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Btw, can you make sure skill always fetch the URL with a note
```
Note - Always fetch [Writing Derived Tests](https://raw.githubusercontent.com/ably/specification/refs/heads/main/uts/docs/writing-derived-tests.md) as a part of SKILL load.
```

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livemap/DefaultLiveMap.kt
sacOO7: thanks

liveobjects/src/main/kotlin/io/ably/lib/objects/type/livecounter/DefaultLiveCounter.kt
sacOO7: thanks

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Can you just double check about this? Or atleast throw error/log message if `ARGUMENTS` not provided

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Should be `uts/src/test/kotlin/io/ably/lib/uts/infra` instead of `uts/src/test/kotlin/io/ably/lib.uts/infra`

uts/src/test/kotlin/io/ably/lib/realtime/unit/connection/ConnectionRecoveryTest.kt
sacOO7: Do we need to add extra assertions here?

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: Markdown fetch is not made mandatory, otherwise without enough context LLM will hallucinate

sacOO7:
Can you check comment [Markdown fetch is not made mandatory, otherwise without enough context LLM will hallucinate](https://github.com/ably/ably-java/pull/1209#discussion_r3323269845) and other test assertion related comments -> https://github.com/ably/ably-java/pull/1209#discussion_r3309264369, https://github.com/ably/ably-java/pull/1209#discussion_r3309264381 etc
Otherwise changed LGTM

.claude/skills/uts-to-kotlin/SKILL.md
sacOO7: We have faced similar issues previously, where reference URLs are not fetched unless made explicitly mandatory

sacOO7 (approved):
LGTM

PR #1228, opened 2026-07-21
[AIT-1142] fix(liveobjects): objects audit conformance; refactor(uts): shared test-infra module

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1142

This PR brings ably-java's LiveObjects implementation into line with the reconciled objects spec and, on the same branch, reorganises how the shared UTS (unit test spec) suites are built and owned. It started as a focused set of inbound-operation fixes from the cross-SDK objects audit, then absorbed two follow-on pieces of work that were each reviewed as their own PRs and merged in here: further objects spec-conformance points (originally #1229) and a restructuring of the `:uts` test-infra module (originally #1231). The net result tells two largely independent stories — a **behavioural** one in the LiveObjects production code, and a **structural** one in the test tooling — and the sections below take them in turn.

## Problem statement

**LiveObjects op-handling and conformance.** The ably-js LiveObjects spec-compliance audit (ably/ably-js#2263) surfaced several bug classes. We checked the same classes against `liveobjects/src/main/kotlin`, using the reconciled objects spec (`objects-features.md` plus `uts/objects`) as the reference. Two of the four ably-js bugs were present here — one of them in a worse, transport-dependent form — and a cluster of newer op-path and lifecycle spec points from ably/specification#512 were not yet honoured at all. Concretely: a single malformed inbound operation could discard its siblings, because the batch loop in `ObjectsManager.applyObjectMessages` has no per-operation catch and any throw aborts the rest of the same `ProtocolMessage` (silent data loss within the batch); a missing counter `number`/`count` was unrepresentable on the wire and failed differently per transport (msgpack threw at decode, while JSON silently defaulted to `0.0` and emitted a spurious update); and the op-path return value, channel-state data lifecycle, and root-object safeguards did not yet match RTO27, RTLC9g/RTLM7f, RTO10c1b1/RTLO4e10 and RTO18d.

**UTS test-infra ownership.** Separately, the shared UTS infrastructure (mock WebSocket/HTTP transports, `FakeClock`, client factories, `SandboxApp`, proxy control) lived in `:uts`'s `java-test-fixtures` variant, and every spec-derived UTS suite lived inside `:uts` regardless of which module's code it actually exercised. That arrangement had three growing costs. Consuming the infra was awkward: each module needed the `testFixtures(project(":uts"))` plumbing and, worse, had to re-declare the whole test-framework stack itself. Test ownership was wrong: realtime suites tested `:java` but lived in `:uts`, and objects integration/proxy suites tested the LiveObjects plugin from outside `:liveobjects`, which forced a `testRuntimeOnly` back-edge to get the plugin onto the runtime classpath. And there was no path to publishing, because a testFixtures variant of a test-host module is not a publishable artifact — a future cross-repo consumer such as the Chat SDK would have had no clean way in.

## Summary of changes

The production fixes and the tooling refactor are described in their own groups below. A reviewer who only wants the behavioural changes can read group 1; groups 2–5 are the structural work.

### 1. LiveObjects production conformance and op-handling

The batch loop no longer lets one malformed inbound operation abort its siblings. Two wire-triggerable throws are converted to warn-and-skip so the rest of the batch still applies. `canApplyOperation` previously threw for an empty `serial`/`siteCode`; per RTLO4a3 it now logs a warning and returns `false`, and the caller's RTLC7b/RTLM15b "op serial ≤ site serial" skip log is guarded so it no longer fires with null values. Nil operation payloads (an absent `counterInc`/`mapSet`/`mapRemove`) previously threw `objectError`; they now warn and skip only the offending operation, matching the unsupported-action gates (RTLC7d3/RTLM15d4) that were already correct.

A missing counter `number`/`count` is now representable and behaves the same across transports (RTLC9h/RTLC16d). `WireCounterInc.number` and `WireCounterCreate.count` are nullable, and the msgpack codec round-trips absence — nothing is packed when the value is null, and decode no longer throws when the field is missing. `applyCounterInc` with a missing `number` returns the no-op update and emits no event. `mergeInitialDataFromCreateOperation` with a missing `count` likewise returns the no-op update, but sets `createOperationIsMerged` **before** the no-op return: RTLC16b is unconditional, so RTLC8b's duplicate-create dedup still engages. On the public API, `CounterInc.getNumber()` and `CounterCreate.getCount()` are now `@Nullable`, with the no-op semantics documented in the Javadoc.

The remaining conformance points mirror ably/specification#512. The op-path now returns an `ObjectUpdate` rather than a `Boolean` (RTLC9g/RTLM7f), and the RTO9a2a4 on-ack serial gate uses `!update.noOp`, matching the UTS model where `result == true` is equivalent to `!update.noop`. Objects data now follows the channel-state lifecycle (RTO27): the DETACHED and FAILED transitions clear pooled data without emitting, while SUSPENDED retains it. And the root object is protected two ways — it is excluded from GC (RTO10c1b1) and rejects tombstone attempts (RTLO4e10).

For completeness, several ably-js bug classes were confirmed **not** present here, matching post-audit ably-js: nonce generation is already RTLCV4d-compliant (16 characters, ~95 bits of entropy), and the unsupported-action gates already warn-and-skip. A handful of throws are deliberately kept, again in parity with ably-js: `validateObjectId` mismatch, MAP_SET invalid-value (92000), MAP_CREATE semantics mismatch, and create-op validation during sync.

### 2. `:uts` becomes a shared test-infra module

The 16 infra files now live in `:uts`'s ordinary main source set (`uts/src/main/kotlin/io/ably/lib/uts/infra/…`), replacing the old testFixtures-based approach, with their packages unchanged so consumers see zero import churn. The module now `api`-exports the full UTS test-writing toolkit (JUnit 5 BOM/aggregator/params, the `kotlin-test` Jupiter binding, and coroutines core+test), so a consuming module needs exactly one line:

```kotlin
testImplementation(project(":uts"))
```

`:uts` is a `java-library` + `kotlin.jvm` module that depends on `api(:java)` and `api(:network-client-core)`, keeps ktor as an `implementation` dependency (so it never leaks), and declares Java-8 outgoing variants so `:java` (which targets 1.8) can consume it. The build file documents one invariant: `:uts`'s main configurations never depend on `:liveobjects`, which keeps `:liveobjects test → :uts main → :java` acyclic and lets the old `testRuntimeOnly(:liveobjects)` back-edge disappear entirely.

### 3. UTS suites move to the modules that own the code they test

The objects suites move into `:liveobjects` under `liveobjects/src/test/.../uts/{unit,integration,proxy}` (namespace `io.ably.lib.liveobjects.uts.*`), and the unit tier is expanded with new `InternalLiveCounter`, `InternalLiveMap`, `ObjectId`, `ObjectsPool` and `ParentReferences` suites, alongside white-box tests (`LiveObjectTombstoneTest`, `DefaultRealtimeObjectChannelStateTest`) that pin the new RTO10c1b1/RTLO4e10/RTO27 safeguards. The realtime suites move into `:java` at `lib/src/test/kotlin/io/ably/lib/uts/…` with their packages preserved, run by two new `:java:runUtsUnitTests` / `:java:runUtsIntegrationTests` tasks. The 64 legacy JUnit4 tests and their existing suite tasks are untouched.

`:uts` itself keeps three permanent smoke tests, one per tier (`UnitInfraSmokeTest`, `IntegrationInfraSmokeTest`, `ProxyInfraSmokeTest`), modelled on ably-cocoa#2223. They act as the infra's acceptance gate and double as the worked examples the rewritten `uts/README.md` teaches from; they are deliberately not spec-derived and carry no `@UTS` markers. Deviations now live next to the tests that record them, in `lib/src/test/.../uts/deviations.md` for realtime/rest and `liveobjects/.../uts/deviations.md` for objects.

One deviation is retired rather than moved. The shared-gap entry for RTLO4b4c1 ("to be fixed in both SDKs together") is gone: the ably-js half landed in ably/ably-js#2263 and this PR is the other half, so `LiveObjectSubscribeTest.RTLO4b4c1` now runs the spec-verbatim `counterInc: {}` no-op stimulus ungated.

### 4. Mock-infra contract fixes

These review-driven fixes were checked against the mock contracts documented in the UTS docs. `FakeClock.advance` now runs due work to quiescence — cascades and timers created mid-advance fire within the same advance, which is the Guarantee from ably/specification#518 — and its timer state is hardened against SDK-thread races. Its `waitOn` seam performs a real timed wait (documented as advisory in `uts/README.md` §6.4); that behaviour was the root cause of a CI flake, now fixed by having the smoke test own its reconnect attempts deterministically. Separately, transport `cancel()` now delivers `listener.onClose` per the SDK's own `WebSocketClient` contract; `respondWith` honours the `headers` param and derives the body content-type case-insensitively; `SandboxApp.create()` checks the HTTP status before parsing; teardown rethrows `CancellationException`; and there is assorted `@Volatile` / listener-cleanup hygiene. Finally, `RTO24b1` now awaits its seed's observable effect before subscribing, closing an async-delivery race on slow runners in the pattern the file already established.

### 5. CI, skill and docs

CI is re-pointed in this same change so nothing goes silently green: `check.yml` runs `:java:runUtsUnitTests` and `:uts:runUtsUnitTests` alongside the existing tasks, `integration-test.yml`'s `check-uts` job runs the corresponding integration tasks, and `check-liveobjects` picks up the moved objects tiers through its extended filter. The `uts-to-kotlin` skill's mapping is simplified to one repo-root-relative path per tier, and its resolver now derives and reports the owning Gradle module. `uts/README.md` is rewritten around the new layout with walkthroughs that teach from the smoke tests, and its "Future work" note records the one still-open decision — whether to publish `:uts` for an out-of-repo consumer (the Chat SDK).

## Related spec & cross-SDK context

- **ably/specification#512** — the objects conformance points mirrored here (RTO27, RTLC9g/RTLM7f, RTO10c1b1/RTLO4e10, and RTO18d, whose optional note lets de-duplicating SDKs assert `call_count == 1`).
- **ably/specification#506** — companion spec-side corrections for the op-handling audit.
- **ably/specification#518** — the normative fake-time semantics (ADVANCE_TIME guarantee, hard-gate vs. advisory latitude, test-authoring rule) that `FakeClock` implements.
- **ably/ably-js#2263** — the originating audit and the ably-js fixes.
- **ably/ably-cocoa#2259** — companion mock-contract fixes on the cocoa side; **ably-cocoa#2223** — the smoke-test model.
- **#1231** (uts refactor) and **#1229** (spec-conformance + unit expansion) — the constituent, separately-reviewed PRs now merged into this one.

## Verification

Unit and integration suites are green across every tier:

| Task | Result |
|---|---|
| `:java:runUnitTests` (legacy) | unchanged |
| `:java:runUtsUnitTests` | 6 / 0 |
| `:uts:runUtsUnitTests` | 3 / 0 |
| `:liveobjects:runLiveObjectsUnitTests` | 389 / 0 |
| `:java:runUtsIntegrationTests` | 5 / 0 |
| `:uts:runUtsIntegrationTests` | 4 / 0 |
| `:liveobjects:runLiveObjectsIntegrationTests` | 29 / 0 (real sandbox + uts-proxy) |

The `@UTS` test-ID sets are identical before and after every move, so there is zero coverage loss. For publication isolation, the `:java` POM and jar contain no `org.jetbrains.kotlin` entries and the jar's file list is byte-identical to pre-change. `checkWithCodenarc checkstyleMain checkstyleTest` is green.

## Review guide

Read the diff as behaviour plus structure, not as move noise. The behavioural changes are the production op-handling and conformance work (group 1) and the mock-infra contract fixes (group 4); everything else is structural. Most of the moved files are pure renames (R098–R100): the 16 infra files are content-identical, and the moved objects tests changed only their `package` lines. The one exception is `AuthReauthTest`, which needed a single-token change (`it.message.get` → `it.message?.get`), explained inline — tests living outside `:uts` lose the Kotlin friend-module smart-casts on the infra's public nullable properties.

A few implementation details worth knowing while reviewing the build files, none of which affect the shipped artifact:

- **testFixtures archaeology.** An intermediate `java-test-fixtures` stage existed during development and was superseded on this same branch, so the net diff contains no `testFixtures` at all. Promotion to the main source set was chosen over a separate `:test-support` module — zero import churn, no settings change, and it leaves `:uts` publishable later without restructuring.
- **kotlin-stdlib guardrail.** `:java` gains the Kotlin plugin for tests only. A guardrail strips the plugin's auto-added `kotlin-stdlib` from every main-artifact scope, so the published `:java` artifact stays Kotlin-free (verified via the POM/jar checks above); the stdlib leak comes from the plugin, not from the `:uts` test dependency. `runUnitTests` additionally excludes `io.ably.lib.uts.*`, and the two frameworks cannot discover each other's classes.
- **JUnit platform in `:liveobjects`.** The incoming Jupiter suites require the JUnit Platform, so `:liveobjects` adopts it with `kotlin("test-junit5")` pinned (auto-selection is non-deterministic in mixed-runner modules) while the vintage engine keeps running the module's own legacy JUnit4 tests. `runLiveObjectsUnitTests` filters both `…unit.*` and `…uts.unit.*`; `runLiveObjectsIntegrationTests` additionally covers `…uts.{integration,proxy}.*`.
- **Build-file diffs are minimal.** `:liveobjects` differs from the base by `-kotlin("test")`, `+project(":uts")`, `+vintage-engine`; `:java` adds one dependency line plus test-only mechanics. `gradle/libs.versions.toml` gains only the five JUnit entries (catalog-first is the repo convention, and the one pre-existing raw-string dependency is removed here).

The public-interface nullability change on `CounterInc`/`CounterCreate` was verified against all consumers, with `:liveobjects:compileKotlin` and `:java:compileJava` both clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Counter create and increment messages now support missing numeric values without failing or applying unintended changes.
  - Invalid or incomplete live-object operations are safely treated as no-ops.
  - Root objects are protected from accidental tombstoning or garbage collection.
  - Improved handling of malformed synchronization data and clearer synchronization errors.
  - Unicode map keys now use UTF-8 byte length for accurate message-size calculations.

- **Documentation**
  - Updated testing and integration guidance for live-object and realtime functionality.
  - Expanded validation coverage for live-object operations and synchronization scenarios.


uts/README.md
```diff
@@ -163,7 +163,7 @@ what's missing". The reference tests this guide walks through correspond to thes
   `realtime/integration/proxy/auth_reauth.md` → **`AuthReauthTest.kt`**.
 
 > There is also a fifth, *referenced* spec:
-> [`realtime/integration/helpers/proxy.md`](https://github.com/ably/specification/blob/main/uts/realtime/integration/helpers/proxy.md)
+> [`docs/proxy.md`](https://github.com/ably/specification/blob/main/uts/docs/proxy.md)
 > (in the spec repo under `uts/realtime/integration/helpers/`). It defines the proxy's control API, rule format,
```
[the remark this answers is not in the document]
sacOO7: Good catch — fixed in f047486d: the parenthetical now reads "in the spec repo under `uts/docs/`", matching the `uts/docs/proxy.md` link above it. (Copilot flagged the same line — covered by the same fix.)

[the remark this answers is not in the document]
sacOO7: Fixed in f047486d — the parenthetical `uts/realtime/integration/helpers/` is corrected to `uts/docs/`, matching the link target. (Same issue as CodeRabbit's comment on this line.)

uts/src/test/kotlin/io/ably/lib/uts/deviations.md
```diff
@@ -80,47 +80,6 @@ Entries are grouped by actionability:
...
-**Tests affected (LiveObjectSubscribeTest.kt):**
-- `RTLO4b4c1 - noop update does not trigger listener` (RTLO4b4c1/noop-no-trigger-0) — env-gated.
-
-> Note: the internal `LiveObjectUpdate.noop` diff flag is also not exposed on the public
-> `InstanceSubscriptionEvent` (only `getObject()`/`getMessage()`, mapping §8); the spec frames noop
-> suppression as *the listener not firing*, which is the observable the env-gated test asserts.
-
----
-
 # 3) Expected — typed-SDK / language adaptations (objects) — NOT bugs, no action
```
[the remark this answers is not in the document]
sacOO7: Outdated — this file was split by the uts-infra refactor into `lib/src/test/kotlin/io/ably/lib/uts/deviations.md` (realtime/rest) and `liveobjects/src/test/kotlin/io/ably/lib/liveobjects/uts/deviations.md` (objects). The group table is now an explicit shared taxonomy prefaced with "only groups with entries in this file appear as sections below", so an absent group section is intentional rather than a numbering skip.

java/build.gradle.kts
```diff
@@ -28,6 +30,31 @@ dependencies {
         runtimeOnly(project(":network-client-default"))
     }
     testImplementation(libs.bundles.tests)
+
+    // The UTS test toolkit — the whole JUnit 5 + kotlin.test-junit5 + coroutines stack arrives
+    // transitively via :uts's exported (`api`) toolkit, so consumers declare only this one edge.
+    // The UTS Kotlin suites run via the runUts* tasks only (JUnit4 tasks don't discover Jupiter
+    // classes and vice versa). Deliberately NO junit-vintage-engine here: unlike :liveobjects, the
+    // legacy JUnit4 tests stay on the JUnit4 runner, never the platform.
+    testImplementation(project(":uts"))
```
sacOO7, line 35: We have moved `uts-infra` as shared module for `realtime`, `rest` and `liveobjects` packages.
So, tests now resides in their own packages with access to internal members. So, UTS unit tests don't need to use reflection and can safely access internal methods/properties etc : )

So, you can check this config. I validated locally, so config. works as expected.
You can review this once more @ttypic

sacOO7: Also, check `liveobjects/build.gradle.kts`

.claude/skills/uts-to-kotlin/references/objects-mapping.md
```diff
@@ -674,4 +699,199 @@ wrapped in `LiveMapValue.of`; `at(...)` followed by `asLiveCounter()` before cou
...
+
+### 17.1 Instantiation — no public constructors, use the factories
+
+The internal classes have `private constructor`s; each pairs with an `internal` companion factory that
+needs a `DefaultRealtimeObject`. Build one from the mocked adapter (helper already exists in
+`liveobjects/src/test/.../unit/TestHelpers.kt`):
+
+```kotlin
+val ro = DefaultRealtimeObject("test", getMockAblyClientAdapter())
+// teardown: unmockkAll() and ro.objectsPool.dispose()  (the pool init starts a real GC coroutine)
```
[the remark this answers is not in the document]
sacOO7: Partly disagree, partly taken: `getMockAblyClientAdapter()` is defined in `liveobjects/src/test/kotlin/io/ably/lib/liveobjects/unit/TestHelpers.kt` — not in `uts/unit/Helpers.kt`, which doesn't define the symbol, so the suggested path would misdirect translators. The abbreviated `.../unit/` ellipsis was genuinely ambiguous though (both `unit/` and `uts/unit/` helper files exist), so the doc now spells out the full path. Fixed in e1ed6cd5.

lib/src/test/kotlin/io/ably/lib/uts/deviations.md
```diff
@@ -0,0 +1,99 @@
...
+---
+
+## RTN16g2 — Fatal ERROR must be sent without closing the transport
+
+**Spec point:** RTN16g2  
+**What the spec requires:** Trigger FAILED state by sending a fatal ERROR message followed by closing the WebSocket (`send_to_client_and_close`), using error code 50000/statusCode 500.  
+**What the SDK does (two issues):**  
+1. Error code 50000/statusCode 500 is not treated as fatal by `isFatalError()` (requires code 40000–49999 or statusCode < 500), so FAILED is never reached with the spec's values.  
+2. Sending `close(1000)` after the ERROR dispatches a synchronous `DISCONNECTED` action that races with and preempts the async `FAILED` transition triggered by the ERROR message.  
+**Workaround in tests:** Use `sendToClient` (no close frame) with code 40000/statusCode 400. The SDK's own FAILED-state handler calls `clearTransport()`, so the explicit close is not needed.
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — the heading now describes the actual deviation (the spec's fatal-ERROR-plus-close trigger can't drive this SDK to FAILED) instead of restating the workaround as a requirement. While here, the whole file was restructured into the canonical four-section deviations format from the UTS docs (`writing-derived-tests.md`): UTS Spec Errors / Failing Tests / Adapted Tests / Mock Infrastructure Limitations, preserving all recorded entries.

liveobjects/src/main/kotlin/io/ably/lib/liveobjects/ObjectsState.kt
```diff
@@ -90,66 +106,72 @@ internal abstract class ObjectsStateCoordinator : ObjectStateChange, HandlesObje
...
-        deferred.await()
-      } finally {
-        // off() the one-shot on either path (same cleanup pattern as awaitSyncCompletion) so it never
-        // lingers if the waiting coroutine is cancelled (e.g. dispose while a get() awaits sync).
-        internalObjectStateEmitter.off(ObjectStateEvent.SYNCED, syncedListener)
-      }
+      // RTO23c1 - route get()'s wait through the same [pendingSyncWaiters] machinery publishAndApply uses (RTO20e/RTO20e1) so [failSyncWaiters] fails it with the 92008 error instead of orphaning it on DETACHED/SUSPENDED/FAILED
+      awaitSyncCompletion("the object could not be retrieved")
     }
   }
```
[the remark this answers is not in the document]
sacOO7: The two error paths cannot co-occur for a single `get()`, so there's no ambiguity or race here. `getRootAsync()` awaits `adapter.ensureAttached()` (RTO23e → RTL33; the only 90001 producer reachable from `get()`) to completion **before** `objectsManager.ensureSynced()` runs — both on the single-lane `sequentialScope`. Either the precondition throws 90001 (channel already FAILED at call time, RTL33c) and no waiter is ever registered, or it returns and only the parked-waiter path remains, which fails with 92008 iff the channel leaves a usable state **during** the wait (RTO23c1, produced solely by `failSyncWaiters`). That per-temporal-case split is exactly what RTO23e + RTO23c1 prescribe. The CI failure this referenced was a test-side ordering race (the injected state change could beat the waiter registration), fixed in f9492ab5 and pinned spec-side via ably/specification#519.

liveobjects/src/main/kotlin/io/ably/lib/liveobjects/ObjectsSyncTracker.kt
```diff
@@ -31,32 +47,47 @@ internal class ObjectsSyncTracker(syncChannelSerial: String?) {
...
+        ParsedSyncChannelSerial(
+          syncId = match.groupValues[1],
+          syncCursor = match.groupValues[2],
+          isMalformed = false,
+        )
       } else {
-        Pair(null, null)
+        // RTO5a6 - present but lacks the `:` separator to split into <sequence id>:<cursor value>; flag it so the caller warns and treats it as absent
+        ParsedSyncChannelSerial(syncId = null, syncCursor = null, isMalformed = true)
       }
```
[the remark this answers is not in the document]
sacOO7: Valid finding — the `[\w-]+` charset restriction has no spec basis: RTO5a6 defines malformed solely as "does not contain the `:` separator", and nothing in the spec constrains the sequence-id character set. It's a latent cross-SDK issue rather than a java one: ably-js carries the identical regex (this code was ported from it), ably-cocoa already splits on the first colon charset-agnostically, and the UTS corpus has no test that would catch it. Rather than fix it piecemeal here, it's tracked comprehensively — per-repo fixes, spec clarification, and the UTS coverage gap — in ably/specification#520. Intentionally not changed in this PR.

liveobjects/src/test/kotlin/io/ably/lib/liveobjects/integration/setup/Sandbox.kt
```diff
@@ -1,62 +1,30 @@
...
-      val body = JsonParser.parseString(response.bodyAsText())
-
-      return Sandbox(
-        appId = body.asJsonObject["appId"].asString,
-        // From JS chat repo at 7985ab7 — "The key we need to use is the one at index 5, which gives enough permissions to interact with Chat and Channels"
-        apiKey = body.asJsonObject["keys"].asJsonArray[0].asJsonObject["keyStr"].asString,
-      )
+      val app = SandboxApp.create()
+      // defaultKey is the full-capability "appId.keyId:keySecret" key (index 0 of the app spec)
+      return Sandbox(appId = app.appId, apiKey = app.defaultKey)
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — `Sandbox` now retains the provisioned `SandboxApp` and `tearDownAfterClass()` deletes it (best-effort per the UTS docs — `SandboxApp.delete()` already swallows failures and rethrows only `CancellationException`), matching the pattern the :java integration suites use.

liveobjects/src/test/kotlin/io/ably/lib/liveobjects/unit/LiveObjectTombstoneTest.kt
```diff
@@ -0,0 +1,89 @@
...
+  private fun rootMapWithNameEntry(): InternalLiveMap {
+    val realtimeObject = DefaultRealtimeObject("test", getMockAblyClientAdapter())
+    val map = InternalLiveMap.zeroValue("root", realtimeObject)
+    map.data["name"] = LiveMapEntry(timeserial = "01", data = WireObjectData(string = "Alice"))
+    map.siteTimeserials["site1"] = "00"
+    return map
+  }
+
+  @After
+  fun tearDown() = unmockkAll() // getMockAblyClientAdapter uses mockkStatic - clean up global state
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — the test now retains the `DefaultRealtimeObject` and disposes its `objectsPool` before `unmockkAll()`, per the documented teardown contract. A sweep for the same pattern found one more offender (`DefaultRealtimeObjectAsyncTest` — instances weren't even retained), fixed the same way, and `ValueTypesTest`'s inverted teardown ordering was aligned to dispose-before-unmockkAll. All other internal-unit suites already complied.

liveobjects/src/test/kotlin/io/ably/lib/liveobjects/uts/unit/InternalLiveMapApiTest.kt
```diff
@@ -0,0 +1,325 @@
...
+ *
+ * Public-tier spec: uses only the public API plus the module-local `Helpers.kt`. The spec's
+ * hand-rolled capturing mock maps to the standard `setupSyncedChannel` transport plus the
+ * mock's recorded `MessageFromClient` event log as the spec's `captured_messages` — see
+ * [capturedObjectMessages].
+ *
+ * Note: evaluating a `LiveCounter`/`LiveMap` value type generates its objectId from server
+ * time (RTO16), which the SDK fetches once per JVM via REST `GET /time` — the mock transport
+ * does not intercept HTTP, so the first `*_CREATE` test in a run performs that single
+ * unauthenticated request against the real endpoint.
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — the note was stale: `setupSyncedChannel` installs a `MockHttpClient` that answers `GET /time` locally (Helpers.kt), so no real network request is ever made. The note now documents the hermetic behavior and the unit-tier no-network contract.

uts/README.md
```diff
@@ -156,122 +159,182 @@ segregation exists because proxy tests have different infra needs, CI cadence, a
...
-    useJUnitPlatform()                                   // JUnit 5
+    useJUnitPlatform()                       // JUnit 5
     jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
     jvmArgs("--add-opens", "java.base/java.lang=ALL-UNNAMED")
-    // Propagate a local proxy build override (see ProxyManager):
-    systemProperty("uts.proxy.localPath", /* -Duts.proxy.localPath=… or $UTS_PROXY_LOCAL_PATH */ …)
+    // Propagate a local proxy-build override (see ProxyManager): -Duts.proxy.localPath=… or
+    // $UTS_PROXY_LOCAL_PATH.
+    systemProperty("uts.proxy.localPath", …)
 }
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — replaced the ellipsis with the real `systemProperty(...)` expression from `uts/build.gradle.kts`, so the snippet is copy-pasteable and documents the `-D`-or-env fallback.

uts/src/main/kotlin/io/ably/lib/uts/infra/Utils.kt
```diff
@@ -28,11 +28,18 @@ suspend fun awaitState(
...
+          if (change.current == target && cont.isActive) {
+            client.connection.off(listener)
+            cont.resume(Unit)
+          }
         }
         client.connection.on(listener)
-        if (client.connection.state == target && cont.isActive) cont.resume(Unit)
+        if (client.connection.state == target && cont.isActive) {
+          client.connection.off(listener)
+          cont.resume(Unit)
```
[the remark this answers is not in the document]
sacOO7: Fixed in e1ed6cd5 — both `awaitState` and `awaitChannelState` now use the atomic single-winner `tryResume(Unit)?.let(cont::completeResume)` in the listener and the immediate check, so only one racer resumes (the pair is `@InternalCoroutinesApi`, opted in with a comment). The `invokeOnCancellation` cleanup was already in place. Verified across all UTS unit + integration suites.

java/build.gradle.kts
```diff
@@ -1,11 +1,13 @@
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
     alias(libs.plugins.build.config)
     alias(libs.plugins.maven.publish)
     alias(libs.plugins.test.retry)
     checkstyle
     `java-library`
+    alias(libs.plugins.kotlin.jvm)        // NEW — test-only usage; see stdlib guardrail (step 4)
```
ttypic: I think we can remove comment, it's not informative

java/build.gradle.kts
```diff
@@ -28,6 +30,31 @@ dependencies {
...
+// main dependency scope, which would leak into :java's published POM/runtime. The leak comes from the
+// PLUGIN, NOT from the testImplementation(project(":uts")) dependency — test scopes never enter the
+// POM (verified: removing the :uts dep leaves the leak identical). :java is Kotlin-free at
+// runtime, so strip it from the main artifact scopes. kotlin-stdlib still reaches the TEST classpath
+// transitively (via :uts's kotlin-test-junit5), so the UTS Kotlin suites compile and run.
+// Verified empirically on Kotlin 2.1.10: the plugin adds stdlib lazily (it does not appear in any
+// declared `(n)` view but resolves top-level onto compile/runtimeClasspath), so the removeIf must
+// cover the base scopes the outgoing variants (apiElements/runtimeElements) and classpaths inherit
+// from. Test scopes are untouched.
+listOf("api", "implementation", "runtimeOnly").forEach { cfg ->
```
ttypic: Don't think we need this

java/build.gradle.kts
ttypic: good, I would drop "Deliberately NO junit-vintage-engine here: unlike :liveobjects, the
    // legacy JUnit4 tests stay on the JUnit4 runner, never the platform." - doesn't add any useful information

java/build.gradle.kts
```diff
@@ -47,7 +74,15 @@ sourceSets {
         java {
             srcDirs("src/test/java", "../lib/src/test/java")
         }
+        kotlin {
+            srcDirs("src/test/kotlin", "../lib/src/test/kotlin")   // NEW — UTS Kotlin suites only
```
ttypic: not useful comment

java/build.gradle.kts
```diff
@@ -47,7 +74,15 @@ sourceSets {
...
+        }
     }
+    // main gets NO kotlin srcDir — :java main stays pure Java.
```
ttypic: I would delete this as well

uts/build.gradle.kts
```diff
@@ -1,23 +1,48 @@
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 
 plugins {
-    alias(libs.plugins.kotlin.jvm)
+    `java-library`                          // NEW — provides the `api` configuration. Previously
```
ttypic: let's remove comment

uts/build.gradle.kts
```diff
@@ -1,23 +1,48 @@
...
+                                            // arrived transitively via `java-test-fixtures`;
+                                            // kotlin.jvm alone applies only the plain `java` plugin.
+    alias(libs.plugins.kotlin.jvm)          // `java-test-fixtures` REMOVED
```
ttypic: and here

uts/build.gradle.kts
```diff
@@ -1,23 +1,48 @@
...
+}
+
+java {
+    // Declare Java-8 outgoing variants (org.gradle.jvm.version=8) so :java's 8-requesting resolvable
```
ttypic: also not useful comments talks about implementation phases

ttypic (commented):
the new approach for uts location looks good, but I would clean up gradle files a bit: remove comments that only useful for claude and gradle tasks that delete dependencies that should be added anyway

java/build.gradle.kts
sacOO7: Good catch — removed in 960d044d. These were refactor-time notes ("NEW", "REMOVED", etc.) that don't help a future reader.

java/build.gradle.kts
sacOO7: Good catch — removed in 960d044d. These were refactor-time notes ("NEW", "REMOVED", etc.) that don't help a future reader.

java/build.gradle.kts
sacOO7: Good catch — removed in 960d044d. These were refactor-time notes ("NEW", "REMOVED", etc.) that don't help a future reader.

uts/build.gradle.kts
sacOO7: Good catch — removed in 960d044d. These were refactor-time notes ("NEW", "REMOVED", etc.) that don't help a future reader.

uts/build.gradle.kts
sacOO7: Good catch — removed in 960d044d. These were refactor-time notes ("NEW", "REMOVED", etc.) that don't help a future reader.

java/build.gradle.kts
sacOO7: Dropped in 960d044d — the whole 5-line block is now a single line noting the toolkit arrives transitively via :uts's `api`. The vintage-engine sentence and the rest were refactor-time narration.

uts/build.gradle.kts
sacOO7: Dropped the internal-phase note in 960d044d. I kept a single line explaining why this module targets Java 8 (so the Java-8 `:java` module can consume it on its classpath), since that's a real constraint rather than history.

java/build.gradle.kts
sacOO7: Agreed the imperative strip block is ugly — removed in 960d044d. A heads-up on why it existed, though: applying the Kotlin plugin to `:java` (for the test-only Kotlin sources) makes the plugin auto-add `kotlin-stdlib` to the module's *main* scopes, and that flows into the published `io.ably:ably-java` POM as a `compile` dependency. You can verify it yourself: with the block deleted and nothing else, run `./gradlew :java:generatePomFileForMavenPublication --rerun-tasks` and then `grep org.jetbrains.kotlin java/build/publications/maven/pom-default.xml` — `kotlin-stdlib:2.1.10` (scope `compile`) shows up, which would turn our Java-only artifact into a Kotlin-requiring one.

So instead of adding-then-removing, I switched to the plugin's first-class flag: `kotlin.stdlib.default.dependency=false` in `java/gradle.properties`, so the stdlib is never added in the first place. Verified: the published POM is Kotlin-free again, `:java:runUtsUnitTests` still compiles and passes (stdlib reaches the test classpath transitively via `:uts`), and `:liveobjects` is unaffected (the property is scoped to `:java`).

ttypic (approved):

kjoker281099-creator (approved):

PR #1233, opened 2026-09-01
[Split 2/3] Add io.ably.pubsub:device and :server door packages with side-declaring agents

umair-ably:
Second PR in the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) split stack (stacked on #1232; diff shows only this PR's changes). Reference: [ably-js#2293](https://github.com/ably/ably-js/pull/2293), whose `packages/shared/side.ts` contract this ports.

## What this PR does

* **`io.ably.pubsub:server`** (jar → depends on `:core`): `PubSubServer.httpClientBuilder(...)` / `PubSubServer.realtimeClientBuilder(...)`, per the door names agreed in [PDR-091b2](https://ably.atlassian.net/wiki/spaces/product/pages/5348425729). Builders accept everything the core constructors accept (`ClientOptions`, API key string, token string — reusing the core's colon-rule disambiguation) and stamp the side at `build()`.
* **`io.ably.pubsub:device`** (aar → depends on `:core-android`): `PubSubDevice.clientBuilder(...)` — one door, per PDR-091 (device-side connectionless ops stay available on the one client).
* **Side-agent contract** in a shared source dir (`shared/`, compiled into both door artifacts, not published — the Java analogue of ably-js's `packages/shared/side.ts`):
  * identifiers `ably-pubsub-device` / `ably-pubsub-server`, with the load-bearing-suffix warning comment (the `-server` suffix is what earns the MAU exemption on API-key auth);
  * caller `agents` entries preserved; the side entry is applied last and cannot be overridden; the caller's `ClientOptions` is never mutated; `null` passes through to the core's own initialization error.
  * The core keeps its `ably-java/<version>` base identifier — unchanged, already in the ably-common registry.
* **Fixes `ClientOptions.copy()`**, which silently dropped `headers`, `fallbackHosts`, `transportParams` and `agents`. The doors rely on `copy()` for non-mutating stamping; the pre-existing internal callers of `copy()` also benefit. Unit-tested.

## Tests (what billing reads — they fail loudly)

* `server`: unit tests for stamping/preservation/override/no-mutation/null, plus a **wire-level test** that spins a local HTTP server and asserts the actual `Ably-Agent` header contains `ably-pubsub-server/2.0.0` and `ably-java/`. Runs in `check.yml` via the existing unqualified `runUnitTests` invocation.
* `device`: instrumentation tests asserting the same contract, added to the `emulate.yml` matrix (`:device:connectedAndroidTest`).

## Open items flagged for review

* `ably-pubsub-device` / `ably-pubsub-server` are **not yet in the ably-common agents registry** (checked `protocol/agents.json` on main, 2026-09-01) — needs an ably-common PR before any release; identifiers here match the ably-js branch exactly.
* Builder surface is deliberately minimal (no fluent option setters yet) pending sign-off on the door surface with the chat-kotlin owner.
* Both door artifacts bundle an identical `io.ably.pubsub.internal.Side` class — same pattern as the existing `lib/` source shared between the two cores; harmless if both artifacts ever meet on a classpath.

## Summary by CodeRabbit

- **New Features**
  - Added dedicated device and server SDK entry points for building authenticated REST and realtime clients.
  - Device and server clients now identify their SDK side in connection and request metadata.
  - Added published device and server library modules with streamlined builder APIs.

- **Bug Fixes**
  - Client option copies now retain headers, fallback hosts, transport settings, and agent information.
  - Caller-provided agent metadata is preserved while SDK-specific side identifiers are applied consistently.


umair-ably:
Updated for [ably/ably-common#361](https://github.com/ably/ably-common/pull/361), mirroring [ably/ably-js#2297](https://github.com/ably/ably-js/pull/2297): the side-declaring agent entries are now **versionless flags** — bare tokens on the wire, like `browser` — with identity/version continuing to travel on the SDK's own `ably-java/<version>` entry:

```
ably-java/2.0.0 jre/17 ably-pubsub-server
```

Unlike ably-js (which needed a `getAgentString` fix for `name/undefined`), our `AgentHeaderCreator` already emits a null-valued map entry as a bare token, so no core change was needed — the doors stamp `null` and drop their BuildConfig version plumbing. Tests now assert the flag appears as a bare token and fail if any `/version` form regresses, and that each side's client doesn't carry the other side's flag.

umair-ably:
Added the family-identifier half of the agent convention (mirrors [ably-js#2297](https://github.com/ably/ably-js/pull/2297)): `Defaults.ABLY_AGENT_VERSION` now stamps `ably-pubsub-java/<version>`, with the four identifier assertions updated (`RealtimeHttpHeaderTest` canary, `PubSubServerTest`, `Side` javadoc). Verified locally: `:server:test` and `:core:test --tests '*RealtimeHttpHeader*'` pass on JDK 17. Registry entries for both the family identifier and the versionless side flags are in [ably-common#361](https://github.com/ably/ably-common/pull/361), which should merge before any release from this branch.

PR #1234, opened 2026-09-01
[Split 3/3] Lockstep release pre-flight, dry-run CI job, README note

umair-ably:
Third PR in the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) split stack (stacked on #1233; diff shows only this PR's changes).

## What this PR does

* **`verifyReleaseArtifacts`** root task: asserts the exact set of published artifacts (`core`, `core-android`, `device`, `server`, `liveobjects`, `pubsub-adapter`, `network-client-*` — all `io.ably.pubsub` at one `VERSION_NAME`), failing loudly on lockstep drift (module-local `VERSION_NAME` overrides) or a changed artifact set. Verified locally: passes on the branch, and both failure modes abort with an explanatory error.
* **`release.yaml`** runs the pre-flight before `publishAndReleaseToMavenCentral`, so a partial/mismatched release fails before anything is uploaded (Central Portal staging gives us the atomic-release backstop on top).
* **`release-dry-run` job in `check.yml`**: pre-flight + `publishToMavenLocal` on every PR, so release-day regressions surface continuously.
* **README**: a note describing the in-development 2.0 artifact set and doors; the full README/docs rewrite is a later phase.

## Where the stack stops (deliberately — same point as ably-js#2293)

Matching the agreed stopping point:
* ❌ No repo rename (`ably-java` → `ably-pubsub-java`) — the IAM role from [infrastructure#13005](https://github.com/ably/infrastructure/pull/13005) is provisioned and waiting; `javadoc.yml`'s role reference, `features.yml`'s `repository-name` and `release.yaml`'s repo guard all change in the rename pass.
* ❌ Nothing published — first publish under `io.ably.pubsub` should be a validation-only portal run, then an RC (per the rollout plan; Maven Central is immutable, so no placeholder versions).
* ❌ No ⚠️PDR-091d API renames (`AblyRest`→`HttpClient` etc.) — still in review; sequenced as the last change on `integration/v2`.
* Remaining before GA (tracked in the rollout plan): ably-common agents-registry PR for `ably-pubsub-device`/`ably-pubsub-server`, namespace decision for the new artifacts, chat-kotlin/ably-flutter coordination, maintenance branch for 1.x, migration guide.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added automated release validation to confirm all required packages use consistent versions and are included before publication.
[...]

PR #1235, opened 2026-09-02
[Split 4/4] Run the UTS through the server door's builders

umair-ably:
Fourth PR in the [PDR-091b](https://ably.atlassian.net/wiki/spaces/product/pages/5362810886) split stack (stacked on #1234; diff shows only this PR's changes). Mirrors [ably/ably-js#2294](https://github.com/ably/ably-js/pull/2294).

## What this PR does

* **`uts.side` mode on the UTS's single construction seam** (`ClientFactories.kt`, used by all 10 unit + integration test files): `core` (default) keeps the core constructors; `server` routes both client kinds through `PubSubServer.httpClientBuilder`/`realtimeClientBuilder`. The builders only stamp the versionless side flag and pass everything else through, so **every UTS behaviour must pass identically through the door** — that's the conformance claim this PR adds.
* **CI runs both legs**: `check.yml` (UTS unit) and `integration-test.yml` (`check-uts`) each gain a `-Duts.side=server` invocation.
* **Harness self-test** (`SideModesTest`, the analogue of ably-js's `side_modes.test.ts`): captures the actual `Ably-Agent` header via the mock HTTP engine and asserts the selected mode's stamp — bare `ably-pubsub-server` token in server mode (fails on any `/version` regression, per [ably-common#361](https://github.com/ably/ably-common/pull/361)), no side token in core mode, `ably-java/` always present. Exists so a broken seam can't silently turn the server leg into a duplicate core run. `@Timeout(30)` so a seam regression fails rather than hangs.
* **Completes `DebugOptions.copy()`** with the four fields base `ClientOptions.copy()` was also missing (fixed in #1233): `headers`, `fallbackHosts`, `transportParams`, `agents`. This matters here because the door stamping relies on the *polymorphic* `copy()` to carry the suite's mock hooks (`httpEngine`, `webSocketEngineFactory`, `clock`, listeners) through construction — the full server-mode suite passing green is the proof.

## Where this differs from ably-js (deliberately)

* **No `device` leg**: `io.ably.pubsub:device` is an Android AAR and cannot run on the JVM the UTS uses (ably-js's device package runs fine in Node). The device door's stamping contract is covered by the instrumentation tests in the `device` module (`emulate.yml`). Recorded in the uts README.
* No constructor-shaped wrapper trickery needed: Kotlin's factory-function seam (`TestRealtimeClient {}`) already is the ably-js `sideConstructor` in spirit, so the change is a `when` on the mode.
* The two auxiliary `AblyRest(...)` constructions (liveobjects REST provisioning, proxy token signer) stay on core constructors — they're test infrastructure, not the SDK under test.

Verified locally: full `:uts:runUtsUnitTests` green in both modes; `SideModesTest` runs and asserts in both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Copying debug configuration now preserves headers, fallback hosts, transport parameters, and agent settings.

* **Testing**
  * Expanded automated coverage for core and server client modes, including authentication, token requests, proxy reauthentication, and agent identification.
  * Added validation for server-side client builders and consistent behavior across REST and realtime clients.

* **Documentation**
  * Documented test configuration options for selecting client modes and authentication behavior.


umair-ably:
The failed `check-uts` leg turned out to be a **real finding, not a test bug** — arguably the strongest justification for this PR existing:

> `40167: a connection or request may only declare itself as a server via a signed x-ably-clientType token claim`

Sandbox realtime **already enforces** that on token auth the agent entry alone may not claim the server side — the side must come from a signed `x-ably-clientType` token claim, which is [PDR-091 deferred decision D2](https://ably.atlassian.net/wiki/spaces/product/pages/5220106242) and isn't implemented anywhere yet. Key-auth server clients connect fine (verified against sandbox with the bare flag); exactly the token-auth tests (`TokenRequestTest` ×2, `AuthReauthTest`) failed. ably-js's UTS never hit this because it's fully mocked — this repo's sandbox-backed UTS leg is what caught it.

Resolution here: token-auth tests call `assumeSideSupportsTokenAuth()` and are **skipped (not failed) on the server leg** with the 40167/D2 rationale in the message; the core leg still runs them. When D2 lands, the test infra can mint the claim and the assumption gets deleted (tracked in the uts README).

**Programme-level implication worth escalating** (cc D2's owner): as enforced today, *any* customer using the server package with token auth gets 40167 — if prod behaves like sandbox, the server door is effectively key-auth-only until D2 ships. That sequencing constraint (D2 before/with the pubsub GA) doesn't appear in the rollout plan.

umair-ably:
**No more skips on the server leg.** The token-auth handling is now per token format:

* **JWTs** (`AuthReauthTest`): the test mints its own JWT (`AblyJwt`, HS256 via JDK crypto) and adds `x-ably-clientType=server` on the server leg — the declaration realtime requires on token auth. Verified against sandbox; also the first end-to-end proof the claim path works through the new server package.
* **Native tokens** (`TokenRequestTest`): the format can't carry the claim, but the failing part was only a client authenticating *itself* with a native token while flagged as a server. That's not how the feature is used — servers mint native tokens *for devices* — so the test now splits its clients across the seam: the **minting** client (the `createTokenRequest` surface under test) goes through the server door on every leg, and the **consuming** client models the device and stays a plain core client. Full RSA9 coverage through the door, no skip, and the structure now mirrors real usage.

`assumeSideSupportsTokenAuth()` is deleted — nothing needs it. If the platform later adds the claim to native tokens, a "server self-auth via native token" test can be added then; if it never does, the suite already reflects that combination as unsupported.

PR #1240, opened 2026-09-13
fix(lib): give each thread its own VCDiffDecoder

xwang1498:
The shared static decoder keeps per-decode state, so two channels decoding on different threads at the same time were corrupting each other's deltas (40018, then a re-attach). Use a ThreadLocal instead.

Use older syntax for backwards compatibility.

I can add tests if you'd like, but it felt like a lot of ceremony to verify a pretty clear fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing compressed messages concurrently by providing each thread with its own decoder instance.


ttypic (approved):
LGTM, thanks for contribution

PR #1241, opened 2026-09-14
fix(okhttp): complete the close handshake when the server closes the websocket

xwang1498:
When the server closes the websocket, the OkHttp engine never responds with its own close frame. The connection sits half-open and the transport only finds out once the idle timer expires. (Is a server-side close expected to be common? It shows up fairly often in practice.)

OkHttp signals the remote close through onClosing, and [completing](https://github.com/lysine-dev/okhttp/issues/3386) the handshake from there with a close() call lets onClosed fire as normal.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebSocket shutdown handling when the server initiates a connection close.
  - The client now explicitly acknowledges the server’s closing request, helping connections terminate more reliably and reducing the risk of lingering WebSocket sessions.


network-client-okhttp/src/main/java/io/ably/lib/network/OkHttpWebSocketClient.java
```diff
@@ -59,6 +59,11 @@ private WebSocketHandler(WebSocketListener listener) {
             this.listener = listener;
         }
 
+        @Override
+        public void onClosing(WebSocket webSocket, int code, String reason) {
+            webSocket.close(code, reason);
```
ttypic: Could you use 

```suggestion
            webSocket.close(1000, null);
```

instead? 

`onClosing` can hand you `1005` (peer sent an empty close payload), and `close(1005, …)` throws `IllegalArgumentException` (1004–1006 and 1015–2999 are reserved)

ttypic (changes requested):
Thanks a lot for spotting this! requested small changes

network-client-okhttp/src/main/java/io/ably/lib/network/OkHttpWebSocketClient.java
xwang1498: Yes, I chose to echo the code since that seemed to be the norm among other websocket implementations. You'd know better than me whether Ably's infra will ever send an empty close payload. If that exception path is taken then I believe the connection will still be cleaned up correctly.

ttypic (approved):
LGTM

PR #1244, opened 2026-09-14
test(uts): Fix flaky tests - transient-DISCONNECTED CI flake — record-and-verify instead of awaitState

sacOO7:
## Problem

Three CI failures across two suites shared one signature — `TimeoutCancellationException` in `awaitState(..., DISCONNECTED)` immediately after a disconnect stimulus:

- `UnitInfraSmokeTest > unit infra drives the full mock-WebSocket connection lifecycle` — [run 33174366038](https://github.com/ably/ably-pubsub-java/actions/runs/33174366038) and again on `main` at [run 34841083986](https://github.com/ably/ably-pubsub-java/actions/runs/34841083986)
- `ObjectsFaultsTest > RTO7, RTO8 - mutations during re-sync are buffered and applied` — [run 33166346294](https://github.com/ably/ably-pubsub-java/actions/runs/33166346294)

All three passed on rerun; none was a code regression (the commits on those SHAs were exonerated by diff analysis).

## Root cause

`DISCONNECTED` after a drop from `CONNECTED` is **transient at microsecond scale**. Per RTN15a, `Disconnected.enact` queues the reconnect *before* the DISCONNECTED emit (`ConnectionManager.enactState` runs `enact()` before `connection.onConnectionStateChange` — the RTL3d1 ordering), so the ActionHandler emits `DISCONNECTED` → `CONNECTING` back-to-back. `disconnectedRetryTimeout` and the fake clock are bypassed on this path.

`awaitState` is level-triggered from its call point: it registers a listener and checks the current state. A transition that fires **and is superseded** before the listener registers is lost forever — there is no JMM bug, purely a temporal window. On a loaded CI runner, the coroutine dispatch hop to register the listener can lose to that microsecond cascade; in the unit test the transition even completes synchronously inside `simulateDisconnect()`, before `awaitState` is called. One missed window is unrecoverable (the state never recurs in these scenarios), so the wait consumes its full timeout.

This is exactly the failure mode the UTS corpus already warns about: `writing-test-specs.md` § *Verifying Transient States (Record-and-Verify Pattern)* forbids awaiting transient states post-stimulus, citing RTN15a.

## Fix

Render the corpus's record-and-verify convention inline at every transient-target wait — register a state-recording listener **before** the stimulus, then poll the recorded list (or assert the recorded order after awaiting a sticky state). No new infra API: the pattern is the 1:1 rendering of the spec pseudocode using the existing `pollUntil`/`awaitState` helpers.

- **`UnitInfraSmokeTest`** — recording registered before `simulateDisconnect()`; `pollUntil { disconnected in stateChanges }`; the step comment now states the real mechanism (the previous comment attributed the race to the 300 ms retry timer, which this path bypasses).
- **`ObjectsFaultsTest`** — both disconnect sites: the mid-sync observation asserts the recorded `DISCONNECTED → CONNECTING → CONNECTED` order (subsequence check) after awaiting the sticky `CONNECTED`; the `trigger_action` site gates client A's publish on `pollUntil` over the recording (that wait is load-bearing for sequencing).
- **`ConnectionRecoveryTest`** — same pattern for the transient `CLOSING` wait.
- **`Utils.kt`** — KDoc on `awaitState` documenting it as sticky-targets-only (transient targets must record before the stimulus).
- **`uts/README.md`** — the smoke-test walkthrough now teaches the record-before-stimulus pattern (consistent with the proxy walkthroughs, which already used it) instead of the racy await.

Companion spec PR: https://github.com/ably/specification/pull/521 fixes the same shapes at their source — the two `AWAIT_STATE … DISCONNECTED` sites in `objects_faults.md`, the missing `process_pending_events()` in the RTO17-RTO18 fresh-channel scenario, and the derivation-docs rules — so future derivations in any SDK render the safe pattern. Seven further realtime proxy spec sites share the shape and are tracked as a follow-up there.

## Second race fixed on this branch: RTO17/RTO18 sync-events (fresh-channel pipeline readiness)

The CI re-roll surfaced one more deterministic race in `RealtimeObjectTest > "RTO17 RTO18 - sync event sequences"`: the fresh-channel scenario is the only one that never calls `get()`, so nothing flushed the objects sequential scope before `attach()` — the mock's synchronous ATTACHED+OBJECT_SYNC could beat the internal message collector's subscription to its replay-0 shared flow, silently dropping the OBJECT_SYNC (`tryEmit` to zero subscribers) so SYNCED never fires. Reproduced deterministically under `-XX:ActiveProcessorCount=2` (fails at iteration 1); fixed by rendering the spec's `process_pending_events()` step after `channels.get()`, and by making the file's event-recording lists `CopyOnWriteArrayList` (appended on SDK threads, read from `pollUntil`'s poller thread — one pre-existing site iterated while appending, a latent `ConcurrentModificationException`). Post-fix: 2500 stress iterations + 25/25 runs green under the 2-core constraint. The rendering rules are documented once in `uts/README.md` §8 and the uts-to-kotlin mapping, not per call site.

## Verification

- `:uts:runUtsUnitTests` — **10/10 consecutive stress runs green** (previously the flaking suite).
- `:java:runUtsUnitTests` 6/0 · `:liveobjects:runLiveObjectsUnitTests` 389/0.
- `checkWithCodenarc checkstyleMain checkstyleTest` — green.
- The liveobjects integration suite (covers `ObjectsFaultsTest`) could not run locally due to a network timeout fetching the sandbox test-app fixture; it compiles cleanly and will run in this PR's CI.


ttypic (approved):
LGTM

PR #1245, opened 2026-09-16
Integration branch for  io.ably.pubsub:* v2 packages

ttypic:
Integration branch contains packages split, renaming and deletion of deprecated methods

## Summary by CodeRabbit

* **New Features**
  * Added separate device and server SDK entry points with fluent client builders for REST and realtime connections.
  * Added side-specific client identification in requests and agent headers.
  * Added support for publishing device and server artifacts under the new Pub/Sub package namespace.

* **Deprecations**
  * Deprecated direct `AblyRest` and `AblyRealtime` constructors; use the device or server builders instead.

* **Documentation**
  * Updated README and contribution guidance for the 2.0 package split and migration context.


sacOO7 (commented):
The core split mechanics are implemented well — the exact door surfaces, the `io.ably.pubsub:{core,core-android,device,server}` coordinates with the core POMs marked internal-use, versionless side flags applied last so they win collisions, the renamed family identifier with a wire-level canary, and single-version lockstep enforced by a release pre-flight. But the branch is not GA-ready: 3 Critical and 9 Major issues follow as individual comments below (plus several Minor/Info ones).

gradle.properties
```diff
@@ -1,5 +1,5 @@
-GROUP=io.ably
-VERSION_NAME=1.8.0
+GROUP=io.ably.pubsub
```
sacOO7, line 1: **[CRITICAL] Silent duplicate classes in mixed 1.x/2.x graphs — no Gradle capability or relocation POM**

The split moves the group to `io.ably.pubsub` (here in `gradle.properties`) but the classes stay in `io.ably.lib.*` (`core` compiles `../lib/src/main/java`, and `core-android` keeps `namespace = "io.ably.lib"`). Because `io.ably:ably-java:1.x` and `io.ably.pubsub:core:2.x` are different coordinates, neither Gradle nor Maven can tell they're the same code — a consumer whose graph pulls both (very likely during the migration window: e.g. chat on the new core in one branch of the tree, another dependency still on `ably-java`) ends up with two jars defining `io.ably.lib.rest.AblyRest` on one classpath, and the JVM loads whichever comes first. The PubSub device/server split rollout (PDR-091b2) explicitly keeps the old package alive for a year, so overlapping graphs are the expected case, not the exception. Please declare a Gradle capability on the new artifacts matching the old modules (`io.ably:ably-java` / `io.ably:ably-android`) so Gradle raises a duplicate-capability error consumers can resolve, and consider a relocation POM on the old coordinates. Without one of these, the collision is completely silent. The same applies to the regrouped `network-client-*` artifacts.

.github/workflows/release.yaml
```diff
@@ -46,6 +46,12 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@d9c87d481d55275bb5441eef3fe0e46805f9ef70 # v3
 
+      # Fails before anything is uploaded if the artifact set, group or lockstep
+      # version drifts (core, core-android, device and server release
+      # together on the same version; partial release must be impossible).
+      - name: Release pre-flight
```
sacOO7, line 52: **[CRITICAL] `release.yaml` silently no-ops after the repo rename**

`release.yaml` guards the release job with `if: github.repository == 'ably/ably-java'` (line 9), but this repository is now `ably/ably-pubsub-java` (as `settings.gradle.kts`'s `rootProject.name` already reflects, per the PubSub device/server split rollout, PDR-091b2). With the old name in the guard the condition is false, the job is skipped, and the workflow reports success having published nothing — releases silently stop. Update the guard to `ably/ably-pubsub-java`, keeping the fork-protection intent, and sweep for any other `github.repository` guards. (Anchored on the added pre-flight step; the guard itself is on line 9, outside this PR's hunks.)

device/src/androidTest/java/io/ably/pubsub/device/PubSubDeviceTest.java
```diff
@@ -0,0 +1,72 @@
...
+    private static void assertDeviceFlag(Map<String, String> agents) {
+        assertTrue("expected the device side flag", agents.containsKey(Side.DEVICE_AGENT_IDENTIFIER));
+        assertNull("the side flag is versionless", agents.get(Side.DEVICE_AGENT_IDENTIFIER));
+        assertFalse("a device client must not carry the server entry",
+            agents.containsKey(Side.SERVER_AGENT_IDENTIFIER));
+    }
+
+    @Test
+    public void client_stampsDeviceAgent() throws Exception {
+        AblyRealtime client = PubSubDevice.clientBuilder(offlineOptions(FAKE_KEY)).build();
```
sacOO7, line 44: **[CRITICAL] Device `androidTest` cannot compile; `emulate.yml` runs it**

This test can't compile against the shipped door: it calls `PubSubDevice.clientBuilder(offlineOptions(FAKE_KEY))`, `clientBuilder(FAKE_KEY)` and `clientBuilder(options)`, but `PubSubDevice` only exposes a no-arg `clientBuilder()` plus fluent setters — those overloads don't exist. It also imports `io.ably.pubsub.internal.Side`, which lives in the `shared/` directory that isn't registered in `settings.gradle.kts` and isn't a dependency of `:device`. Since `emulate.yml` runs `:device:connectedAndroidTest` (on `pull_request` and `push` to `main`, so it is PR-blocking), this breaks the Android CI job. It looks like a leftover from an earlier options-passing design — the fluent `PubSubDeviceTest.kt` already covers the same cases. Either delete this file or rewrite it against the actual fluent API (importing `io.ably.lib.util.Side`, the class the door actually uses). Broken door tests mean the device side ships without verified stamping.

shared/src/main/java/io/ably/pubsub/internal/Side.java
```diff
@@ -0,0 +1,103 @@
...
+
+import io.ably.lib.types.AblyException;
+import io.ably.lib.types.ClientOptions;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Internal helper shared by the {@code io.ably.pubsub:device} and {@code io.ably.pubsub:server}
+ * door artifacts. It is compiled into each artifact's output from a shared source directory
```
sacOO7: **[MAJOR] Two divergent `Side` helpers; the canonical copy-based one is dead code (and the live one leaks into core's public API)**

This PR adds two `Side` helpers. `io.ably.pubsub.internal.Side` (in `shared/`) is the good one — `optionsWithSideAgent` copies via `ClientOptions.copy()`, never mutates the caller's options, has key/token overloads, and carries the full rationale that the side entry is a versionless flag registered in the ably-common agents registry and load-bearing for MAU billing. But `shared` isn't in `settings.gradle.kts` and nothing depends on it, so that class is never compiled — its own Javadoc's claim (this line) that it "is compiled into each artifact's output" is false. The doors instead call `io.ably.lib.util.Side.injectSideAgent(...)`, which reassigns `options.agents` on the object it's given and documents only the suffix — and, living in `lib`, it ships as public API of the published `io.ably.pubsub:core`. Please collapse to a single implementation: either register `:shared`, route both doors through the copy-based `optionsWithSideAgent`, and delete `io.ably.lib.util.Side`; or delete `shared/` and move the full versionless-flag documentation onto the live class (and restrict its visibility if possible). Two divergent copies of a billing helper, with the authoritative docs on the dead one, is a trap for the next maintainer — and it's what the broken device androidTest imported.

.github/workflows/javadoc.yml
```diff
@@ -39,6 +39,6 @@ jobs:
       - name: Upload Documentation
         uses: ably/sdk-upload-action@4e694297f208b72b5a9f6b1248a1556f19f821d6 # v2
         with:
-          sourcePath: java/build/docs/javadoc
+          sourcePath: core/build/docs/javadoc
```
sacOO7, line 42: **[MAJOR] `javadoc.yml` assumes the old per-repo IAM role — docs upload fails silently after the rename**

Before running `ably/sdk-upload-action` on this `sourcePath`, `javadoc.yml` assumes `role/ably-sdk-builds-ably-java` (line 24). This repo is renamed to `ably-pubsub-java` under the PubSub device/server split rollout (PDR-091b2), and the rollout explicitly warns that these per-repo OIDC/IAM bindings break on rename: the role's trust policy no longer matches the workflow's OIDC subject, so the docs upload starts failing silently on the next push to main. The IAM role (name and trust-policy repo binding) needs re-provisioning for `ably/ably-pubsub-java`, and the ARN on line 24 updated, in the same pass as the rename.

sacOO7, line 42: **[MAJOR] Docs published only for the internal `core`; the public doors get no docs site**

This uploads only `core/build/docs/javadoc`, and the `device` module publishes an empty javadoc jar. Under the PubSub device/server split rollout (PDR-091b2), `core` is the internal-only artifact and `PubSubDevice`/`PubSubServer` are the recommended public entry points — so as configured, the docs site documents the one artifact users must not depend on and omits the public API. Please generate and publish docs for the `device` and `server` modules (in addition to, or instead of, `core`).

server/src/test/java/io/ably/pubsub/server/PubSubServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+public class PubSubServerTest {
+
+    private static final String FAKE_KEY = "fakeAppId.fakeKeyId:fakeKeySecret";
+    private static final String FAKE_TOKEN = "fakeTokenString";
+
+    /**
+     * The stamped entry is present as a versionless flag, and the other side's is absent.
+     */
+    private static void assertServerFlag(Map<String, String> agents) {
+        assertTrue("expected the server side flag", agents.containsKey(Side.SERVER_AGENT_IDENTIFIER));
```
sacOO7, line 43: **[MAJOR] No test pins the literal side-flag strings — a value rename fails no test**

The side-flag tests are strict about *form* (versionless, no `/suffix`) but never assert the *value*: they all reference `Side.SERVER_AGENT_IDENTIFIER`/`DEVICE_AGENT_IDENTIFIER` symbolically (as here), and the only hardcoded wire literal is the family id `ably-pubsub-java/`. So changing the constant's value — say `"ably-pubsub-server"` → `"ably-server"` — keeps every test green while silently reclassifying every server client's MAU billing (the `-server` suffix is what earns the server exemption in the PubSub device/server split rollout, PDR-091b2). Please pin the literals: `assertEquals("ably-pubsub-device", Side.DEVICE_AGENT_IDENTIFIER)` / same for server, and assert the raw string in the on-the-wire header test.

uts/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 dependencies {
-    testImplementation(project(":java"))
+    testImplementation(project(":core"))
```
sacOO7: **[MAJOR] UTS conformance does not run through the doors (#1235 unmerged)**

The UTS still constructs clients directly (`ClientFactories.kt` → `new AblyRealtime(...)` / `new AblyRest(...)`); this file only renames the `:java` dependency to `:core`. The "UTS through the doors" change is PR #1235, still open and not merged into `integration/v2`. As it stands this branch could GA with no conformance evidence that a client built via `PubSubDevice.clientBuilder()` / `PubSubServer.realtimeClientBuilder()` behaves identically to a directly-constructed one — the acceptance bar the other SDKs in the PubSub device/server split rollout (PDR-091b2) are holding. Please merge #1235 (construction-mode seam + harness self-test) before GA, or record explicitly why door-routed conformance is deferred.

server/src/test/java/io/ably/pubsub/server/PubSubServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    /**
+     * Wire-level assertion: the Ably-Agent header actually sent over HTTP carries the
+     * side-declaring flag as a bare token alongside the core's base identifier. This is the
+     * value billing classification reads.
+     */
+    @Test
+    public void httpRequests_carryTheServerAgentHeaderOnTheWire() throws Exception {
```
sacOO7, line 138: **[MAJOR] Wire-level side-flag coverage: only server-HTTP; device (HTTP+WS) and server-WS missing**

The only on-the-wire side-flag assertion is this server HTTP test (`httpRequests_carryTheServerAgentHeaderOnTheWire`). The device tests inspect `client.options.agents` in memory only, and the WebSocket `agent`-param test covers a core-constructed client, not a door client. That a core-constructed client carries no side flag is also only proven implicitly (via the WS canary's exact-equality on the whole agent string), with no explicit HTTP negative assertion. Since the bare-token form of `ably-pubsub-device`/`ably-pubsub-server` on both transports is what MAU classification reads (PubSub device/server split rollout, PDR-091b2), please add: a device HTTP `Ably-Agent` wire test mirroring this one, and WebSocket `agent`-param assertions for door-built realtime clients on each side (bare token, no `/version`).

device/build.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+}
+
+android {
+    namespace = "io.ably.pubsub.device"
+    compileSdk = 34
+
+    defaultConfig {
+        minSdk = 24
```
sacOO7, line 54: **[MAJOR] `device` raises the Android floor to minSdk 24 (core-android is 19), undocumented**

`device` sets `minSdk = 24`, but the artifact it supersedes (`io.ably:ably-android`, now `core-android`) is `minSdk 19` — and since `device` exposes `core-android` via `api(...)`, every device consumer inherits the 24 floor. An API 19–23 app migrating from `ably-android` to the new device artifact under the PubSub device/server split rollout (PDR-091b2) fails to build. (`emulate.yml`'s matrix also runs `:device:connectedAndroidTest` on an API-21 emulator, below this floor, so that leg can't install the device APK even once the androidTest compile break is fixed.) If the raise is intended, document it as a breaking change in the README migration note and CHANGELOG (and reconcile with the in-flight `drop-api-19` work); if not, drop it to 19 to match `core-android`.

gradle.properties
```diff
@@ -1,5 +1,5 @@
...
+VERSION_NAME=2.0.0
 POM_INCEPTION_YEAR=2015
 POM_URL=https://github.com/ably/ably-java
```
sacOO7: **[MINOR] POM `url`/`scm`/license metadata still point at `github.com/ably/ably-java`**

These POM properties still point at `github.com/ably/ably-java`, but the repo is now `ably-pubsub-java` (PubSub device/server split rollout, PDR-091b2). They're baked into every `io.ably.pubsub:*` POM published to Maven Central and can't be edited after release — GitHub's redirect keeps the links working, but the brand-new 2.0 coordinates would permanently advertise the old repo. Please repoint `POM_URL` (this line), `POM_SCM_URL`, `POM_SCM_CONNECTION`, `POM_SCM_DEV_CONNECTION` and `POM_LICENSE_URL` (the latter three are just below, outside this PR's hunk) before the first 2.0 publish.

README.md
```diff
@@ -15,6 +15,23 @@ Find out more:
 
 ---
 
+> [!NOTE]
```
sacOO7, line 18: **[MINOR] No migration guide or 2.0.0 CHANGELOG; README install/sample still uses the deprecated 1.x path**

This new README note is a good start, but before release there's no migration path for `io.ably:ably-java` 1.x users: which coordinate to move to, which door to pick per side, and that the old constructors will throw once MAU-based pricing is live (the forcing function in the PubSub device/server split rollout, PDR-091b2). The install section and sample further down still use the now-deprecated `new AblyRealtime(...)` without flagging it, the badges still resolve the legacy `io.ably/ably-java` artifact, and CHANGELOG has no 2.0.0 entry. (Minor: the door signature in the table above reads `clientBuilder(...)` though the factory takes no args — the chain is `clientBuilder().key(...).build()`.) Please add a migration section and a 2.0.0 CHANGELOG entry before this ships.

build.gradle.kts
```diff
@@ -30,6 +31,56 @@ subprojects {
...
+ * Release pre-flight: the split ships core, core-android, device and server in lockstep
+ * (one version, one run), so the set of published artifacts and their
+ * coordinates are asserted here and the release workflow fails before anything is
+ * uploaded if they drift. If you add or remove a published module, update this list
+ * deliberately.
+ */
+val expectedReleaseArtifacts = sortedSetOf(
+    "io.ably.pubsub:core:jar",
+    "io.ably.pubsub:core-android:aar",
+    "io.ably.pubsub:device:jar",
```
sacOO7, line 44: **[MINOR] Lockstep pre-flight models the KMP `device` module as a single `device:jar`**

`verifyReleaseArtifacts` asserts `io.ably.pubsub:device:jar`, but `device` is a Kotlin Multiplatform module that publishes `device` (metadata), `device-jvm` and `device-android`. The pre-flight therefore doesn't model the coordinates consumers actually resolve and wouldn't catch drift in the `-jvm`/`-android` variants. Please enumerate the real KMP publications in `expectedReleaseArtifacts`, or document that the KMP variants are approximated.

server/src/test/java/io/ably/pubsub/server/PubSubServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+            .agents(callerAgents)
+            .build();
+        assertNotSame("the stamp must go into a copy, not the caller's map",
+            callerAgents, client.options.agents);
+        assertEquals(1, callerAgents.size());
+        assertFalse(callerAgents.containsKey(Side.SERVER_AGENT_IDENTIFIER));
+    }
+
+    @Test
+    public void noAuthParameters_getTheCoreConstructorsOwnError() {
```
sacOO7, line 123: **[MINOR] Device door has no invalid/no-auth error-path test**

The server door tests the no-auth path (`noAuthParameters_getTheCoreConstructorsOwnError`, this test), but the device door has no equivalent — the case only existed in the non-compiling androidTest. Please add a device test asserting `PubSubDevice.clientBuilder().build()` with no auth surfaces the core constructor's own error, matching the server side.

lib/src/main/java/io/ably/lib/util/Side.java
```diff
@@ -0,0 +1,67 @@
...
+ * declares.
+ */
+public final class Side {
+    private Side() {}
+
+    /*
+     * The `-device` / `-server` suffix on both identifiers below is load-bearing, not
+     * cosmetic. On API-key auth the realtime system grants the server exemption by matching
+     * an agent entry ending in `-server`, and an identifier that is not yet in the
+     * ably-common registry is classified by that suffix alone. Renaming either without
```
sacOO7, line 25: **[MINOR] No in-repo validation that the new agent identifiers are registered in ably-common**

The side identifiers and the renamed family id exist only as constants here; there's no ably-common submodule or vendored `agents.json`, so nothing in this repo proves `ably-pubsub-device`, `ably-pubsub-server` and `ably-pubsub-java` are actually registered — and this comment already relies on it. Registration-before-ship is a requirement of the PubSub device/server split rollout (PDR-091b2). Please link the ably-common registration PR here and confirm it's merged before GA; ideally add a check pinning these against the registry.

deploy/check-archives.sh
```diff
@@ -36,8 +36,8 @@ case "${ABLY_VERSION}" in
 		;;
 esac
 
-ABLY_JAVA_RELEASE_DIR="${ABLY_HOME}/java/build/release/${ABLY_VERSION}/io/ably/ably-java/${ABLY_VERSION}"
-ABLY_ANDROID_RELEASE_DIR="${ABLY_HOME}/android/build/release/${ABLY_VERSION}/io/ably/ably-android/${ABLY_VERSION}"
+ABLY_JAVA_RELEASE_DIR="${ABLY_HOME}/core/build/release/${ABLY_VERSION}/io/ably/ably-java/${ABLY_VERSION}"
```
sacOO7, line 39: **[MINOR] `deploy/check-archives.sh` is half-migrated (and appears dead)**

This edit updates the build-dir prefixes but leaves the Maven path `io/ably/ably-java` and the `ably-java`/`ably-android` filename prefixes, which don't exist under the new `io.ably.pubsub` coordinates — and it still reads the version from `build.gradle`, which is now `build.gradle.kts` with the version in `gradle.properties`. Nothing in CI invokes this script; if it's dead please delete it, otherwise finish the coordinate migration.

sacOO7 (commented):
(Duplicate submission of the review above, caused by a tooling retry — its inline comments have been removed; please refer to the review above.)

sacOO7:
**[MAJOR] `features.yml` still passes `repository-name: ably-java`**

`features.yml` passes `repository-name: ably-java` into the shared `ably/features` workflow, which uses it for the OIDC role assumption and the sdk.ably.com upload path. With the repo renamed to `ably-pubsub-java` (PubSub device/server split rollout, PDR-091b2), the features report will publish under the wrong key or fail role assumption — silently, on push to main. Please update it to `ably-pubsub-java` in lockstep with the rename (and confirm the `ably/features` repo knows the new name).

_(Posted as a general comment because `features.yml` is not touched by this PR.)_

sacOO7:
**[MAJOR] No 1.x maintenance branch; merge to `main` strands 1.x fixes**

Merging this branch renames the Maven group to `io.ably.pubsub`, bumps to 2.0.0, and deletes the `io.ably:ably-java`/`io.ably:ably-android` artifact definitions — after which `main` can no longer cut a 1.x security or critical fix. The PubSub device/server split rollout (PDR-091b2) requires a dedicated maintenance branch for the existing packages (security/critical fixes only, 1-year support window, then EOL). No such branch exists on the remote today — only ephemeral `release/*` branches and the `v1.8.2` tag. Please cut the 1.x maintenance branch (e.g. from `v1.8.2`) before this merges, and state the support-window/EOL plan in the PR description.

_(Posted as a general comment because this is about repository/branch state, not a specific changed line.)_

sacOO7:
**[MINOR] `LIBRARY_NAME` build-config fields are dead after the family-identifier flip**

`LIBRARY_NAME` is still generated into `BuildConfig` for both `core` ("java") and `core-android` ("android"), but nothing reads it — `Defaults.ABLY_AGENT_VERSION` now hardcodes `ably-pubsub-java`. That also means the Android artifact no longer reports a distinct `ably-android`-style family; fine under the PubSub device/server split rollout (PDR-091b2), which defines a single family per SDK, but worth a deliberate note. Please drop the dead fields (or add a comment recording that Android intentionally shares the `ably-pubsub-java` family) so the next reader isn't misled into thinking they still drive the agent string.

_(Posted as a general comment because `core/build.gradle.kts` and `core-android/build.gradle.kts` are pure renames with no changed lines to anchor to in this PR.)_

sacOO7:
**[INFO] No API/binary-compatibility gate on the new public surface**

There's no API/binary-compatibility check on the new public door surface — no `japicmp`, `metalava`, `revapi` or Kotlin `apiValidation` anywhere in the build (also true pre-split, so not a regression); `device` has Kotlin `explicitApi()`, which governs visibility only. Reasonable to defer for a greenfield v2 — flagging it so it's a conscious choice rather than an oversight, given these are brand-new public APIs.

sacOO7 (commented):
You can take a look at pending review comments.
If already addressed, then you can accordingly close or reply to review comments 👍

